PR #2, opened 2021-11-04
EDU-39 Create inital user by configuration

ahelmberger:


test-app/index.js
```diff
@@ -19,5 +19,10 @@ educandu({
     port: 8025,
     ignoreTLS: true
   },
-  publicFolders: ['./test-app/dist', './test-app/static'].map(x => path.resolve(x))
+  publicFolders: ['./test-app/dist', './test-app/static'].map(x => path.resolve(x)),
+  initialUser: {
+    username: 'admin',
+    password: 'admin',
+    email: 'admin@test.com'
```
adrianaluchian, line 26: Could we call this `user` and not `admin`? I think it will get confusing as there's a role `admin` and this user does not have it.

ahelmberger: Sure, we can continue call it test

ahelmberger: Done!

adrianaluchian (approved):

PR #3, opened 2021-11-04
EDU-33 Serve default settings when we have none

laurentiu-ilici:


src/services/setting-service.js
```diff
@@ -5,10 +5,29 @@ class SettingService {
...
+      footerLinks: [],
+      defaultTags: []
+    };
   }
 
   async getAllSettings() {
     const settings = await this.settingStore.find();
+    if (!settings.length) {
+      return this.defaultSettings;
+    }
```
ahelmberger, line 29: What happens when there are only partial settings (eg: `{ termsPage: { ... } }` in the db?

adrianaluchian, line 29: This covers the case when no settings exist at all. 
For consistency I'd suggest to do smth like
```
const settingsArray = await this.settingStore.find();
const settings = settingsArray.reduce((all, { _id, value }) => {
      all[_id] = value;
      return all;
});
return {
   ...defaultSettings,
   ...settings
 }
```

ahelmberger (approved):

adrianaluchian (approved):

PR #6, opened 2021-11-05
EDU-36 Remove elmu related references

laurentiu-ilici:
Resolvels: https://educandu.atlassian.net/jira/software/projects/EDU/boards/1?selectedIssue=EDU-36


src/services/user-service.js
```diff
@@ -8,7 +8,7 @@ import { SAVE_USER_RESULT } from '../domain/user-management.js';
 import PasswordResetRequestStore from '../stores/password-reset-request-store.js';
 
 const DEFAULT_ROLE_NAME = ROLE.user;
-const PROVIDER_NAME_ELMU = 'elmu';
+const PROVIDER_NAME_EDUCANDU = 'elmu';
```
adrianaluchian, line 11: I think this should not change, or? It's getting confusing.

laurentiu-ilici: We need to make these configurable / client. I left the value elmu because we don't have that yet, but the provider name will probably have to be configurable for each client.
My bigger problem with this one was if to rename it to this or EDUCANDU_PROVIDER_NAME

adrianaluchian: Then just `PROVIDER_NAME`, since it will be configurable?

ahelmberger: In the end, I think this should be "educandu" as it is the built-in auth provider. Opposed to "shibboleth" etc.

ahelmberger: But we need a migration for that

src/plugins/audio/audio.yml
```diff
@@ -7,7 +7,7 @@ source:
 externalLink:
   en: External link
   de: Externer Link
-elmuCdn:
+internalCdn:
   en: Elmu CDN
```
ahelmberger, line 11: We can call these "Internal CDN"/"Internes CDN"

src/styles/variables.less
```diff
@@ -15,7 +15,7 @@
 @global-content-width-narrow: 800px;
 @global-button-spacing: 6px;
 
-@global-elmu-blue: @primary-color;
+@global-educandu-blue: @primary-color;
```
ahelmberger, line 18: or maybe `@global-primary-color`?

src/plugins/audio/audio.yml
laurentiu-ilici: Ah, I thought we can override somehow the values in the future.

src/styles/variables.less
laurentiu-ilici: Ok

src/plugins/audio/audio.yml
laurentiu-ilici: Done

src/styles/variables.less
laurentiu-ilici: Done

src/services/user-service.js
laurentiu-ilici: Renamed to provider name so far since I don't know what this entails.

adrianaluchian (approved):

PR #7, opened 2021-11-05
EDU-42 Add archived flag to documents

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-42

### DIFF
--- a/src/services/document-service.js
+++ b/src/services/document-service.js
@@ -217,7 +217,8 @@ class DocumentService {
         namespace: doc.namespace,
         language: doc.language,
         sections: newSections,
-        tags: doc.tags
+        tags: doc.tags,
+        archived: false
       };
 
       logger.info('Saving new document revision with id %s', newDocumentRevision._id);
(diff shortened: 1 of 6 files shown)


src/services/document-service.js
```diff
@@ -342,7 +343,8 @@ class DocumentService {
       language: lastRevision.language,
       sections: lastRevision.sections,
       contributors,
-      tags: lastRevision.tags
+      tags: lastRevision.tags,
+      archived: lastRevision.archived
     };
   }
 }
```
ahelmberger, line 350: I'm kinda missing the `default-prop-types.js`. I guess we want to avoid console errors in the frontend.

adrianaluchian: Course, overlooked this, thanks!
FYI I also corrected tags mapping for document revision restore, seems we were reintroducing null tags in the DB this way (and losing old ones)

ahelmberger (approved):

laurentiu-ilici (approved):

PR #8, opened 2021-11-08
feature/EDU-27 Add archiving functionality

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-27


src/components/pages/search.yml
```diff
@@ -2,7 +2,7 @@ title:
   en: Title
   de: Titel
 udateDate:
-  en: Udate date
+  en: Update date
```
laurentiu-ilici, line 5: Nice spot, maybe udate the key as well :D

src/services/document-api-client.js
```diff
@@ -61,6 +61,24 @@ class DocumentApiClient {
       .accept('json')
       .then(res => res.body);
   }
+
+  archiveDocument(documentKey) {
+    return this.httpClient
+      .post(`/api/v1/docs/${documentKey}/archive`)
```
laurentiu-ilici, line 67: Not really restful, but I guess we don't care that much.

src/services/document-service.js
```diff
@@ -33,7 +34,8 @@ const searchResultsProjection = {
   slug: 1,
   contributors: 1,
   updatedOn: 1,
-  tags: 1
+  tags: 1,
+  archived: 1
```
laurentiu-ilici, line 38: I would not include this in the search projection, it's not needed.

src/services/document-service.js
```diff
@@ -82,16 +84,20 @@ class DocumentService {
     this.userService = userService;
   }
 
-  getAllDocumentsMetadata() {
-    return this.documentStore.find({}, { sort: lastUpdatedFirst, projection: metadataProjection });
-  }
-
-  getDocumentsMetadataByKeys(documentKeys) {
-    return this.documentStore.find({ _id: { $in: documentKeys } }, { sort: lastUpdatedFirst, projection: metadataProjection });
+  getAllDocumentsMetadata({ includeArchived } = {}) {
```
laurentiu-ilici, line 87: This could be more generic:
 (extraFilters = {})

Then you have 
const filter = {...extraFilters}

src/services/document-service.js
```diff
@@ -82,16 +84,20 @@ class DocumentService {
...
+    const filter = {};
+    if (!includeArchived) {
+      filter.archived = false;
+    }
+    return this.documentStore.find(filter, { sort: lastUpdatedFirst, projection: metadataProjection });
   }
 
-  async getDocumentsByTags(searchTags) {
-    const result = await this.documentStore.find({ tags: { $all: searchTags } }, { projection: searchResultsProjection });
+  async getDocumentsByTags(tags, { includeArchived } = {}) {
```
laurentiu-ilici, line 95: Why not always include the tags, I guess those should be relevant anyway if they were included in the system at some point

src/services/document-service.js
```diff
@@ -324,6 +330,23 @@ class DocumentService {
     }
   }
 
+  async toggleDocumentArchived({ documentKey, user, archived }) {
```
laurentiu-ilici, line 333: This name is misleading, a toggle would work in a sense that it gets the last state and toggles it to the other state.
Mayne setArchivedState?

src/server/document-controller.js
```diff
@@ -129,6 +130,26 @@ class DocumentController {
       return res.send({ documentRevisions });
     });
 
+    router.post('/api/v1/docs/:key/archive', needsPermission(permissions.MANAGE_ARCHIVED_DOCS), async (req, res) => {
```
laurentiu-ilici, line 133: Why not go with the normal way of updating the documents? Making 2 new endpoints for this seems excessive :)

adrianaluchian: Cause I don't want to pass in all the data of the new revision, just the value of the flag 🤷 Less error prone also.

src/services/document-service.js
adrianaluchian: Not really, `includeArchived` is not a db filter as it's not part of the model. `archived` is part of the model. I specifically did not want the controller to call this method passing in the DB filter, since it's out of its scope to handle the filtering specific to the DB type used.

src/services/document-service.js
adrianaluchian: I thought we said we'll want to show for admins the documents that are archived with some sort of labelling (can't really test this yet though, waiting to get your PR merged first)

src/services/document-api-client.js
adrianaluchian: Not all our endpoints are completely REST-ful, it's a matter of convenience with this one. Making this fully RESTful would mean quite some changes 😬

src/components/pages/search.yml
adrianaluchian: Ah true, corrected, thanks!

src/services/document-service.js
laurentiu-ilici: Ah, okay, I didn't know that

src/server/document-controller.js
laurentiu-ilici: Feels like a PATCH operation to me, but okay.

src/services/document-service.js
adrianaluchian: Yes makes sense, renamed.

src/services/document-service.js
adrianaluchian: Um, what do you mean? I changed nothing regarding tags, apart from streamlining their name.

laurentiu-ilici: Ahm, my bad, I didn't read which method this was.

src/server/document-controller.js
adrianaluchian: Ok so you meant from verb perspective not path. I can change to PATCH, true

adrianaluchian: gePATCHed. I kept 2 endpoints though, as we have multiple routes starting with `/api/v1/docs/` and wanted to make sure there'll be no clashed now (if the order of the declarations is not correct) or in the future

ahelmberger: Sounds good to me. The rest of the code is not very religious about REST though, especially as we do not really patch anyway (it's always a new revision, not an in-place update).

adrianaluchian: Yes true, basically we always POST

src/server/document-controller.js
```diff
@@ -129,6 +130,26 @@ class DocumentController {
       return res.send({ documentRevisions });
     });
 
+    router.patch('/api/v1/docs/:key/archive', needsPermission(permissions.MANAGE_ARCHIVED_DOCS), async (req, res) => {
```
laurentiu-ilici, line 133: This is not what I meant by patch, more like this:
https://restcookbook.com/HTTP%20Methods/patch/
I think it does exactly what you want it do do and it can be used for other updates as well. Again we don't need to go this route, it's just the more restful way.

laurentiu-ilici: Maybe a better link:
https://stackoverflow.com/questions/28459418/use-of-put-vs-patch-methods-in-rest-api-real-life-scenarios

adrianaluchian: I don't get it, based on the links you did mean PATCH as the HTTP verb, or?

laurentiu-ilici: Yes, but the implementation here is not according to "spec":
 you could have 
api/v1/docs/:key
verb: PATCH
body:
{
   archived: true/false
}

Then for future such updates we could avoid to make new endpoints. The caveats are that these are not idempotent operations and we could introduce some race conditions. 

The reason why I suggested this was that you wanted yesterday to avoid doing a PUT which would need to send the whole object.

adrianaluchian: Ah ok, thanks for explaining what you meant, that helps 😬 
I wrote here https://github.com/educandu/educandu/pull/8#discussion_r745358276 why I preferred to go with 2 endpoints.

ahelmberger (approved):

src/services/document-service.js
```diff
@@ -69,6 +71,26 @@ const getTagsQuery = searchString => [
...
+    key: documentKey,
+    order: nextOrder,
+    restoredFrom,
+    createdOn: dateTime.now(),
+    createdBy: userId,
+    title: doc.title || '',
+    slug: doc.slug || '',
+    namespace: doc.namespace,
+    language: doc.language,
+    sections: sections?.length ? sections : doc.sections,
```
ahelmberger, line 88: What if I want the empty array?

adrianaluchian: That's true, you could have deleted all sections in one revision. Corrected, thanks!

ahelmberger (approved):

laurentiu-ilici (approved):

PR #9, opened 2021-11-08
Feature/edu 26 intuitive search

laurentiu-ilici:


src/components/pages/search.js
```diff
@@ -1,26 +1,49 @@
...
   const { docs } = initialState;
+  const { query } = useRequest();
+
   const sortedDocs = useMemo(
-    () => docs.sort(firstBy(doc => doc.contributors)
-      .thenBy(doc => doc.updatedOn, 'desc')),
+    () => docs
+      .map(doc => ({
+        ...doc,
+        tagsSet: new Set(doc.tags)
```
adrianaluchian, line 24: Can it be that a document has duplicated tags? 🤔

src/components/pages/search.js
```diff
@@ -1,26 +1,49 @@
...
+  const [filteredDocs, setFilteredDocs] = useState([...sortedDocs]);
+
+  const [selectedTags, setSelectedTags] = useState([]);
+
+  const allTags = docs.reduce((acc, doc) => {
+    doc.tags.forEach(tag => acc.add(tag));
+    return acc;
+  }, new Set());
+
+  const handleTagsChanged = selectedValues => {
```
adrianaluchian, line 40: Small thing: maybe `selectedTags` would be better here.

src/components/pages/search.js
```diff
@@ -58,13 +82,29 @@ function Search({ initialState }) {
...
+
+      <div className="Search-searchSelectContainer">
+        <Form.Item label={t('refineSearch')} >
+          <Select
+            mode="multiple"
+            tokenSeparators={[' ']}
+            value={selectedTags}
+            onChange={selectedValues => handleTagsChanged(selectedValues)}
+            placeholder={searchPlaceholder()}
+            options={Array.from(allTags).map(tag => ({ value: tag, key: tag }))}
```
adrianaluchian, line 95: I think it's maybe better to extract this in a variable and have less logic in the HTML.

src/components/pages/search.yml
```diff
@@ -7,6 +7,9 @@ udateDate:
...
   de: Tags
-numberOfContributors:
-  en: Contributor count
-  de: Anzahl Mitwirkender
+refineSearch:
+  en: Filter by tags
+  de: Nach Tags filtern
+searchResultPrefix:
+  en: Results for
+  de: Ergibnise für
```
adrianaluchian, line 15: This part I know: "Ergebnisse" 😬

src/services/document-service.js
```diff
@@ -90,9 +89,28 @@ class DocumentService {
...
+      }))
+    };
+
+    const documents = await this.documentStore
+      .find(query, { projection: searchResultsProjection }) || [];
+
+    return documents.map(result => ({
+      ...result,
+      tagMatchCount: result.tags
+        .reduce((count, tag) => searchTags.has(tag.toLowerCase()) ? count + 1 : count, 0)
```
adrianaluchian, line 112: I think this is a bit overcomplicated, could just be
```
result.tags.filter(tag => searchTags.has(tag.toLowerCase())).length;
```

adrianaluchian:
Having a bit of a blond moment, but I can't find the search input 🤦‍♀️

ahelmberger:
> Having a bit of a blond moment, but I can't find the search input woman_facepalming

Having a blond moment, too. I cannot see the search input :thinking:

src/components/pages/search.yml
ahelmberger: :+1:

src/components/pages/search.js
laurentiu-ilici: The set is used for a O(1) search during counting. 

Old documents have duplicate tags because of the way that they were generated unfortunately.

src/components/pages/search.yml
laurentiu-ilici: I'm zi German here ja? Thanks! Fixed

src/components/pages/search.js
adrianaluchian: I saw where it was used. Ok so there can be duplicate tags, that was my question, otherwise the code would have been a bit misleadin

src/components/pages/search.js
laurentiu-ilici: That would conflict with the state variable "selectedTags"

src/services/document-service.js
laurentiu-ilici: A reduce is semantically more correct, but okay.

src/components/pages/search.js
laurentiu-ilici: Done

src/services/document-service.js
adrianaluchian: Don't see it so, you want a subset that fits a requirement 🤷 You don't sum up/gather up info from each iterated item.

laurentiu-ilici: I don't want the subset, I want one number :). I am reducing a collection to a number.

src/services/document-service.js
```diff
@@ -90,9 +89,28 @@ class DocumentService {
...
 
-  async getDocumentsByTags(searchTags) {
-    const result = await this.documentStore.find({ tags: { $all: searchTags } }, { projection: searchResultsProjection });
-    return result || [];
+  async getDocumentsByTags(searchQuery) {
+    const searchTags = new Set(searchQuery.trim()
+      .split(/\s+/)
+      .map(tag => tag.toLowerCase())
+      .filter(tag => (/^\w{3,30}$/).test(tag)));
+
```
adrianaluchian, line 97: We should do an early return here if `searchTags.length === 0`, otherwise we get a 500 cause mongo throws.
Try to search for ". . ."

src/services/document-service.spec.js
```diff
@@ -312,4 +312,70 @@ describe('document-service', () => {
...
+
+        const resultMap = results.reduce((acc, doc) => {
+          acc[doc.title] = { ...doc };
+          return acc;
+        }, {});
+
+        expect(resultMap[doc1.title].tagMatchCount).toEqual(2);
+        expect(resultMap[doc2.title].tagMatchCount).toEqual(3);
+        expect(resultMap[doc3.title].tagMatchCount).toEqual(1);
+      });
```
adrianaluchian, line 378: Maybe you could also add a test for this unhappy path.
https://github.com/educandu/educandu/pull/9/files#r745483179

Actually might be nice to have a `const noSearchResultCases = [..]` data driven test and also include searching by string with spaces, tags that are all shorter than 3 or longer than 30..

src/services/document-service.js
adrianaluchian: Yeah, by filtering and counting how many you get 😄

src/services/document-service.js
laurentiu-ilici: I got that same error, I added a bad request response for strings that have less than 3 chars. But maybe I should do this as well.

laurentiu-ilici: Yeah true, this doesn't really prevent the ... scenario.

adrianaluchian: Basically any scenario where all tags are shorter than 3 or longer than 30

src/services/document-service.js
laurentiu-ilici: What I'm trying to say is that I don't need the filtered instances, I only need a number. You can get that number in either of the 2 ways, somehow to me a reduce makes more sense since it's the intent here. Anyway, the code now uses filter :).

adrianaluchian:
Regarding the language flags on top of search, I think they overlap purpose now a bit. There are the layout flags in top right corner that dictate the layout translation, and then the ones above search do too, but when you get to the results page, you see the layout in the other language.

I think we should have a different way of setting the search language and changing it should not change the input+button language, just the content language (what you search for). For the UI to be more explicit then, maybe we could have a little dropdown with languages between input and search button. 

What'ch'a think?

src/services/document-service.js
ahelmberger: Can we all agree that `map` and `filter` are both subsets of `reduce` that are easier to write and read ? :laughing:

adrianaluchian: Definitely, which makes `reduce` overcomplicated at times 😁

ahelmberger:
> Regarding the language flags on top of search, I think they overlap purpose now a bit. There are the layout flags in top right corner that dictate the layout translation, and then the ones above search do too, but when you get to the results page, you see the layout in the other language.
> 
> I think we should have a different way of setting the search language and changing it should not change the input+button language, just the content language (what you search for). For the UI to be more explicit then, maybe we could have a little dropdown with languages between input and search button.
> 
> What'ch'a think?

Yes, I would really like to separate the search box from the document displayed, make it its own thing, getting the labels from translations and not from settings. I'll create a ticket for that!

src/services/document-service.js
laurentiu-ilici: Thanks! this was a bigger issue with escaping the special characters as well. Have a look at the tests please :)

adrianaluchian: I would still do an early return 
```
if (searchTags.length === 0) {
return [];
}
```
and bypass the DB query altogether.

Also, could we maybe rather use an npm package to escape here, since regex has more than just this 2 characters as having special meaning? e.g. maybe https://www.npmjs.com/package/escape-string-regexp ?

ahelmberger: :+1: Yes, please

laurentiu-ilici: But I am doing that:

```
if (!searchTags.size) {
      return [];
    }
```

adrianaluchian: Ok cool diffing was not showing this at all to me.

laurentiu-ilici: I added escaping with the library, thanks! Didn't know about this one and didn't search.

adrianaluchian: Didn't know about it either, just found it. Maybe there are better ones also, dunno, was just a suggestion.

src/services/document-service.spec.js
laurentiu-ilici: This should be covered by the test with a tab and sequential spaces.

adrianaluchian (approved):

ahelmberger (approved):

PR #18, opened 2021-11-10
fix/EDU-60 correct document cloning

adrianaluchian:
Closes: https://educandu.atlassian.net/browse/EDU-60

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -42,6 +42,7 @@ const supportedLanguages = ['en', 'de'];
 
 let testAppServer = null;
 let testAppBuildResult = null;
+const containerCommandTimeoutMs = 2000;
 
 Graceful.on('exit', () => {
   testAppServer?.kill();
@@ -60,11 +61,11 @@ const ensureContainerRunning = async ({ containerName, runArgs, afterRun = () =>
   const container = data.containerList.find(c => c.names === containerName);
   if (!container) {
     await docker.command(`run --name ${containerName} ${runArgs}`);
-    await delay(1000);
+    await delay(containerCommandTimeoutMs);
     await afterRun();
   } else if (!container.status.startsWith('Up')) {
     await docker.command(`restart ${containerName}`);
-    await delay(1000);
+    await delay(containerCommandTimeoutMs);
   }
 };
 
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -42,6 +42,7 @@ const supportedLanguages = ['en', 'de'];
 
 let testAppServer = null;
 let testAppBuildResult = null;
+const containerCommandTimeoutMs = 2000;
```
laurentiu-ilici, line 45: I guess this will help me as well, if the laptop ever gets here...

src/components/pages/docs.js
```diff
@@ -129,13 +136,17 @@ class Docs extends React.Component {
...
-      newDocLanguage: doc.language,
-      newDocSlug: doc.slug ? `${doc.slug}-${t('copySlugSuffix')}` : DEFAULT_DOCUMENT_SLUG,
-      newDocBlueprintKey: doc.key,
+    this.setState(prevState => ({
+      newDoc: {
+        ...prevState.newDoc,
+        title: doc.title ? `${doc.title} ${t('copyTitleSuffix')}` : t('defaultDocumentTitle'),
+        language: doc.language,
+        slug: doc.slug ? `${doc.slug}-${t('copySlugSuffix')}` : DEFAULT_DOCUMENT_SLUG,
+        tags: doc.tags || [],
```
laurentiu-ilici, line 145: Would be better to spread doc.tags here and create a new array in order to avoid any nasty side effects.

src/components/pages/docs.js
```diff
@@ -44,70 +44,77 @@ class Docs extends React.Component {
...
-      title: toTrimmedString(title) || t('defaultDocumentTitle'),
-      slug: toTrimmedString(slug) || '',
+      title: toTrimmedString(newDocState.title) || t('defaultDocumentTitle'),
+      slug: toTrimmedString(newDocState.slug) || '',
       namespace: DEFAULT_DOCUMENT_NAMESPACE,
-      language,
+      language: newDocState.language,
       sections: [],
-      tags: [],
+      tags: newDocState.tags,
```
laurentiu-ilici, line 72: Would be better to spread here newDocState.tags to avoid any side effects.

src/components/pages/docs.js
adrianaluchian: Same here

laurentiu-ilici: Ahm, I might have said something dumb though, spreading will not work if the doc.tags are not yet defined, so we need : doc.tags?.slice(0) || []
Thanks for changing, I had my mind that we'd forget when introducing the tags editor on this thing.

adrianaluchian: I already covered the undefined case, no worries, didn't take the comment ad litteram ☺️

laurentiu-ilici (approved):

PR #20, opened 2021-11-12
EDU-71 Add document language to the search results

laurentiu-ilici:
Resoles: https://educandu.atlassian.net/browse/EDU-71


src/components/pages/search.js
```diff
@@ -79,6 +84,11 @@ function Search({ initialState }) {
       title: t('updateDate'),
       dataIndex: 'updatedOn',
       render: renderUpdatedOn
+    },
+    {
+      title: t('language'),
+      dataIndex: 'language',
+      render: lang => <CountryFlagAndName code={languageData[lang]?.flag} name={languageData[lang]?.name || lang} />
     }
```
adrianaluchian, line 92: Just from consistency perspective, would be nice to go with either defining the render function inline or separately (like `renderUpdateOn`).
For me separately is better, makes reading easier. If you go for this, please also do the same for tags 😬

src/components/pages/search.js
```diff
@@ -79,6 +84,11 @@ function Search({ initialState }) {
...
   ];
```
adrianaluchian, line 93: It's a small edge case maybe, but is there maybe a quick way to also set a min width for some columns when defining them? 🤔 

Cause of tags it could happen that the other columns get shrunk and for example the language flag + name get wrapped on separate lines 👇 

Screenshot 2021-11-12 at 10 25 25

Another quick idea would be to make the space between the icon and text in `CountryFlagAndName` a non breaking space, if that's possible.

src/components/pages/search.js
laurentiu-ilici: Done, thanks!

src/components/pages/search.js
laurentiu-ilici: I've done a fix, not sure how to do it in another way, but it works :).

adrianaluchian (approved):

PR #21, opened 2021-11-12
Feature/edu 72 replace momentjs with date fns

laurentiu-ilici:


yarn.lock
```diff
@@ -6500,7 +6500,7 @@ mkdirp@^1.0.3:
   resolved "https://registry.yarnpkg.com/mkdirp/-/mkdirp-1.0.4.tgz#3eb5ed62622756d79a5f0e2a221dfebad75c2f7e"
   integrity sha512-vVqVZQyf3WLx2Shd0qJ9xuvqgAyKPLAiqITEtqW0oIUjzo3PePDd6fW9iFz30ef7Ysp/oiWqbhszeGWW2T6Gzw==
 
-moment@^2.24.0, moment@^2.25.3, moment@~2.29.1:
+moment@^2.24.0, moment@^2.25.3:
```
adrianaluchian, line 6503: Can't we actually remove moment now or is it still needed for something?

src/services/user-service.js
```diff
@@ -10,8 +10,8 @@ import PasswordResetRequestStore from '../stores/password-reset-request-store.js
 const DEFAULT_ROLE_NAME = ROLE.user;
 const PROVIDER_NAME = 'elmu';
 const PASSWORD_SALT_ROUNDS = 1024;
-const PENDING_USER_REGISTRATION_EXPIRATION_IN_HOURS = 24;
-const PENDING_PASSWORD_RESET_REQUEST_EXPIRATION_IN_HOURS = 24;
+const PENDING_USER_REGISTRATION_EXPIRATION_IN_HOURS = { hours: 24 };
+const PENDING_PASSWORD_RESET_REQUEST_EXPIRATION_IN_HOURS = { hours: 24 };
```
adrianaluchian, line 14: These 2 guys could now drop the `_IN_HOURS` suffix.

src/components/language-context.js
```diff
@@ -15,9 +15,9 @@ import {
...
 } from '../resources/ui-language.js';
-
-import 'moment';
-import 'moment/locale/de.js';
+import { format, parseISO } from 'date-fns';
+import deDeLocale from 'date-fns/locale/de/index.js';
+import enUsLocale from 'date-fns/locale/en-US/index.js';
 
 const enUS = enUSNs.default || enUSNs;
 const deDE = deDENs.default || deDENs;
```
adrianaluchian, line 23: Gets a little confusing with 2 sets of locale variables here.
An idea would be to maybe make 2 clearly named groups, like
```
const antLocales = {
  enUS: enUSNs.default || enUSNs;
  deDE: deDENs.default || deDENs;
};

const dateLocales = {
 enUS: enUsLocale,
 deDE: deDeLocale,
}
```
What do you think?

adrianaluchian:
Looks good to me. Noticed a tiny thing but I guess it's ok anyway, the comma missing now between date and time.

yarn.lock
ahelmberger: That is here in the lockfile because its a hard dependency of antd during install (in its package.json). I've compiled the bundles though and saw that moment is actually not there anymore, so it is not imported by antd at runtime :+1:

src/components/language-context.js
```diff
@@ -94,15 +94,31 @@ export function useLanguage() {
...
+
+  const dateLocale = locale === 'de-DE' ? deDeLocale : enUsLocale;
+  const localePattern = 'P p';
+  const formatDate = (date, pattern) => date
+    ? format(parseISO(date), pattern || localePattern, { locale: dateLocale })
+    : '';
+
+  return {
+    formatDate
+  };
```
ahelmberger, line 108: Lines 100-108 should be memoized (we will only have two cases), otherwise we create a new `formatDate` function and return object of the hook (which in turn will cause unnecessary re-rendering) on each call of the hook (= on each rerender)

src/repositories/cdn.js
```diff
@@ -53,7 +52,7 @@ class Cdn {
   }
 
   _getDefaultMetadata() {
-    return { createdon: moment.utc().toISOString() };
+    return { createdon: new Date().toISOString() };
```
ahelmberger, line 55: `new Date()` in JS return a local date, not UTC!

src/services/user-service.js
```diff
@@ -117,7 +117,7 @@ class UserService {
       passwordHash: await this._hashPassword(password),
       email: lowerCasedEmail,
       roles,
-      expires: verified ? null : moment.utc().add(PENDING_USER_REGISTRATION_EXPIRATION_IN_HOURS, 'hours').toDate(),
+      expires: verified ? null : add(new Date(), PENDING_USER_REGISTRATION_EXPIRATION_IN_HOURS),
```
ahelmberger, line 120: Same here, it's not UTC

src/services/user-service.js
```diff
@@ -173,7 +173,7 @@ class UserService {
     const request = {
       _id: uniqueId.create(),
       userId: user._id,
-      expires: moment.utc().add(PENDING_PASSWORD_RESET_REQUEST_EXPIRATION_IN_HOURS, 'hours').toDate()
+      expires: add(new Date(), PENDING_PASSWORD_RESET_REQUEST_EXPIRATION_IN_HOURS)
```
ahelmberger, line 176: Same here

src/services/user-service.js
ahelmberger: I realized that we have a `date-time.js` in the utils folder with a function `now` that returns `new Date()`. It seems to be used only once in the document service for the `createdOn` date. This has always been wrongly using the server date instead of UTC, so maybe you could also change that, as you're on the subject anyways? (We're lucky that the Fargate server time is UTC, so all the DB times are correct, but we should make the code location-independent) :-)

src/services/user-service.js
ahelmberger: ...maybe replacing it with `_TIMESPAN` (to make it clearer that it is an object not a scalar value)?

yarn.lock
ahelmberger: The little bit sobering aspect is that with switching from moment.js to date-fns, the bundle size changed from 3.30 MB to 3.28 MB (and we haven't even imported the UTC-relevant package from date-fns yet :-(

adrianaluchian: In this case, is it then worth going through these changes and risk introducing bugs, or better cut our losses and drop this change?

src/services/user-service.js
laurentiu-ilici: Hey, I will. The strange thing is that the new Date() was giving me UTC time in the console, or at least I thought it was.

yarn.lock
laurentiu-ilici: I don't know where you looked for this bundle size. I was looking at what it loads in the production build where the optimisations are applied.

@adrianaluchian: I wouldn't throw in the towel just yet.

src/services/user-service.js
laurentiu-ilici: done

src/services/user-service.js
adrianaluchian: That's true, I hadn't noticed this change 🤦‍♀️ 
In the console you can actually see it's not UTC as the +1 timezone mentioned. If it were UTC the current time would be 08:12.

src/components/language-context.js
```diff
@@ -94,15 +94,31 @@ export function useLanguage() {
   return createLanguageAndLocale(useContext(languageContext));
 }
 
+export function useDateFormat() {
+  const { locale } = useLanguage();
+
+  const dateLocale = locale === 'de-DE' ? deDeLocale : enUsLocale;
+  const localePattern = 'P p';
```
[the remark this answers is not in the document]
laurentiu-ilici: Yep

src/components/language-context.js
laurentiu-ilici: Done

src/components/language-context.js
laurentiu-ilici: Done

yarn.lock
ahelmberger: I compared the size of the js files in dist for the production build, but as I measured only with the file explorer I don't have more accurate numbers. I agree, we should not throw in the towel, small changes combined will also have a huge impact, I was just surprised that it is not that much, actually.

laurentiu-ilici: But isn't what counts what gets delivered to the browser? When I get to the website, I have to load for instance the "view-bundle", that one is what counts or? On the server you also have stuff that only has to do with the server side, or am I missing something?

ahelmberger: OK, these are the "official" numbers, taken directly from the bundler output:

Main branch:
~~~
[10:07:02] Starting 'buildTestAppJs'...
test-app/dist/edit-bundle.js: 457.30 kB
test-app/dist/settings-bundle.js: 40.74 kB
test-app/dist/chunk-RCM2QDPR.js: 1.94 kB
test-app/dist/view-bundle.js: 14.46 kB
test-app/dist/chunk-ZQIMAAA7.js: 674.46 kB
test-app/dist/chunk-M7TF3I6H.js: 1681.64 kB
test-app/dist/midi-53MIUYGM.js: 428.45 kB
test-app/dist/chunk-FJBKNBIP.js: 2.01 kB
TOTAL: 3301.00 kB
~~~

This branch:
~~~
[10:07:52] Starting 'buildTestAppJs'...
test-app/dist/edit-bundle.js: 457.21 kB
test-app/dist/settings-bundle.js: 40.74 kB
test-app/dist/chunk-J2XJBGNO.js: 1.94 kB
test-app/dist/view-bundle.js: 14.35 kB
test-app/dist/chunk-QDZTRKRA.js: 674.46 kB
test-app/dist/chunk-OV5JMYNK.js: 1661.49 kB
test-app/dist/midi-53MIUYGM.js: 428.45 kB
test-app/dist/chunk-FJBKNBIP.js: 2.01 kB
TOTAL: 3280.64 kB
~~~

ahelmberger: So, my explorer was pretty accurate. Anyways, it's 20 kB less and that's good, we just have to continue optimizing IMO.

laurentiu-ilici:
> Looks good to me. Noticed a tiny thing but I guess it's ok anyway, the comma missing now between date and time.

I saw that too but it's the way it formats inside, not sure it's worth the time to try to fix that.

adrianaluchian:
> > Looks good to me. Noticed a tiny thing but I guess it's ok anyway, the comma missing now between date and time.
> 
> I saw that too but it's the way it formats inside, not sure it's worth the time to try to fix that.

Isn't it probably related to what Andreas mentioned here https://github.com/educandu/educandu/pull/21#discussion_r748617426 ?

laurentiu-ilici:
> > > Looks good to me. Noticed a tiny thing but I guess it's ok anyway, the comma missing now between date and time.
> > 
> > 
> > I saw that too but it's the way it formats inside, not sure it's worth the time to try to fix that.
> 
> Isn't it probably related to what Andreas mentioned here [#21 (comment)](https://github.com/educandu/educandu/pull/21#discussion_r748617426) ?

It is, but this is how this library formats the same thing, "L, LT"doesn't exist in this setup

adrianaluchian:
> > > > Looks good to me. Noticed a tiny thing but I guess it's ok anyway, the comma missing now between date and time.
> > > 
> > > 
> > > I saw that too but it's the way it formats inside, not sure it's worth the time to try to fix that.
> > 
> > 
> > Isn't it probably related to what Andreas mentioned here [#21 (comment)](https://github.com/educandu/educandu/pull/21#discussion_r748617426) ?
> 
> It is, but this is how this library formats the same thing, "L, LT"doesn't exist in this setup

But the comma does, doesn't it? 😄

ahelmberger:
> > > > > Looks good to me. Noticed a tiny thing but I guess it's ok anyway, the comma missing now between date and time.
> > > > 
> > > > 
> > > > I saw that too but it's the way it formats inside, not sure it's worth the time to try to fix that.
> > > 
> > > 
> > > Isn't it probably related to what Andreas mentioned here [#21 (comment)](https://github.com/educandu/educandu/pull/21#discussion_r748617426) ?
> > 
> > 
> > It is, but this is how this library formats the same thing, "L, LT"doesn't exist in this setup
> 
> But the comma does, doesn't it? smile

I just wanted to create another grey reply bar :laughing:

laurentiu-ilici:
> > > > > > Looks good to me. Noticed a tiny thing but I guess it's ok anyway, the comma missing now between date and time.
> > > > > 
> > > > > 
> > > > > I saw that too but it's the way it formats inside, not sure it's worth the time to try to fix that.
> > > > 
> > > > 
> > > > Isn't it probably related to what Andreas mentioned here [#21 (comment)](https://github.com/educandu/educandu/pull/21#discussion_r748617426) ?
> > > 
> > > 
> > > It is, but this is how this library formats the same thing, "L, LT"doesn't exist in this setup
> > 
> > 
> > But the comma does, doesn't it? smile
> 
> I just wanted to create another grey reply bar 😆

Whaterfaaaaaaaalll

src/repositories/cdn.js
laurentiu-ilici: Are we fine now like this or?

ahelmberger: Yes

adrianaluchian (approved):

ahelmberger (approved):

PR #23, opened 2021-11-15
EDU-64 Make dependencies more flexible

ahelmberger:


src/components/pages/edit-doc.js
```diff
@@ -54,13 +54,9 @@ class EditDoc extends React.Component {
 
     autoBind(this);
 
-    const { editorFactory, rendererFactory, documentApiClient, initialState } = this.props;
+    const { initialState } = this.props;
     const { documentRevision, proposedSections } = initialState;
 
-    this.editorFactory = editorFactory;
```
laurentiu-ilici, line 60: Why are these changes here? It doesn't seem to be part of the story.

adrianaluchian, line 60: Uh, more leftovers from menus 😬

laurentiu-ilici (approved):

adrianaluchian (approved):

src/components/pages/edit-doc.js
ahelmberger: Linting rules have changed and it now realizes that this line is not used

ahelmberger: Now the good linter tells us more about unused code

PR #24, opened 2021-11-15
EDU-85 Remove aurelia-dependency-injection

ahelmberger:


src/common/di.js
```diff
@@ -1,3 +1,124 @@
-import 'core-js/features/reflect/index.js';
+const emptyList = [];
```
laurentiu-ilici, line 1: This empty list has no value and could be a source for trouble in the future. Why not just use [] where needed (I think there are 2 spots) that way we won't have any future trouble :).

laurentiu-ilici (approved):

src/common/di.js
ahelmberger: I put it there to avoid creating new arrays every time we resolve a dependency (it is only used when resolving it the first time to check for circular dependencies). If you think that it might cause trouble in the future then I'd rather put `null` there so we do not have that cost (it's in the performance critical path, called a lot of times during rendering).

src/common/di.js
```diff
@@ -1,3 +1,124 @@
...
+  _createResolver(key) {
+    if (key === null || typeof key === 'undefined') {
+      throw new Error('key must not be null or undefined');
+    }
+
+    return typeof key === 'function'
+      ? new ServiceResolver(this, key)
+      : new InstanceResolver(key);
+  }
+}
```
adrianaluchian, line 127: Did you borrow all this from aurelia?

src/common/di.js
ahelmberger: Done

src/common/di.spec.js
```diff
@@ -0,0 +1,105 @@
...
+        container.registerService(A, D);
+        container.registerService(B, D);
+
+        const resultA = container.get(A);
+        const resultB = container.get(B);
+
+        await container.dispose();
+
+        expect(resultA).toBe(resultB);
+        expect(resultA.disposedCalled).toBe(1);
```
adrianaluchian, line 99: Maybe for the sake of idempotence we could here do
```
        const resultA = container.get(A);
        const resultB = container.get(B);

        const previousDisposedCalled = resultA.disposedCalled;

        await container.dispose();

        expect(resultA).toBe(resultB);
        expect(resultA.disposedCalled).toBe(previousDisposedCalled + 1)
```

ahelmberger: Sure, done!

laurentiu-ilici (approved):
I have some uneasiness about this PR since we have our own DI container. It looks pretty good though

src/common/di.js
ahelmberger: If so, then only by accident :laughing:

adrianaluchian: Was just wondering how deep should I go trying to understand the logic 😅

adrianaluchian (approved):

PR #28, opened 2021-11-16
EDU-96 Prefill filters on search results

laurentiu-ilici:
Resolves: https://educandu.atlassian.net/browse/EDU-96

### DIFF
--- a/src/components/pages/index.js
+++ b/src/components/pages/index.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import DocView from '../doc-view.js';
 import { Button, Input } from 'antd';
 import SiteLogo from '../site-logo.js';
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useService } from '../container-context.js';
 import { useLanguage } from '../language-context.js';
@@ -21,12 +21,6 @@ function Index({ initialState }) {
   const { document: doc, homeLanguages, currentHomeLanguageIndex } = initialState;
   const currentHomeLanguage = homeLanguages[currentHomeLanguageIndex];
   const [isSearching, setIsSearching] = useState(false);
-
-  useEffect(() => {
-    return () => {
-      setIsSearching(false);
-    };
-  });
 
   const handleSearch = () => {
     setIsSearching(true);
--- a/src/components/pages/index.js
+++ b/src/components/pages/index.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import DocView from '../doc-view.js';
 import { Button, Input } from 'antd';
 import SiteLogo from '../site-logo.js';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useService } from '../container-context.js';
 import { useLanguage } from '../language-context.js';
@@ -21,6 +21,12 @@ function Index({ initialState }) {
   const { document: doc, homeLanguages, currentHomeLanguageIndex } = initialState;
   const currentHomeLanguage = homeLanguages[currentHomeLanguageIndex];
   const [isSearching, setIsSearching] = useState(false);
+
+  useEffect(() => {
+    return () => {
+      setIsSearching(false);
+    };
+  });
 
   const handleSearch = () => {
     setIsSearching(true);
--- a/src/components/pages/search.js
+++ b/src/components/pages/search.js
@@ -28,27 +28,27 @@ function Search({ initialState }) {
     () => docs
       .map(doc => ({
         ...doc,
-        tagsSet: new Set(doc.tags)
+        tagsSet: new Set(doc.tags.map(tag => tag.toLowerCase()))
       }))
       .sort(firstBy(doc => doc.tagMatchCount, 'desc')
         .thenBy(doc => doc.updatedOn, 'desc')),
     [docs]
   );
 
-  const [filteredDocs, setFilteredDocs] = useState([...sortedDocs]);
-
   const sanitizedQueryTags = new Set(decodedQuery.split(/\s/)
     .filter(tag => tag.length > 2)
     .map(tag => tag.toLowerCase()));
 
-  const allTags = docs.map(doc => doc.tags).flat();
+  const allTags = docs.map(doc => doc.tags).flat().map(tag => tag.toLowerCase());
 
   const allUniqueTags = [...new Set(allTags)];
 
   const initialSelectedTags = allUniqueTags
-    .filter(tag => sanitizedQueryTags.has(tag.toLowerCase()));
+    .filter(tag => sanitizedQueryTags.has(tag));
 
   const [selectedTags, setSelectedTags] = useState(initialSelectedTags);
+  const [filteredDocs, setFilteredDocs] = useState([...sortedDocs]
+    .filter(doc => initialSelectedTags.every(tag => doc.tagsSet.has(tag))));
 
   const tagOptions = allUniqueTags.map(tag => ({ value: tag, key: tag }));
 
--- a/src/components/pages/search.js
+++ b/src/components/pages/search.js
@@ -37,26 +37,26 @@ function Search({ initialState }) {
 
   const [filteredDocs, setFilteredDocs] = useState([...sortedDocs]);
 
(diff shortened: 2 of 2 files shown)


ahelmberger:
Test comment

src/components/pages/search.js
```diff
@@ -35,14 +37,27 @@ function Search({ initialState }) {
...
+  }, { allTags: new Set(), tagMap: {} });
+
+  const initialSelectedTags = [...queryTagsToLower]
+    .map(tag => tagMap[tag])
+    .filter(tag => !!tag);
+
+  const [selectedTags, setSelectedTags] = useState(initialSelectedTags);
 
-  const tagOptions = Array.from(allTags).map(tag => ({ value: tag, key: tag }));
+  const tagOptions = [...allTags.map(tag => ({ value: tag, key: tag }));
```
ahelmberger, line 60: test comment

src/components/pages/search.js
```diff
@@ -99,7 +114,7 @@ function Search({ initialState }) {
 
   return (
     <Page headerActions={[]}>
-      <h1>{`${t('searchResultPrefix')}: ${urls.decodeUrl(query.query)}`} </h1>
+      <h1>{`${t('searchResultPrefix')}: ${decodedQuery}`} </h1>
```
ahelmberger, line 117: test 2

src/components/pages/index.js
```diff
@@ -22,6 +22,12 @@ function Index({ initialState }) {
   const currentHomeLanguage = homeLanguages[currentHomeLanguageIndex];
   const [isSearching, setIsSearching] = useState(false);
 
+  useEffect(() => {
+    return () => {
+      setIsSearching(false);
```
laurentiu-ilici, line 27: This works on resetting the page to the false state of searching. I could also reset the input text but I think it's better if I leave it. What do you guys think?

src/components/pages/search.js
```diff
@@ -35,14 +37,26 @@ function Search({ initialState }) {
 
   const [filteredDocs, setFilteredDocs] = useState([...sortedDocs]);
 
-  const [selectedTags, setSelectedTags] = useState([]);
+  const queryTagsToLower = new Set(decodedQuery.split(/\s/)
+    .filter(tag => tag.length > 3)
```
adrianaluchian, line 41: I think here it should be `>= 3`

src/components/pages/search.js
```diff
@@ -35,14 +37,26 @@ function Search({ initialState }) {
...
+    .map(tag => tag.toLowerCase()));
+
+  const { allTags, tagMap } = docs.reduce((acc, doc) => {
+    doc.tags.forEach(tag => {
+      acc.allTags.add(tag);
+      acc.tagMap[tag.toLowerCase()] = tag;
+    });
```
adrianaluchian, line 48: To me this `forEach` in a `reduce` that processes 2 things at the same time looks kind of strange.
I think this construct might become simple/more clear by first getting all tags, then processing them.
Thinking of something like:

```
const allTags = docs.map(doc => doc.allTags).flat();
const allUniqueTags = new Set(allTags);
const tagMap = allTags.reduce(acc, tag => { acc[tag.toLowerCase()] = tag; return acc; }, {});
```

src/components/pages/search.js
```diff
@@ -35,14 +37,26 @@ function Search({ initialState }) {
 
   const [filteredDocs, setFilteredDocs] = useState([...sortedDocs]);
 
-  const [selectedTags, setSelectedTags] = useState([]);
+  const queryTagsToLower = new Set(decodedQuery.split(/\s/)
```
adrianaluchian, line 40: `queryTagsToLowerAndLongerThan3` :)) to be completely accurate.
Might I suggest `sanitizedQueryTags`?

src/components/pages/search.js
adrianaluchian: Actually maybe there's no need to complicate things through this `tagMap` at all. Since it's used just once, I think you can just do this directly:

```
  const initialSelectedTags = allUniqueTags.filter(tag => queryTagsToLower.includes(tag.toLowerCase()));
```
I think this even saves you an iteration, just as trivia info :)

Also I guess `initialSelectedTags `would need to be in the original casing 🤔 The way you obtain them yields you the lowercased tags, but I guess the user needs to see the actual casing of the preselected tags, right? If it's so, then my suggestion covers this.

src/components/pages/index.js
adrianaluchian: Not sure I get why this is needed, since after a search you navigate away.
Could you describe the scenario maybe to clarify?

src/components/pages/search.js
```diff
@@ -35,14 +37,26 @@ function Search({ initialState }) {
...
+  }, { allTags: new Set(), tagMap: {} });
+
+  const initialSelectedTags = [...queryTagsToLower]
+    .map(tag => tagMap[tag])
+    .filter(tag => !!tag);
+
+  const [selectedTags, setSelectedTags] = useState(initialSelectedTags);
 
-  const tagOptions = Array.from(allTags).map(tag => ({ value: tag, key: tag }));
+  const tagOptions = [...allTags].map(tag => ({ value: tag, key: tag }));
```
adrianaluchian, line 59: If you end up changing as I suggested, here you can replace `[...allTags]` with `allUniqueTags` as it's already an array.

src/components/pages/search.js
laurentiu-ilici: The initialSelectedTags are in the casing done by the user, that was the point of that map. Have you tried the code?

Q: I think this even saves you an iteration, just as trivia info :). 
Let's assume that the number of query tags is n and the number of returned tags is m.
The tagMap is there to search in O(1) in the M entries + you have a O(n) because you search each tag once and another O(n) because you do the mapping and another O(n) for "filtering" the tags.
Putting it all together you have my algorithm:  n * O(1) + O(n) + O(n) = O(n) + O(n) + O(n) = 3 * O(n) = O(n)
With the other algo:
you have almost the same but the line:
 allUniqueTags.filter(tag => queryTagsToLower.includes(tag.toLowerCase()));
introduces an O(n * m) since you look for n tags into m elements = O(m*n)

Putting it all together O(m*n) > O(n) always :).

adrianaluchian: TLDR thanks, I didn't take it as far, I said 'I think' 😄

adrianaluchian: "The initialSelectedTags are in the casing exiting in the db which is the casing they were found in there, that was the point of that map. Have you tried the code?"
It's a suggestion, the ultime try has to be yours, the code maker. I think it should work.

src/components/pages/search.js
```diff
@@ -35,14 +37,20 @@ function Search({ initialState }) {
...
-  const [selectedTags, setSelectedTags] = useState([]);
+  const sanitizedQueryTags = new Set(decodedQuery.split(/\s/)
+    .filter(tag => tag.length > 2)
+    .map(tag => tag.toLowerCase()));
+
+  const allTags = docs.map(doc => doc.tags).flat();
+
+  const allUniqueTags = [...new Set(allTags)];
+
+  const initialSelectedTags = [...allUniqueTags]
```
adrianaluchian, line 48: You don't need to spread here and make a new array, `allUniqueTags` is already an array now

laurentiu-ilici: True, thanks.

adrianaluchian:
Question, how do we want to handle this case?

Screenshot 2021-11-17 at 09 51 34

laurentiu-ilici:
> Question, how do we want to handle this case?
> 
> Screenshot 2021-11-17 at 09 51 34

Not sure what to do in this case, the Musik vs musik is treated by the search algorithm as the same tag, but what to display up there is a bit more tricky, unless we agree on something like "display to lower case all the time".

adrianaluchian:
> > Question, how do we want to handle this case?
> > Screenshot 2021-11-17 at 09 51 34
> 
> Not sure what to do in this case, the Musik vs musik is treated by the search algorithm as the same tag, but what to display up there is a bit more tricky, unless we agree on something like "display to lower case all the time".

Maybe we can leave it for now and see if it's ever an issue. From my point of view, if we filter by "tags" then we should actually show the tags as they are, otherwise it's confusing that we show the same component, tag-like, but not the same text.

ahelmberger (approved):

adrianaluchian (approved):

ahelmberger (approved):

PR #29, opened 2021-11-17
EDU-97 Add origin to documents

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-97


src/services/document-service.js
```diff
@@ -3,11 +3,12 @@ import Logger from '../common/logger.js';
 import UserService from './user-service.js';
 import uniqueId from '../utils/unique-id.js';
 import cloneDeep from '../utils/clone-deep.js';
+import escapeStringRegexp from 'escape-string-regexp';
 import DocumentStore from '../stores/document-store.js';
+import { DOCUMENT_ORIGIN } from '../common/constants.js';
 import DocumentLockStore from '../stores/document-lock-store.js';
 import DocumentOrderStore from '../stores/document-order-store.js';
 import DocumentRevisionStore from '../stores/document-revision-store.js';
-import escapeStringRegexp from 'escape-string-regexp';
```
ahelmberger, line 10: :smile:

src/test-helper.js
```diff
@@ -144,7 +145,8 @@ export function createTestDocument(container, user, document) {
       sections: document.sections ?? [],
       tags: document.tags || [],
       appendTo: null,
-      archived: document.archived
+      archived: document.archived,
+      origin: DOCUMENT_ORIGIN.internal
```
ahelmberger, line 149: Could you also for these 2 properties first check if it is defined on `document` and then fall back to the default, we want to be able to override these values.

src/ui/default-prop-types.js
```diff
@@ -152,7 +152,8 @@ export const documentRevisionShape = PropTypes.shape({
   sections: PropTypes.arrayOf(sectionShape).isRequired,
   restoredFrom: PropTypes.string,
   tags: PropTypes.arrayOf(PropTypes.string).isRequired,
-  archived: PropTypes.bool.isRequired
+  archived: PropTypes.bool.isRequired,
+  origin: PropTypes.string
```
ahelmberger, line 156: I think we can make this required, no?

src/services/document-service.js
adrianaluchian: It was hurting my eye 😄

src/ui/default-prop-types.js
adrianaluchian: Yes we can, done.

src/test-helper.js
adrianaluchian: Ah yes, sure, done it for all 3 properties actually. Thanks!

src/components/pages/docs.js
```diff
@@ -206,6 +207,22 @@ class Docs extends React.Component {
...
+  renderOrigin(_value, doc) {
+    const { t } = this.props;
+    const origin = doc.origin || '';
+    let translatedOrigin = origin;
+
+    if (origin === DOCUMENT_ORIGIN.internal) {
+      translatedOrigin = t('originInternal');
+    }
+    if (origin.startsWith(DOCUMENT_ORIGIN.external)) {
+      const nameOfOrigin = origin.split('/')[1];
```
laurentiu-ilici, line 219: You could also split on DOCUMENT_ORIGIN.external and not rely on the "/"being there. If you take the split[1] the "/" should be there already if we didn't misconfigure it (in which case this doesn't work either')

src/services/document-service.js
```diff
@@ -78,16 +80,17 @@ const createNewDocumentRevision = ({ doc, documentKey, userId, nextOrder, restor
...
-    namespace: doc.namespace,
-    language: doc.language,
+    namespace: doc.namespace || '',
+    language: doc.language || '',
     sections: sections || doc.sections,
-    tags: doc.tags,
-    archived: doc.archived || false
+    tags: doc.tags || [],
+    archived: doc.archived || false,
+    origin: doc.origin || DOCUMENT_ORIGIN.internal
```
adrianaluchian, line 93: @ahelmberger Changed as discussed

ahelmberger (approved):

src/services/document-service.js
laurentiu-ilici, line 93: There's a test that ensures the projection has the required fields, maybe you can update that :). It doesn't look for all the stuff that shouldn't be there, just that the sections don't get mapped that's why it didn't fail.

CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @adrianaluchian @laurentiu-ilici @ahelmberger
```
adrianaluchian, line 1: This should work this way, right? 🤔

laurentiu-ilici (approved):

src/components/pages/docs.js
ahelmberger: maybe `origin.match(/^external\/(.+)$/)[1]`? Then the thing after "external/" can even contain another "/" ;-)

CODEOWNERS
ahelmberger: We will see :laughing:

src/components/pages/docs.js
adrianaluchian: I've split it by `DOCUMENT_ORIGIN.external` as Lau suggested, I find this is the simplest less error prone way. We can put `external` followed by whatever we want and only the `external` text will get translated

ahelmberger: Sure (unless the origin name contains "external") :laughing:

laurentiu-ilici: Well, we get put in those names, I hope we don't do that. I'm not really sure why we really need to translate this value anyway, it seems a bit of overkill :)

ahelmberger: I just put this comment everywhere someone uses `string.split` when they actually want to remove a prefix, can't help, I find it the wrong function, sorry ;-)

PR #31, opened 2021-11-17
EDU-98 Add endpoint for importable documents

ahelmberger:


src/server/exchange-controller.js
```diff
@@ -0,0 +1,20 @@
...
+
+class ExchangeController {
+  static get inject() { return [DocumentService]; }
+
+  constructor(documentService) {
+    this.documentService = documentService;
+  }
+
+  registerApi(router) {
+    router.get('/api/v1/exchange/exports', [needsPermission(permissions.LIST_EXPORTABLE_CONTENT)], async (req, res) => {
```
adrianaluchian, line 13: I guess here we need the API key check, right?
Which means the tickets are wrongly blocked.

ahelmberger: Yes, but I made it admin-accessible until we have that

laurentiu-ilici:
Reviewing

src/services/document-service.spec.js
```diff
@@ -312,6 +313,35 @@ describe('document-service', () => {
...
+
+    let result;
+
+    beforeEach(async () => {
+      await Promise.all(testDocs.map(testDoc => createTestDocument(container, user, testDoc)));
+      result = await sut.getAllExportableDocumentsMetadata();
+    });
+
+    it('should return internal unarchived documents', () => {
+      expect(result.find(doc => doc.title === 'doc-1')).not.toBeUndefined();
```
adrianaluchian, line 332: There's `toBeDefined` for this case

laurentiu-ilici: I was also confused by this line :)

laurentiu-ilici (approved):

src/services/document-service.spec.js
ahelmberger: I wanted you to be not unconfused! Nevrtheless, I changed it, thanks :-)

adrianaluchian: That I for one am most surely not sure if I am anymore.

src/server/exchange-controller.js
adrianaluchian: We could merge the API key PR first and do this in one go maybe. Guess that would be easier, or?

ahelmberger: I don't understand, why it would be easier. The API key PR just has to make sure the request is associated with a user that has this permission (either by introducing a new role, or by extending the role <-> permission mapping so that the calling user has `LIST_EXPORTABLE_CONTENT` permission.

src/services/document-service.spec.js
ahelmberger: :exploding_head:

src/server/exchange-controller.js
adrianaluchian: Ok so it's a chicken and egg problem then, whichever PR is second needs to do this correlation 😆 
Then I'll do it in mine.

adrianaluchian (approved):

PR #32, opened 2021-11-17
EDU-102 Add import config

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-102

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,32 @@ The educandu framework
  * Docker
  * gulp installed globally (for conveninence): npm i -g gulp
 
+ ## Configuration options
+
+ | Option | Description | Type | Required |
+ | --- | --- | --- | --- |
+ | env | Environment on which the project is run | `string` | yes |
+ | port | Port on which the project is run | `number`, mininum 1 | yes |
+ | mongoConnectionString | The URI for the project's MongoDB | `string` | yes |
+ | skipMongoMigrations | Whether or not to run newly added MongoDB migration scripts on startup | `boolean` | yes |
+ | skipMongoChecks | Whether or not to run MongoDB checks resulting in ensuring DB collections on startup | `boolean` | yes |
+ | cdnEndpoint | The URL of the AWS-hosted CDN | `string` | yes |
+ | cdnRegion | The region of the AWS-hosted CDN | `string` | yes |
+ | cdnAccessKey | The access key of the AWS-hosted CDN | `string` | yes |
+ | cdnSecretKey | The secret key of the AWS-hosted CDN | `string` | yes |
+ | cdnBucketName | The name of the AWS S3 bucket storing the CDN data | `string` | yes |
+ | cdnRootUrl | The root url of the CDN | `string` | yes |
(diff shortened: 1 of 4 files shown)


src/bootstrap/server-config.js
```diff
@@ -17,15 +17,21 @@ const configSchema = joi.object({
   sessionSecret: joi.string(),
   sessionDurationInMinutes: joi.number().min(1),
   smtpOptions: joi.any(),
-  emailSenderAddress: joi.string(),
+  emailSenderAddress: joi.string().require(),
```
adrianaluchian, line 20: I made this required as I think it kind of is, isn't it?
Shall we keep it and I fix the texts or revert?

ahelmberger: It's good to have it, I think

adrianaluchian: That might explain some things 😂 Thanks

README.md
```diff
@@ -43,7 +70,13 @@ educandu({
     password: 'test',
     email: 'test@test.com'
   },
-  exposeErrorDetails: true //set to false in order to hide error details such as stack trace on productive systems
+  exposeErrorDetails: true,
+  importApiKey: 'GSD54GDFgDGesdfs4'
+  importSources: {
```
laurentiu-ilici, line 75: Shouldn't this be an array?

src/bootstrap/server-config.js
```diff
@@ -17,15 +17,21 @@ const configSchema = joi.object({
...
   resources: joi.array().items(joi.string()),
   initialUser: joi.object({
     username: joi.string().required(),
     password: joi.string().required(),
     email: joi.string().required()
   }).allow(null),
-  exposeErrorDetails: joi.boolean().required()
+  exposeErrorDetails: joi.boolean().required(),
+  importApiKey: joi.string(),
+  importSources: joi.array().items(joi.object({
```
laurentiu-ilici, line 30: Yep, it should :).

src/bootstrap/server-config.js
```diff
@@ -17,15 +17,21 @@ const configSchema = joi.object({
...
+    name: joi.string().required(),
+    baseUrl: joi.string().required(),
+    apiKey: joi.string().required()
+  }))
```
laurentiu-ilici, line 34: I'm now a bit confused, since we said it could be optional, shouldn't we allow here .allow(null).

src/bootstrap/server-config.js
adrianaluchian: Corrected, thanks.

src/bootstrap/server-config.js
adrianaluchian: Well to me `allow(null)` would make sense when it's mandatory, so that the user has to willingly provide something, even if that something is `null`.
The import properties are optional as they are specified now.

adrianaluchian:
Will introduce usage of the `importApiKey` after we merge https://github.com/educandu/educandu/pull/31

ahelmberger (approved):

laurentiu-ilici (approved):

PR #33, opened 2021-11-17
EDU-103 Do a full page refresh (without using the cache) when home page is loaded from cache

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-103

This works in Chrome, Firefox and Safari on MacOS.

### DIFF
--- a/src/components/pages/index.js
+++ b/src/components/pages/index.js
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import DocView from '../doc-view.js';
 import { Button, Input } from 'antd';
 import SiteLogo from '../site-logo.js';
-import React, { useState } from 'react';
 import { useTranslation } from 'react-i18next';
+import React, { useState, useEffect } from 'react';
 import { useService } from '../container-context.js';
 import { useLanguage } from '../language-context.js';
 import { getHomeUrl, getSearchUrl } from '../../utils/urls.js';
@@ -28,7 +28,7 @@ function Index({ initialState }) {
     }
   };
 
-  React.useEffect(() => {
+  useEffect(() => {
     window.addEventListener('pageshow', handlePageShow);
 
     return () => {
--- a/src/components/pages/index.js
+++ b/src/components/pages/index.js
@@ -22,6 +22,20 @@ function Index({ initialState }) {
   const currentHomeLanguage = homeLanguages[currentHomeLanguageIndex];
   const [isSearching, setIsSearching] = useState(false);
 
+  const handlePageShow = event => {
+    if (event.persisted) {
+      window.location.reload(true);
+    }
+  };
+
+  React.useEffect(() => {
+    window.addEventListener('pageshow', handlePageShow);
+
+    return () => {
+      window.removeEventListener('pageshow', handlePageShow);
+    };
+  }, []);
+
   const handleSearch = () => {
     setIsSearching(true);
     window.location = getSearchUrl(searchText.trim());


ahelmberger:
Works on Chrome, Firefox and Edge on Linux :+1:

ahelmberger (approved):

src/components/pages/index.js
```diff
@@ -22,6 +22,20 @@ function Index({ initialState }) {
   const currentHomeLanguage = homeLanguages[currentHomeLanguageIndex];
   const [isSearching, setIsSearching] = useState(false);
 
+  const handlePageShow = event => {
+    if (event.persisted) {
+      window.location.reload(true);
+    }
+  };
+
+  React.useEffect(() => {
```
laurentiu-ilici, line 31: Why not import with import React, { useState, useEffect } from 'react'?

src/components/pages/index.js
```diff
@@ -22,6 +22,20 @@ function Index({ initialState }) {
   const currentHomeLanguage = homeLanguages[currentHomeLanguageIndex];
   const [isSearching, setIsSearching] = useState(false);
 
+  const handlePageShow = event => {
+    if (event.persisted) {
+      window.location.reload(true);
```
laurentiu-ilici, line 27: This would reset everything, don't we want to keep what the user typed in? You can do that by setIsSearching(false). This way I guess we do a round trip to the server.

src/components/pages/index.js
```diff
@@ -22,6 +22,20 @@ function Index({ initialState }) {
...
+    }
+  };
+
+  React.useEffect(() => {
+    window.addEventListener('pageshow', handlePageShow);
+
+    return () => {
+      window.removeEventListener('pageshow', handlePageShow);
```
laurentiu-ilici, line 35: I have some feeling about this not working as it should let me check please.

adrianaluchian: That's what the PR is here for

src/components/pages/index.js
adrianaluchian: No specific reason, maybe habit, changed to named import as both are the same for me.

src/components/pages/index.js
laurentiu-ilici: Yep, this doesn't get fired at all, but I don't understand why the event listener is not registered on the other pages. Hmmmmmmmmm

adrianaluchian: Why would it be on the other pages, when it's deregistered when you leave this page? 🤷

laurentiu-ilici: The problem is that it doesn't deregister, it works the same if you don't do anything

ahelmberger: It doesn't as we call `document.location`, so the component is never unmounted, nevertheless it's good practice to hve it there, I find

src/components/pages/index.js
```diff
@@ -22,6 +22,20 @@ function Index({ initialState }) {
...
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener('pageshow', handlePageShow);
+
+    return () => {
+      window.removeEventListener('pageshow', handlePageShow);
+    };
+  }, []);
```
laurentiu-ilici, line 37: You need to remove this empty dependency array, else it doesn't do the cleanup. I'm just surprised that the pageshow event doesn't trigger on the other pages. It's probably because it's a new page from the server side.

laurentiu-ilici (approved):

src/components/pages/index.js
adrianaluchian: Ok, true, this doesn't get called in this case. But it gets "detached" by default with the reload anyway.

src/components/pages/index.js
adrianaluchian: With the empty array we want only do this on the initial render, we don't want to add the listener with every re-render.

laurentiu-ilici: Fair, but then you can remove the de-registration, it doesn't get called at all.

adrianaluchian: Like Andreas said here https://github.com/educandu/educandu/pull/33#discussion_r751395632 I'd keep this as good practice

laurentiu-ilici: ok

PR #34, opened 2021-11-18
EDU-99 Collect CDN resources

ahelmberger:


src/plugins/ear-training/info.js
```diff
@@ -11,12 +12,13 @@ export default {
...
+          type: SOUND_TYPE.midi,
           url: null,
           text: null
         }
       }
     ]
   }),
-  cloneContent: content => cloneDeep(content)
+  cloneContent: content => cloneDeep(content),
+  getCdnResources: content => content.tests.filter(test => test.sound?.type === SOUND_TYPE.internal && test.sound.url).map(test => test.sound.url)
```
adrianaluchian, line 23: Small thing: can `content.tests` be `null`/`undefined` or is it always `[]`?

src/plugins/image-tiles/info.js
```diff
@@ -9,5 +10,6 @@ export default {
     maxWidth: 100,
     hoverEffect: 'none'
   }),
-  cloneContent: content => cloneDeep(content)
+  cloneContent: content => cloneDeep(content),
+  getCdnResources: content => content.tiles.filter(tile => tile.image?.type === IMAGE_TYPE.internal && tile.image.url).map(tile => tile.image.url)
```
adrianaluchian, line 14: Same question here about `content.tiles`

src/services/document-service.spec.js
```diff
@@ -23,6 +27,42 @@ describe('document-service', () => {
...
+            type: 'image',
+            content: {
+              sourceType: SOURCE_TYPE.internal,
+              sourceUrl: 'media/image-1.png',
+              effect: {
+                sourceType: SOURCE_TYPE.internal,
+                sourceUrl: 'media/image-2.png'
+              }
+            }
+          }
```
adrianaluchian, line 50: Maybe overkill, but how about just adding a second section here, only to have the test include/prove the fact that all sections are iterated?

src/plugins/ear-training/info.js
ahelmberger: It's always `[]`, created by the default content.

src/plugins/image-tiles/info.js
ahelmberger: Same case here

src/services/document-service.spec.js
ahelmberger: OK, good idea

adrianaluchian:
Nice reorganizing and testing, enjoyed reviewing this, thanks 🙇

migrations/educandu-2021-11-18-01-add-document-cdn-resources.js
```diff
@@ -0,0 +1,16 @@
+// eslint-disable-next-line camelcase
+export default class Educandu_2021_11_18_01_add_document_cdn_resources {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    await this.db.collection('documentRevisions').updateMany({}, { $set: { cdnResources: [] } });
```
laurentiu-ilici, line 8: Hmm I named these cdnResourceUrls, I'll rename to your variant.

src/plugins/anavis/editing/anavis-editor.js
```diff
@@ -19,33 +20,6 @@ const RadioButton = Radio.Button;
 const TextArea = Input.TextArea;
 const ButtonGroup = Button.Group;
 
-const KIND_AUDIO = 'audio';
-const KIND_VIDEO = 'video';
-
-const TYPE_EXTERNAL = 'external';
-const TYPE_INTERNAL = 'internal';
```
laurentiu-ilici, line 26: Nice refactor!

laurentiu-ilici (approved):

migrations/educandu-2021-11-18-01-add-document-cdn-resources.js
ahelmberger: I think the tickets say `cdnResources` :thinking:

src/plugins/anavis/editing/anavis-editor.js
ahelmberger: Thanks!

ahelmberger:
> Please account for "null" content on all plugin-info, deleted sections will have this problem. Example from staging: [2c9v4xfqtyEZaxtjgddtnd](https://staging.elmu.online/docs/2c9v4xfqtyEZaxtjgddtnd)

Good catch, thanks! I'll change.

migrations/educandu-2021-11-18-01-add-document-cdn-resources.js
laurentiu-ilici: Yep, my bad

laurentiu-ilici:
> > Please account for "null" content on all plugin-info, deleted sections will have this problem. Example from staging: [2c9v4xfqtyEZaxtjgddtnd](https://staging.elmu.online/docs/2c9v4xfqtyEZaxtjgddtnd)
> 
> Good catch, thanks! I'll change.

It wasn't really a catch :)). My scripts failed on the migration because of this (aka made the same mistake)

ahelmberger:
> > > Please account for "null" content on all plugin-info, deleted sections will have this problem. Example from staging: [2c9v4xfqtyEZaxtjgddtnd](https://staging.elmu.online/docs/2c9v4xfqtyEZaxtjgddtnd)
> > 
> > 
> > Good catch, thanks! I'll change.
> 
> It wasn't really a catch :)). My scripts failed on the migration because of this (aka made the same mistake)

Done!

src/services/document-service.spec.js
ahelmberger: Done!

adrianaluchian (approved):

src/services/document-service.js
```diff
@@ -416,9 +418,24 @@ class DocumentService {
...
-      origin: lastRevision.origin
+      origin: lastRevision.origin,
+      cdnResources: lastRevision.cdnResources
     };
   }
+
+  _getCdnResources(sections) {
+    return [
+      ...sections.reduce((cdnResources, section) => {
+        const info = getPluginInfoByType(section.type);
```
laurentiu-ilici, line 429: But shouldn't we fix this on each plug-in info file? What guarantees that that function will always be used only here and that we will remember about the faulty null content trouble? I'd consider that the getCdnResources is still buggy or?

ahelmberger: As soon as it's deleted, a section is not handed down to the plugin code anymore (e.g. `display`, `renderer`, `cloneContent`), so the idea was to stick to that. But I'm also not super sure about it.

laurentiu-ilici: My question is how are you sure that this will only be used here in the future? If we use this thing somewhere else we might forget that the content might be null...

ahelmberger: We can't be sure, but calling a plugin info function on a deleted section I would consider an error in the business logic (see also the `cloneContent` function in the same file, that won't work either with deleted sections). Nevertheless, I'll change as it's not gonna hurt. I'd like to leave the check for "deleted-ness" in the service though.

laurentiu-ilici: Ok, I don't have a problem with the check. Maybe then it's better to leave it as it is now so we get a problem rather than hiding it.

laurentiu-ilici (approved):

src/services/document-service.js
ahelmberger: That's also a good point, may rather have the error when you're not supposed to call it. As I said, I wasn't sure about this either. If you're ok with it, I'd leave it like it is for now.

PR #35, opened 2021-11-18
EDU-100-migrate-cdn-resources

laurentiu-ilici:
I'm still testing this locally, it's more of a chance to look at it. First tests seem to work

### DIFF
--- a/migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
+++ b/migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
@@ -90,12 +90,12 @@ export default class Educandu_2021_11_17_02_migrate_cdn_resources {
 
     await updateAll(this.db.collection('documents'), {}, (doc) => {
       const relevantSections = doc.sections.filter(section => pluginsWithCdnResources.has(section.type));
-      doc.cdnResourceUrls = [...new Set(aggregateSectionUrls(relevantSections))];
+      doc.cdnResourceUrls = [...new Set(aggregateSectionUrls(relevantSections).filter(url => !!url))];
     });
 
     await updateAll(this.db.collection('documentRevisions'), {}, (doc) => {
       const relevantSections = doc.sections.filter(section => pluginsWithCdnResources.has(section.type));
-      doc.cdnResourceUrls = [...new Set(aggregateSectionUrls(relevantSections))];
+      doc.cdnResourceUrls = [...new Set(aggregateSectionUrls(relevantSections).filter(url => !!url))];
     });
   }
 
--- a/migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
+++ b/migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
@@ -24,7 +24,7 @@ const processVideoSection = ({content}) =>  content.sourceType === INTERNAL_URL_
 
 const processImageTilesSection = ({content}) => {
   return content.tiles?.reduce((acc, tile) => {
-    if (tile.image.type === INTERNAL_URL_TYPE) {
+    if (tile.image?.type === INTERNAL_URL_TYPE) {
       acc.push(tile.image.url);
     }
 
@@ -35,7 +35,7 @@ const processImageTilesSection = ({content}) => {
 const processEarTrainingSection = ({content}) => {
   return content.tests?.reduce((acc, test) => {
     if(test.sound?.type === INTERNAL_URL_TYPE) {
-      acc.push(test.sound?.url);
+      acc.push(test.sound.url);
     }
 
     return acc;
--- a/migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
+++ b/migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
@@ -1,0 +1,106 @@
+// eslint-disable-next-line camelcase
+
+const pluginsWithCdnResources = new Set([
+  'image',
+  'video',
+  'image-tiles',
+  'audio',
+  'anavis',
+  'ear-training'
+]);
+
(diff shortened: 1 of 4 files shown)


migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
```diff
@@ -0,0 +1,106 @@
+// eslint-disable-next-line camelcase
+
```
adrianaluchian, line 2: Could you rename this file to `migrations/educandu-2021-11-18-02-migrate-cdn-resoures.js` so that it comes after `migrations/educandu-2021-11-18-01-add-document-cdn-resources.js` which is added here https://github.com/educandu/educandu/pull/34/files ?
Otherwise they are run in the wrong order.

adrianaluchian:
> I'm still testing this locally, it's more of a chance to look at it. First tests seem to work

Let's maybe prefix the ticket name with [Draft] then and we should not approve a draft.
The ticket owner then needs to "undraft" it when it's ready for the final review and approval.

migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
laurentiu-ilici: Ok

laurentiu-ilici: Done :)

migrations/educandu-2021-11-17-02-migrate-cdn-resoures.js
```diff
@@ -0,0 +1,106 @@
...
+  'video',
+  'image-tiles',
+  'audio',
+  'anavis',
+  'ear-training'
+]);
+
+const INTERNAL_URL_TYPE = 'internal';
+const processImageSection = ({content}) => {
+  const urls = content?.sourceType === INTERNAL_URL_TYPE ? [content.sourceUrl] : [];
```
ahelmberger: The type could be 'internal', but the url could be null (in all plugins)

laurentiu-ilici: They are filtered out before getting saved.

ahelmberger: I see, nevermind

laurentiu-ilici:
> > I'm still testing this locally, it's more of a chance to look at it. First tests seem to work
> 
> Let's maybe prefix the PR name with [Draft] then and we should not approve a draft. The ticket owner then needs to "undraft" it when it's ready for the final review and approval.

It's reviewable :)

adrianaluchian:
Idea:
We could postfix maybe such migration scripts with something like `-manually-run` (so `educandu-2021-11-17-02-migrate-cdn-resoures-manually-run`) and exclude files with this postfix from automatic run.
Cause it would still be nice to have them in the same place and in the correct logical order, but have them skipped in automatic run.

ahelmberger:
> Idea: We could postfix maybe such migration scripts with something like `-manually-run` (so `educandu-2021-11-17-02-migrate-cdn-resoures-manually-run`) and exclude files with this postfix from automatic run. Cause it would still be nice to have them in the same place and in the correct logical order, but have them skipped in automatic run.

:+1: for this idea.

ahelmberger (approved):

laurentiu-ilici:
> Idea: We could postfix maybe such migration scripts with something like `-manually-run` (so `educandu-2021-11-17-02-migrate-cdn-resoures-manually-run`) and exclude files with this postfix from automatic run. Cause it would still be nice to have them in the same place and in the correct logical order, but have them skipped in automatic run.

good idea, done!

ahelmberger (approved):

adrianaluchian:
> > Idea: We could postfix maybe such migration scripts with something like `-manually-run` (so `educandu-2021-11-17-02-migrate-cdn-resoures-manually-run`) and exclude files with this postfix from automatic run. Cause it would still be nice to have them in the same place and in the correct logical order, but have them skipped in automatic run.
> 
> good idea, done!

Can you please also do the "exclude files with this postfix from automatic run" part of my comment? 😅 Otherwise this makes no difference.

src/stores/database.js
```diff
@@ -94,6 +94,7 @@ class Database {
     const allFilesInMigrationDirectory = await pGlob(path.resolve(migrationsDirectory, './*.js'));
     const migrationFileNames = allFilesInMigrationDirectory
       .filter(fileName => MIGRATION_FILE_NAME_PATTERN.test(path.basename(fileName)))
+      .filter(filename => !filename.includes('manually'))
```
adrianaluchian, line 97: Why not put the whole suffix here? You can say `!filename.endsWith('-manually-run')`

ahelmberger: Yeah, that would be much safer

laurentiu-ilici: I considered this safer since you rely on manually and not -manually-run (which is more restrictive). I'll change it though

adrianaluchian: For me it makes no sense set a `-manually-run` postfix and have a rule that doesn't match.
And I think in this situation it's better to accidentally run a script that's made for manual run (which might simply run too long) than to skip running it unintentionally and create DB inconsistency.
Our manually run scripts should anyway be re-runnable.

laurentiu-ilici: Done

adrianaluchian: Pssst, please make it consistent with the rule we agreed on, `!filename.endsWith('-manually-run')`, I already mentioned it here https://github.com/educandu/educandu/pull/35#discussion_r752381844 🙏

laurentiu-ilici: Aye aye, done

adrianaluchian (approved):

PR #36, opened 2021-11-18
EDU-108 Fix duplicate key issue

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-108

### DIFF
--- a/src/services/document-service.spec.js
+++ b/src/services/document-service.spec.js
@@ -324,7 +324,10 @@ describe('document-service', () => {
     let result;
 
     beforeEach(async () => {
-      await Promise.all(testDocs.map(testDoc => createTestDocument(container, user, testDoc)));
+      for (const testDoc of testDocs) {
+        /* eslint-disable-next-line no-await-in-loop */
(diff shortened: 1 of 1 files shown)


src/services/document-service.spec.js
```diff
@@ -324,7 +324,10 @@ describe('document-service', () => {
     let result;
 
     beforeEach(async () => {
-      await Promise.all(testDocs.map(testDoc => createTestDocument(container, user, testDoc)));
+      for (const testDoc of testDocs) {
```
laurentiu-ilici, line 327: I'm not sure that this solves the root cause though...

laurentiu-ilici (approved):

src/services/document-service.spec.js
adrianaluchian: Please read the Jira ticket first, I tried to make it clear there what the line of investigation was and why this fix is so.

laurentiu-ilici: It's clear, thanks!

ahelmberger (approved):

PR #37, opened 2021-11-18
Feature/edu 88 adaptible styles

laurentiu-ilici:


ahelmberger:
Nice :+1:  (I just tried it in green)

ahelmberger (approved):

test-app/main.less
```diff
@@ -1 +1,21 @@
 @import url('../dist/styles/main.less');
+
+//Example font family override
+@global-font-family: 'Roboto', sans-serif;
+
+//Example header size overrides
+//@global-h1-font-size: 5em;
+//@global-h2-font-size: 2.5em;
```
adrianaluchian, line 8: Let's remove commented out code please 🙏

ahelmberger: Or make this THE code, so we have some more obvious overrides in the test app?

laurentiu-ilici: I left this here as a way to document and an easy way for us to try out. The problem with uncommenting this all the time is that we lose how elmu really looks  like. I'll make it the code.

ahelmberger (approved):

adrianaluchian (approved):

PR #38, opened 2021-11-18
EDU-106 Enable API key authentication

ahelmberger:
Resolves [EDU-106](https://educandu.atlassian.net/browse/EDU-106)

### DIFF
--- a/src/domain/permissions.spec.js
+++ b/src/domain/permissions.spec.js
@@ -1,11 +1,8 @@
 import { ROLE } from './role.js';
 import { exchangeUser } from './built-in-users.js';
-import sut, { LIST_EXPORTABLE_CONTENT } from './permissions.js';
+import permissions, { hasUserPermission } from './permissions.js';
 
-const allNonTechnicalPermissions = Object.entries(sut)
-  .filter(entry => (/^[A-Z_]+$/).test(entry[0]))
-  .filter(entry => entry[1] !== LIST_EXPORTABLE_CONTENT)
-  .map(entry => entry[1]);
+const isTechnicalPermission = permission => permission === permissions.LIST_EXPORTABLE_CONTENT;
 
 describe('permissions', () => {
 
@@ -16,18 +13,16 @@ describe('permissions', () => {
         roles: [ROLE.admin]
       };
 
-      allNonTechnicalPermissions.forEach(permission => {
+      Object.values(permissions).forEach(permission => {
         let result;
+        let expected;
         beforeEach(() => {
-          result = sut.hasUserPermission(adminUser, permission);
+          result = hasUserPermission(adminUser, permission);
+          expected = !isTechnicalPermission(permission);
         });
-        it(`should be true for permission '${permission}'`, () => {
-          expect(result).toBe(true);
+        it(`should be ${expected} for permission '${permission}'`, () => {
+          expect(result).toBe(expected);
         });
-      });
-
-      it(`should be false for permission '${LIST_EXPORTABLE_CONTENT}'`, () => {
-        expect(sut.hasUserPermission(adminUser, LIST_EXPORTABLE_CONTENT)).toBe(false);
       });
 
     });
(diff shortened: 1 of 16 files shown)


src/server/user-controller.js
```diff
@@ -52,8 +54,15 @@ class UserController {
 
     router.use(passport.initialize());
     router.use(passport.session());
+    router.use(passport.authenticate('apikey', { session: false }));
 
-    passport.use(new LocalStrategy((username, password, cb) => {
+    passport.use('apikey', new ApiKeyStrategy((apikey, cb) => {
+      return apikey === this.serverConfig.importApiKey
```
laurentiu-ilici, line 60: To me it seems like this is authorizing an export since we have the permission "LIST_EXPORTABLE_CONTENT"

test-app/index.js
```diff
@@ -17,6 +17,7 @@ educandu({
   smtpOptions: 'smtp://localhost:8025/?ignoreTLS=true',
   publicFolders: ['./test-app/dist', './test-app/static'].map(x => path.resolve(x)),
   resources: ['./test-app/resource-overrides.json'].map(x => path.resolve(x)),
+  importApiKey: 'fe160daddb0c44c4963f63ce08272c86',
```
laurentiu-ilici, line 20: I have the feeling this should be renamed to exportAPIKey since this is the key used for a system that imports the data to view what he can import.
Example
elmu is the origin : I would expect this to have an exportAPIKey usable by all importers to get the list of content
oma is the importer: I would expect this to have an importAPIKey to be used when importing from other systems

I know we said we use one API key to rule them all though. I'm just still confused with the naming.

ahelmberger: True, I renamed it to `exportApiKey`

src/server/user-controller.js
ahelmberger: That makes more sense, yes. I renamed the config key to `exportApiKey`

test-app/index.js
adrianaluchian: To me it's more confusing to have two names for the exact same key 😄 Buuut anyway

src/domain/permissions.spec.js
```diff
@@ -10,6 +10,7 @@ describe('permissions', () => {
 
     const permissions = Object.entries(sut)
       .filter(entry => (/^[A-Z_]+$/).test(entry[0]))
+      .filter(entry => entry[1] !== LIST_EXPORTABLE_CONTENT)
```
adrianaluchian, line 13: Could we maybe also have a test for the case when this permission **is** set?

ahelmberger: Ha :laughing: right

ahelmberger: Done!

test-app/index.js
ahelmberger: Maybe we want to drop the names "import" and "export" completely? I'm also not super happy with the terms. But that would be a discussion out of scope for this PR

adrianaluchian: I think the direction of the exchange needs to be clear. But yeah, out of scope, we can see how working with this feels and if we keep on being confused by the terminology we can revisit and rename anytime anyway

src/domain/permissions.spec.js
```diff
@@ -1,25 +1,54 @@
 import { ROLE } from './role.js';
-import sut from './permissions.js';
+import { exchangeUser } from './built-in-users.js';
+import sut, { LIST_EXPORTABLE_CONTENT } from './permissions.js';
+
+const allNonTechnicalPermissions = Object.entries(sut)
+  .filter(entry => (/^[A-Z_]+$/).test(entry[0]))
+  .filter(entry => entry[1] !== LIST_EXPORTABLE_CONTENT)
```
adrianaluchian, line 7: Oh in this case you could even have a `allTechnicalPermissions` list with just this guy and use the list both in this filtering and in line 29 (which can also be an iteration over `allTechnicalPermissions`) 😎

adrianaluchian: Unless this is taking it too far, your call anyway 😄

adrianaluchian (approved):

src/domain/permissions.spec.js
ahelmberger: I unified the exports of `permissions.js` and reworked this test to go through all exported permissions

adrianaluchian: Ha, cool!

laurentiu-ilici (approved):

adrianaluchian (approved):

PR #39, opened 2021-11-19
EDU-75 introduce a changelog

laurentiu-ilici:


gulpfile.js
```diff
@@ -376,6 +378,50 @@ export function restartServer(done) {
...
+    const lineReader = readline.createInterface({
+      input: fileStream,
+      crlfDelay: Infinity
+    });
+
+    for await (const line of lineReader) {
+      let jiraLine = null;
+      if ((/^- EDU-\d+/).test(line)) {
+        const jiraNumber = line.match(/EDU-\d+/);
+        jiraLine = line.replace(/^- EDU-\d+/, `- [${jiraNumber}](https://educandu.atlassian.net/browse/${jiraNumber})`);
```
ahelmberger, line 399: If you want to replace all occurrences of JIRA numbers within one line (e.g. in smth. like "- EDU-100: Revert changes from EDU-99", you could use the callback of `String.prototype.replace`:

~~~js
jiraLine = line.replace(/EDU-\d+/g, num => `[${num}](https://educandu.atlassian.net/browse/${num})`);
~~~

You could also just run this for each line (no need for the `if` and the `line.match`), thus simplifying the code.

gulpfile.js
```diff
@@ -376,6 +378,50 @@ export function restartServer(done) {
...
+    writer = createWriteStream('CHANGELOG.md');
+    for (const line of newLines) {
+      writer.write(line);
+      writer.write(EOL);
+    }
+    writer.end();
+  } finally {
+    writer && writer.close();
+  }
+}
```
ahelmberger, line 421: Just out of curiosity, why not simply `fs.readFile` and `fs.writeFile`, the whole function would then just be:

~~~
export async function generateJiraLinks() {
  const changeLog = await fs.readFile('CHANGELOG.md', 'utf8');
  const replacedChangeLog = changeLog.replace(/EDU-\d+/g, num => `[${num}] 
 (https://educandu.atlassian.net/browse/${num})`);
}
await fs.writeFile('CHANGELOG.md', replacedChangeLog, 'utf8');
~~~

laurentiu-ilici: Several reasons:
 1) I was doing the first thing that comes to mind to try it out.
 2) I didn't work with the file system in node up to now, and I wanted to process each line since it seems like it replaces the whole file (at least that is the behavior when we don't have tags). This would mean that the second time you run your script this would replace the EDU-d everywhere again, which would lead to broken links. If you look at my regex, it looks for exact matching start of line.
If you know a better way, shoot.

gulpfile.js
laurentiu-ilici: Ok

PR #46, opened 2021-11-24
EDU-75 introduce a changelog

laurentiu-ilici:
Let's see how it works like this. I might have to do a change where I delete the previous release all the time, but I'm not sure yet. We will know once we have several releases.


.github/workflows/ci.yml
```diff
@@ -62,6 +62,19 @@ jobs:
...
+        run: sudo gem install github_changelog_generator
+      - name: Export github token
+        run: export CHANGELOG_GITHUB_TOKEN=${{secrets.GITHUB_TOKEN}}
+      - name: Build release notes
+        run:  ./node_modules/.bin/gulp generateChangelog
+      - name: Write github token to file
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" > gitToken.txt
+      - name: Authenticate for changelog
+        run: gh auth login --with-token < gitToken.txt
```
ahelmberger, line 75: I don't think you have to write the token to a file in order to read it in again. Shouldn't this work:

~~~
echo "${{ secrets.GITHUB_TOKEN }}" | gh auth login --with-token
~~~
?

laurentiu-ilici: Hmm, not sure, let me try

laurentiu-ilici: That worked, thanks!

laurentiu-ilici:
As a general note, it seems like the gh create release ... also puts a tag on the code. I guess this will be okay when we use our own tags and do the real releases. For now this will create a tag for each prerelease. Let's see how it works out, as I said I might have to delete the last release always.

.github/workflows/ci.yml
```diff
@@ -62,6 +64,17 @@ jobs:
         run: yarn install --non-interactive --check-files --frozen-lockfile
       - name: Build library
         run: ./node_modules/.bin/gulp build
+      - name: Install github changelog generator
+        run: sudo gem install github_changelog_generator
+      - name: Export github token
+        run: export CHANGELOG_GITHUB_TOKEN=${{secrets.GITHUB_TOKEN}}
```
ahelmberger, line 70: I think this variable is not used anywhere, is it (also on the release task)?

ahelmberger:
> As a general note, it seems like the gh create release ... also puts a tag on the code. I guess this will be okay when we use our own tags and do the real releases. For now this will create a tag for each prerelease. Let's see how it works out, as I said I might have to delete the last release always.

I think I don't really understand that. You mean `gh` pushes another tag to the repo? Why will we have to delete a release each time we create a release?

laurentiu-ilici:
> > As a general note, it seems like the gh create release ... also puts a tag on the code. I guess this will be okay when we use our own tags and do the real releases. For now this will create a tag for each prerelease. Let's see how it works out, as I said I might have to delete the last release always.
> 
> I think I don't really understand that. You mean `gh` pushes another tag to the repo? Why will we have to delete a release each time we create a release?

 I think I don't really understand that. You mean `gh` pushes another tag to the repo? 
This gh cli seems to push a tag when it creates a release if there isn't one. 

Why will we have to delete a release each time we create a release?
Because it seems to me like the latest release note was going at the bottom (meaning you'd have to scroll to see it), but I'm not sure yet if that was only because I confused it with my fake tags or what was the cause.

.github/workflows/ci.yml
laurentiu-ilici: It is used by the github_changelog_generator in the gulp task.

ahelmberger: Ah, OK, thanks

ahelmberger:
> I think I don't really understand that. You mean `gh` pushes another tag to the repo? This gh cli seems to push a tag when it creates a release if there isn't one.

Ah, ok. We can switch to "real" releases anytime, if you ask me. Then this problem will be gone.

> Why will we have to delete a release each time we create a release? Because it seems to me like the latest release note was going at the bottom (meaning you'd have to scroll to see it), but I'm not sure yet if that was only because I confused it with my fake tags or what was the cause.

I can confirm that Github shows the releases in descending order (at least that's how it always worked for me). If we would have to delete a release each time we create a release, I would question the process ;-)

ahelmberger (approved):

laurentiu-ilici:
> > I think I don't really understand that. You mean `gh` pushes another tag to the repo? This gh cli seems to push a tag when it creates a release if there isn't one.
> 
> Ah, ok. We can switch to "real" releases anytime, if you ask me. Then this problem will be gone.
> 
> > Why will we have to delete a release each time we create a release? Because it seems to me like the latest release note was going at the bottom (meaning you'd have to scroll to see it), but I'm not sure yet if that was only because I confused it with my fake tags or what was the cause.
> 
> I can confirm that Github shows the releases in descending order (at least that's how it always worked for me). If we would have to delete a release each time we create a release, I would question the process ;-)

I just thought we wanted the latest release note to be on top all the time, that's all. Let's see how it works, I have some doubts.

PR #47, opened 2021-11-24
EDU-122 set up database schema hash

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-122

### DIFF
--- a/src/stores/database.js
+++ b/src/stores/database.js
@@ -1,4 +1,5 @@
 import url from 'url';
+import md5 from 'md5';
 import path from 'path';
 import glob from 'glob';
 import { promisify } from 'util';
@@ -113,6 +114,14 @@ class Database {
     umzug.on('migrated', ({ name }) => logger.info(`Finished migrating ${name}`));
 
     await umzug.up();
+  }
+
+  async generateSchemaHash() {
+    const migrations = await this._db.collection('migrations').find({}).toArray();
+    const migrationNames = migrations.map(migration => migration.name).join();
+    this.schemaHash = md5(migrationNames);
+
+    return this.schemaHash;
   }
 
   static async create({ connectionString }) {
--- a/src/stores/database.js
+++ b/src/stores/database.js
@@ -2,6 +2,7 @@ import url from 'url';
 import md5 from 'md5';
 import path from 'path';
 import glob from 'glob';
+import memoizee from 'memoizee';
 import { promisify } from 'util';
 import { MongoClient } from 'mongodb';
 import Logger from '../common/logger.js';
@@ -36,6 +37,7 @@ const collectionSpecs = [
 class Database {
   constructor(connectionString) {
     this._connectionString = connectionString;
+    this.getSchemaHash = memoizee(this._generateSchemaHash);
   }
 
   async connect() {
@@ -116,12 +118,11 @@ class Database {
     await umzug.up();
   }
 
-  async generateSchemaHash() {
+  async _generateSchemaHash() {
     const migrations = await this._db.collection('migrations').find({}).toArray();
     const migrationNames = migrations.map(migration => migration.name).sort().join();
-    this.schemaHash = md5(migrationNames);
 
-    return this.schemaHash;
+    return md5(migrationNames);
   }
 
   static async create({ connectionString }) {
--- a/src/stores/database.js
+++ b/src/stores/database.js
@@ -118,7 +118,7 @@ class Database {
 
   async generateSchemaHash() {
     const migrations = await this._db.collection('migrations').find({}).toArray();
-    const migrationNames = migrations.map(migration => migration.name).join();
+    const migrationNames = migrations.map(migration => migration.name).sort().join();
     this.schemaHash = md5(migrationNames);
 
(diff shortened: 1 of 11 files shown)


src/stores/database.js
```diff
@@ -115,6 +116,14 @@ class Database {
     await umzug.up();
   }
 
+  async generateSchemaHash() {
+    const migrations = await this._db.collection('migrations').find({}).toArray();
+    const migrationNames = migrations.map(migration => migration.name).join();
```
laurentiu-ilici, line 121: Would it be good to sort the names just to make extra extra sure?

src/stores/database.js
```diff
@@ -115,6 +116,14 @@ class Database {
     await umzug.up();
   }
 
+  async generateSchemaHash() {
```
laurentiu-ilici, line 119: Since we generate this only once I think there is a very small chance that this can be outdated (if the next version already runs the migrations behind, but is not serving yet), but it's so small that I guess we don't need to worry about it.

laurentiu-ilici (approved):

src/stores/database.js
adrianaluchian: I considered the order but thing is it's the exact same code that places these records, right?
But ok, it would not hurt.

laurentiu-ilici: Well, it should in theory be like that, but you know theory and practice sometimes diverge :))

src/stores/database.js
adrianaluchian: We have already the problem of the DB being migrated but the corresponding code not being served yet, which can cause problems all over the project. So yeah, nothing new or particular to this PR.

src/stores/database.js
adrianaluchian: Done!

src/stores/database.js
```diff
@@ -115,6 +116,14 @@ class Database {
...
+    const migrations = await this._db.collection('migrations').find({}).toArray();
+    const migrationNames = migrations.map(migration => migration.name).sort().join();
+    this.schemaHash = md5(migrationNames);
```
ahelmberger, line 122: I think we should create the schema hash from the migration names in the app, not query it from the DB. What do you think?

laurentiu-ilici: They should be the same... it seems simpler to me to take them form the db.

adrianaluchian: Upf, actually that's tricky. The ones in the DB are the ones **actually run**, but now we have the case of manual scripts which we never know when they were ran. Unless we actually put them in the DB after running, which I think would be the correct way to do it.
So then what do you think about manually registering a manually run script in the DB to keep consistency and clarity over what's in the DB?

adrianaluchian: Oh well in that case I would stick to the DB collection as I find it sure to be correct. For example running the code locally would generate an incorrect hash because we don't run the migrations automatically locally.

ahelmberger: The manually run **are** put into the db when they have been run. It's tricky I agree, I just have the feeling that the importing systems are talking to each other via REST, so actually the schema you get is the schema of the running app (which in turn still has to be aligned to the db). I also thought it would be easier and would spare a db call as the migration names are already there in this file, you just would need to hash them. Anyway, I'm not insisting on this. I would propose though to remove the side effect of the field `schemaHash` just being set if one has not forgotten to call this function first. Maybe make the field "private" and always call `getSchemaHash` (the db call can still be executed only once)?

ahelmberger: We now run all migrations everytime (also locally) BTW

laurentiu-ilici: Actually now that I think about it, yesterday when we ran the manual script we didn't restart the services. This means that the hash would be out of date for a long time possibly. 
I'd propose to either agree to restart the services each time we do a manual run or easier, to regenerate the hash each time we need it.

adrianaluchian: True, in a normal scenario when scripts are run automatically, the service is started afterwards. So I think restart would be the consistent way for manual scripts.

I was running on old knowledge with how the scripts are run locally 😵‍💫 

So coming back to where to generate the hash from, having a manual script file does not indicate whether it was run or not, as it looks like the DB does indicate that. So let's stick to using the migrations collection then.

ahelmberger: OK, but can we please remove the side effect?

adrianaluchian: Where would you think of putting this call 🤔 ? My thought was that it should always come after calling `runMigrationScripts`.

laurentiu-ilici: Sounds good to me

ahelmberger: I wouldn't call it at all without any reason. We can create the hash the first time it is requested and then memorize it.

adrianaluchian: Ok, I did not get that from the previous comment. Thanks, will do that.

adrianaluchian: Done, please check if it's what you meant.

ahelmberger: It's what I meant. Thanks.

ahelmberger (approved):

PR #48, opened 2021-11-24
EDU-123 Persist batch and tasks

ahelmberger:
The "think hard about the indices" part I left for the story where we need indices, because I might think hard but wrong in advance :-)

### DIFF
--- a/migrations/educandu-2021-11-24-01-create-task-processing-collections.js
+++ b/migrations/educandu-2021-11-24-01-create-task-processing-collections.js
@@ -1,0 +1,28 @@
+// eslint-disable-next-line camelcase
+export default class Educandu_2021_11_24_01_create_task_processing_collections {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    await this.db.createCollection('batches');
+    await this.db.createCollection('batchLocks');
+    await this.db.createCollection('tasks');
+    await this.db.createCollection('taskLocks');
+
+    await this.db.collection('taskLocks').createIndexes([
+      {
+        name: '_idx_expires_',
+        key: { expires: 1 },
+        expireAfterSeconds: 0
+      }
+    ]);
+  }
+
+  async down() {
+    await this.db.dropCollection('batches');
+    await this.db.dropCollection('batchLocks');
+    await this.db.dropCollection('tasks');
+    await this.db.dropCollection('taskLocks');
+  }
+}
--- a/src/server/import-controller.js
+++ b/src/server/import-controller.js
@@ -1,9 +1,17 @@
+import express from 'express';
+import httpErrors from 'http-errors';
 import PageRenderer from './page-renderer.js';
 import permissions from '../domain/permissions.js';
 import ServerConfig from '../bootstrap/server-config.js';
 import ImportService from '../services/import-service.js';
 import ExportApiClient from '../services/export-api-client.js';
 import needsPermission from '../domain/needs-permission-middleware.js';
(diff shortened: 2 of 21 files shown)


migrations/educandu-2021-11-24-01-create-task-processing-collections.js
```diff
@@ -0,0 +1,28 @@
...
+    await this.db.createCollection('batchLocks');
+    await this.db.createCollection('tasks');
+    await this.db.createCollection('taskLocks');
+
+    await this.db.collection('taskLocks').createIndexes([
+      {
+        name: '_idx_expires_',
+        key: { expires: 1 },
+        expireAfterSeconds: 0
+      }
```
adrianaluchian, line 18: Should we maybe also have indices on
`tasks`: `taskType`, `batchId`, `processed`
`batches`: `batchType`, `completedOn`
?
Or you were thinking we will add them as we develop, later on?

ahelmberger: I think we should introduce them as needed (see my comment above) :-)

src/server/import-controller.js
```diff
@@ -22,24 +30,30 @@ class ImportController {
...
-        res.status(400).send('importSourceName query param is required');
-        return;
-      }
-
       const importSource = this.serverConfig.importSources.find(source => source.name === importSourceName);
-
       if (!importSource) {
-        res.status(400).send(`import source ${importSourceName} is unknown`);
-        return;
+        throw new NotFound(`${importSourceName} is not a known import source`);
```
adrianaluchian, line 38: Should this rather be BadRequest? Though I would see the arguments for both options 😊

ahelmberger: I considered `importSourceName` an ID, so I thought, if that ID is not in the list of available sources, it could be NotFound, but BadRequest would also work in this scenario, I guess. I don't have a strong opinion on it.

adrianaluchian: Ok then let's keep NotFound.

migrations/educandu-2021-11-24-01-create-task-processing-collections.js
adrianaluchian: Got it 😆 Ok then

src/services/import-service.js
```diff
@@ -59,6 +74,68 @@ class ImportService {
...
+
+    try {
+      const existingActiveBatch = await this.batchStore.findOne({
+        'batchType': BATCH_TYPE.importDocuments,
+        'batchParams.source': importSource.name,
+        'completedOn': null
+      });
+
+      if (existingActiveBatch) {
+        throw new BadRequest('Cannot create a new batch while another batch for the same source is still active');
```
adrianaluchian, line 124: Tiny thing, should we extract a `throwConcurrentBatchError` and have the same message in both cases (lines 113 and 124) ?

ahelmberger: They're not exactly the same thing, but the second message kind of includes the first case, so yeah, I'll do that

adrianaluchian: I know, there's a slight difference but boils down to the same restriction.

ahelmberger: I extracted the error message into a constant and made the test check for the error type instead of the message

ahelmberger: I didn't want to extract a function in order to see the correct line number in stack traces on top where the error happened

adrianaluchian: Genius 🧠 Great!

adrianaluchian (approved):

PR #54, opened 2021-11-30
EDU-136 Test batch and task processors

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-136

### DIFF
--- a/src/services/batch-processor.spec.js
+++ b/src/services/batch-processor.spec.js
@@ -19,7 +19,6 @@ describe('batch-processor', () => {
   let taskProcessor;
 
   beforeAll(async () => {
-    sandbox.useFakeTimers(now);
     container = await setupTestEnvironment();
     taskStore = container.get(TaskStore);
     batchStore = container.get(BatchStore);
@@ -28,6 +27,7 @@ describe('batch-processor', () => {
   });
 
   beforeEach(() => {
+    sandbox.useFakeTimers(now);
     sandbox.stub(taskStore, 'findRandomOne');
     sandbox.stub(batchStore, 'findOne');
     sandbox.stub(batchStore, 'save');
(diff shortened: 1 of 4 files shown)


src/services/batch-processor.spec.js
```diff
@@ -0,0 +1,165 @@
...
+      it('should return false', () => {
+        expect(result).toBe(false);
+      });
+    });
+
+    describe('when there are no tasks to process', () => {
+      let uncompletedBatch;
+
+      beforeEach(async () => {
+        uncompletedBatch = {};
```
ahelmberger: While reading this, I think I found an error in our implementation (or maybe I have a knot in my brain?): After we have found an uncomplete batch, we try to get a random task for it **without** filtering by `batchId` (`taskStore.findRandomOne({ processed: false })`, from then on we process this task in connection with the uncomplete batch (e.g. by handling down the batch params to the concrete task processor. But what if there are two uncomplete batches and the randomly fetched task does not really belong to the batch we queried before? If I'm not mistaken, than we have to filter the random sample query by `batchId` :thinking:

src/services/task-processor.js
```diff
@@ -71,10 +71,10 @@ export default class TaskProcessor {
       }
 
       logger.debug('Saving task');
-      this.taskStore.save(nextTask);
+      await this.taskStore.save(nextTask);
 
     } finally {
-      this.taskLockStore.releaseLock(lock);
+      await this.taskLockStore.releaseLock(lock);
```
laurentiu-ilici, line 77: Very important await :)

src/services/batch-processor.spec.js
```diff
@@ -0,0 +1,165 @@
...
+
+        result = await sut.process(ctx);
+      });
+
+      it('should call taskStore.findRandomOne', () => {
+        sinon.assert.calledWith(taskStore.findRandomOne, { processed: false });
+      });
+
+      it('should complete the batch', () => {
+        uncompletedBatch.completedOn = now;
```
ahelmberger, line 89: I don't understand why we set this property here in the test setup, shouldn't we rather check if it was set by the processor?

laurentiu-ilici (approved):

src/services/batch-processor.spec.js
laurentiu-ilici: I lingered on the same thought a few minutes ago :))

adrianaluchian: Very good catch, and I think we might have mentioned this while implementing and forgot to actually do it 😬 Will correct!

src/services/task-processor.js
adrianaluchian: Oh there were even 2 of them that we had missed 😅

src/services/batch-processor.spec.js
adrianaluchian: Oh because I got carried away and this is wrong, corrected!

ahelmberger (approved):

PR #55, opened 2021-11-30
EDU-124 extend export pages

laurentiu-ilici:
Resolves: https://educandu.atlassian.net/browse/EDU-124


src/components/pages/import.js
```diff
@@ -1,152 +1,149 @@
...
+  const getErrorsExtra = ({ errors }) => <span>{errors?.length || 0} {t('errors')}</span>;
+
+  const attemptsRenderer = (attempts, task) => {
+    const errors = task.attempts.map(attempt => attempt.errors).flat();
+    return (
+      <Collapse>
+        <Panel key={task._id} header={header()} extra={getErrorsExtra({ errors })}>
+          <Table
+            key={task._id}
+            rowKey={attempt => `${attempt.startedOn}: ${attempt.completedOn || Math.random()}`}
```
laurentiu-ilici, line 75: If you have a better idea, please shoot

src/server/client-data-mapper.js
```diff
@@ -55,6 +55,38 @@ class ClientDataMapper {
     });
   }
 
+  async mapImportBatches(batches, user) {
```
laurentiu-ilici, line 58: I'll add tests for these 2 functions

src/services/import-service.js
```diff
@@ -139,6 +141,63 @@ class ImportService {
...
+      if (current._id === true) {
+        accumulator.processedCount += current.count;
+      }
+      return accumulator;
+    }, { totalCount: 0, processedCount: 0 });
+
+    return stats.totalCount === 0 ? 1 : stats.processedCount / stats.totalCount;
+  }
+
+  async getImportBatches() {
```
laurentiu-ilici, line 176: I'll add tests here

src/components/pages/create-import.js
```diff
@@ -0,0 +1,159 @@
...
+
+        <Input
+          placeholder={t('source')}
+          className="CreateImportPage-source"
+          defaultValue={selectedSource?.name}
+          disabled
+          />
+        <br /> <br />
+
+        <Table
```
laurentiu-ilici, line 139: I wanted to use an expandable table here, but the expand button wouldn't appear... So I went the collapse panel variant, not bad but not great either.

src/components/pages/import.js
```diff
@@ -1,152 +1,149 @@
...
-    return <span>{formatDate(doc.updatedOn)}</span>;
+  const header = () => <span>{t('attemptsHeader')}</span>;
+  const getErrorsExtra = ({ errors }) => <span>{errors?.length || 0} {t('errors')}</span>;
+
+  const attemptsRenderer = (attempts, task) => {
+    const errors = task.attempts.map(attempt => attempt.errors).flat();
+    return (
+      <Collapse>
+        <Panel key={task._id} header={header()} extra={getErrorsExtra({ errors })}>
+          <Table
```
laurentiu-ilici, line 73: I wanted to use an expandable table here, but the expand button wouldn't appear... So I went the collapse panel variant, not bad but not great either.

src/components/pages/import.js
```diff
@@ -1,152 +1,149 @@
...
+          </Panel>
+        </Collapse>
         <Table
-          bordered
-          size="small"
-          rowSelection={{
-            type: 'checkbox',
-            ...tableRowSelection
-          }}
+          style={{ marginTop: '20px' }}
```
laurentiu-ilici, line 132: I wanted to use an expandable table here, but the expand button wouldn't appear... So I went the collapse panel variant, not bad but not great either.

src/components/pages/import.js
adrianaluchian: Oh let's please not use random artefacts for keys, they should actually identify the rendered item uniquely **and repeatedly** and not be new each time. Why not just `attempt.startedOn`? It's a timestamp down to millisecond. Or even `${task._id}-${attempt.startedOn}`, then that will surely be unique.

src/components/pages/create-import.js
```diff
@@ -0,0 +1,159 @@
...
+import ClientConfig from '../../bootstrap/client-config.js';
+import { useReloadPersistedWindow } from '../../ui/hooks.js';
+import { DOCUMENT_IMPORT_TYPE } from '../../common/constants.js';
+import ImportApiClient from '../../services/import-api-client.js';
+import { useDateFormat, useLanguage } from '../language-context.js';
+import LanguageNameProvider from '../../data/language-name-provider.js';
+import CountryFlagAndName from '../localization/country-flag-and-name.js';
+import { CloudDownloadOutlined, CloudSyncOutlined } from '@ant-design/icons';
+import urls, { getArticleUrl, getImportDetailsUrl, getImportSourceBaseUrl } from '../../utils/urls.js';
+
```
adrianaluchian, line 18: My OCD appreciates this, thanks 😄

adrianaluchian: Unless it's from my original page 🤔 But still 🙏

src/components/pages/import.js
```diff
@@ -1,152 +1,149 @@
...
+    return <span>{t('batchStatusProcessing', { progress: batch.progress })}</span>;
   };
 
-  const renderImportType = doc => {
+  const renderDate = date => <span>{formatDate(date)}</span>;
+  const renderTitle = ({ title, slug }) => (
+    <a
+      target="_blank"
+      href={getImportedArticleUrl({ hostName, allowUnsecure, slug })}
+      rel="noreferrer"
```
adrianaluchian, line 42: Tiny thing, we could throw in `noopener` too, I know the linter prompted for `noreferrer` but then I'd say have both to be consistent with other links.

src/components/pages/import.js
```diff
@@ -1,152 +1,149 @@
...
+  const renderErrors = errors => <span>{errors.join(';')}</span>;
+  const attemtpsColumns = [
+    { title: t('startedOn'), dataIndex: 'startedOn', width: '50px', render: renderDate },
+    { title: t('completedOn'), key: '_id', dataIndex: 'completedOn', width: '150px', render: renderDate },
+    { title: t('errors'), dataIndex: 'errors', width: '150px', render: renderErrors }
+  ];
 
-  const renderUpdateDate = doc => {
-    return <span>{formatDate(doc.updatedOn)}</span>;
+  const header = () => <span>{t('attemptsHeader')}</span>;
```
adrianaluchian, line 65: Maybe then let's call this `renderHeader`, for consistency.

src/components/pages/import.js
adrianaluchian: Since you style everything else with classes here, you could maybe throw this style in a class too.

src/components/pages/import.js
```diff
@@ -1,152 +1,149 @@
-import by from 'thenby';
```
adrianaluchian, line 1: Shall we rename this file maybe to `batch-imports.js`? I feel it can get confusing between `imports` and `import`, when both contain imports of multiple documents.
I think this page will be under the route `imports/<batchId>`, so it would correspond with the file name.

Or `imports-batch.js`, `create-imports.js` -> `imports-create.js`? And then all 3 files reflect their URL route and are in the file system one after the other?

What do you think?

adrianaluchian: Or `imports.js`, `imports-view.js`, `imports-create.js`? 😬

src/components/pages/imports.js
```diff
@@ -0,0 +1,112 @@
...
+    batches: value.batches.sort(by(batch => batch.createdOn, 'desc'))
+  }));
+
+  const handleCreateImport = source => {
+    window.location = urls.getCreateImportUrl(source.importSourceHost);
+  };
+
+  const renderId = id => <a href={urls.getBatchUrl(id)}>{t('viewBatch')}</a>;
+
+  const renderStatus = (_, batch) => {
```
adrianaluchian, line 44: Also small, but can we put the name of the first argument here as `_name`? I think this helps alot for future development, in case that argument is needed, to directly see what is expected there and not have to dig for it.

src/components/pages/imports.js
```diff
@@ -0,0 +1,112 @@
...
+      <Button
+        disabled={isCreateButtonDisabled}
+        onClick={() => { handleCreateImport(source); }}
+        >{t('createImport')}
+      </Button>);
+  };
+
+  const header = source => {
+    return (
+      <span style={{ marginRight: '20px' }}>{t('importsHeaderPrefix')}: {source.importSourceName}</span>
```
adrianaluchian, line 78: Same remark here, regarding styling with classes, to have consistency.

src/resources/page-names.yml
```diff
@@ -13,6 +13,9 @@ settings:
 import:
   en: Import
   de: Import
+imports:
+  en: Imports overview
+  de: Übersicht über Importe
```
adrianaluchian, line 18: This would sound to me more like `Import-Übersich` or `Importsübersicht`, but of course the native has the call here 😋

src/server/client-data-mapper.js
```diff
@@ -55,6 +55,38 @@ class ClientDataMapper {
...
+    const idSet = new Set(batches.map(batch => batch.createdBy));
```
adrianaluchian, line 59: I think `userIdSet` would be better, I thought initially it's the batches Ids.

src/components/pages/import.js
laurentiu-ilici: I'm currently following the convention in the codebase. We have for example "docs.js" for all docs and doc.js for one doc. It's a bit confusing, I know @ahelmberger what's your opinion, I for one like Adriana's sugggestion

src/resources/page-names.yml
ahelmberger: I think `Import-Übersicht` sounds best.

src/components/pages/import.js
laurentiu-ilici: Added, thanks!

src/components/pages/import.js
laurentiu-ilici: Renamed, and I renamed the getErrorExtra to renderErrorExtra.

src/server/client-data-mapper.js
```diff
@@ -55,6 +55,38 @@ class ClientDataMapper {
...
+        createdBy: this._mapUser(userMap.get(batch.createdBy), allowedUserFields)
+      };
+    });
+  }
+
+  async mapImportBatch(batch, user) {
+    const users = await this.userService.getUsersByIds([batch.createdBy]);
+    const allowedUserFields = privateData.getAllowedUserFields(user);
+
+    if (users.length !== 1) {
```
adrianaluchian, line 80: Can this actually ever occur, since the users collection cannot have duplicate `_id`s?

src/components/pages/import.js
laurentiu-ilici: Dooh, didn't think of that. Thanks.

src/components/pages/imports.js
laurentiu-ilici: Actually for this one I realized that the batch itself is the "_" :)). Thanks!

src/resources/page-names.yml
laurentiu-ilici: Okay, will do, google translate lied to me!

src/server/import-controller.js
```diff
@@ -14,18 +15,35 @@ const { NotFound } = httpErrors;
...
+      const importSources = this.serverConfig.importSources.slice();
+
+      return this.pageRenderer.sendPage(req, res, 'edit-bundle', 'imports', { batches, importSources });
+    });
+
+    app.get('/imports/create', needsPermission(permissions.MANAGE_IMPORT), (req, res) => {
+      return this.pageRenderer.sendPage(req, res, 'edit-bundle', 'create-import');
+    });
+
+    app.get('/imports/:key', needsPermission(permissions.MANAGE_IMPORT), async (req, res) => {
```
adrianaluchian, line 41: From a data structure perspective we have `key` for documents only, I'd suggest to call this `:batchId` (same for the corresponding variable). Would make the route clear, plus `getImportBatchDetails` expects something called id.

src/resources/page-names.yml
ahelmberger: No, but GT would not pass a C level exam

src/utils/urls.js
```diff
@@ -65,8 +66,29 @@ export function getSettingsUrl() {
...
+export function getImportsUrl() {
+  return importsPath;
+}
+
+export function getCreateImportUrl(sourceName) {
+  const url = new URL(document.location);
+  const keys = [];
+  for (const key of url.searchParams.keys()) {
+    keys.push(key);
+  }
```
adrianaluchian, line 78: Not sure I get this part 🤔 Can it be reduced to `const keys = url.searchParams.keys()`?

adrianaluchian: Actually, why does this URL creation need to be so complex?
Would it not be simpler to just compose the URL? 🤷‍♀️

src/components/pages/import.js
laurentiu-ilici: Sure

src/ui/hooks.js
```diff
@@ -12,6 +12,23 @@ export function usePermission(permissionToCheck) {
   );
 }
 
+export function useReloadPersistedWindow() {
```
adrianaluchian, line 15: Cool that you extracted this as a hook 👍

src/ui/default-prop-types.js
```diff
@@ -173,3 +177,44 @@ export const formItemLayoutShape = PropTypes.shape({
     sm: formItemDimensionShape.isRequired
   }).isRequired
 });
+
+export const importTaskShape = PropTypes.shape({
+  _id: PropTypes.string.isRequired,
+  batchId: PropTypes.string.isRequired,
+  taskType: PropTypes.oneOf([TASK_TYPE.importDocument]),
```
adrianaluchian, line 184: We could maybe here say `PropTypes.oneOf(Object.values(TASK_TYPE)` to not need to change it when we add new task types.

src/ui/default-prop-types.js
```diff
@@ -173,3 +177,44 @@ export const formItemLayoutShape = PropTypes.shape({
...
+  ],
+  taskParams: {
+    key: PropTypes.string.isRequired,
+    title: PropTypes.string.isRequired,
+    slug: PropTypes.string.isRequired,
+    language: PropTypes.string.isRequired,
+    updatedOn: PropTypes.string,
+    importedRevision: PropTypes.string,
+    importableRevision: PropTypes.string.isRequired,
+    importType: PropTypes.oneOf([DOCUMENT_IMPORT_TYPE.add, DOCUMENT_IMPORT_TYPE.update])
```
adrianaluchian, line 201: Same remark like here https://github.com/educandu/educandu/pull/55/files#r759999704 with this guy

src/ui/default-prop-types.js
```diff
@@ -173,3 +177,44 @@ export const formItemLayoutShape = PropTypes.shape({
...
+  }
+});
+
+export const importBatchShape = PropTypes.shape({
+  _id: PropTypes.string.isRequired,
+  createdBy: userInDocShape.isRequired,
+  createdOn: PropTypes.string.isRequired,
+  completedOn: PropTypes.string,
+  batchType: PropTypes.oneOf([BATCH_TYPE.importDocuments]),
```
adrianaluchian, line 210: And same remark like here https://github.com/educandu/educandu/pull/55/files#r759999704 with this guy too 😊

src/utils/urls.js
ahelmberger: I also thought that smth like `${createImportsPath}?source=${encodeURIComponent(sourceName)}` would be sufficient

src/components/pages/import.js
adrianaluchian: Yeah for docs and doc it makes sense cause it's clear in itself. But for imports not so much, it would make more sense like `import-batches` and `import-batch-view`, `import-batch-create` (👈  which is also a good combination 😂 )

src/components/pages/imports.js
adrianaluchian: Oh, ouch 😄

src/components/pages/create-import.js
laurentiu-ilici: It isn't, I tried my best, you might see other files with small changes that also got the treatment. There's even a commit entitled "arrange imports" or something like that :).

src/server/import-controller.js
laurentiu-ilici: Yep, changed, I was going for consistency with the other endpoints but it really doesn't make sense.

src/services/import-service.js
laurentiu-ilici: Addded

src/ui/default-prop-types.js
laurentiu-ilici: Ah, true, thanks!

src/components/pages/import.js
```diff
@@ -39,7 +39,8 @@ function Import({ initialState }) {
     <a
       target="_blank"
       href={getImportedArticleUrl({ hostName, allowUnsecure, slug })}
-      rel="noreferrer"
+      rel="noreferrer noreferrer"
+      noopener
```
adrianaluchian, line 43: Something weird happened here 😬

src/ui/default-prop-types.js
laurentiu-ilici: Resolved all

src/components/pages/import.js
ahelmberger: The page names mostly follow the URLs (loosely), so as long as they're kinda related, I'm fin with all proposed filenames above.

src/resources/page-names.yml
```diff
@@ -13,6 +13,9 @@ settings:
 import:
   en: Import
   de: Import
+imports:
+  en: Imports overview
+  de: Importsübersicht
```
adrianaluchian, line 18: Andreas' suggestion was for the other option https://github.com/educandu/educandu/pull/55#discussion_r759986889

src/utils/urls.js
laurentiu-ilici: I kinda took this from the creation of the search url, I'm a bit afraid of coming here by a bad link, I want to remove all  keys from the url that don't belong (copy pasta errors can occur). Probably it's just in my head though, no one will get to this page via copy paste :). I'll simplify.

adrianaluchian: Yeah this is a very simple URL, no need to overthink here.

src/services/import-service.js
```diff
@@ -139,6 +141,63 @@ class ImportService {
...
+          count: {
+            $sum: 1
+          }
+        }
+      }
+    ]);
+
+    const stats = countGroups.reduce((accumulator, current) => {
+      accumulator.totalCount += current.count;
+      if (current._id === true) {
```
adrianaluchian, line 167: I'm not sure I get this condition, is `current._id` not the task id?

src/services/import-service.spec.js
```diff
@@ -311,4 +311,100 @@ describe('import-service', () => {
...
+            allowUnsecure: false
+          },
+          errors: [],
+          progress: 0.5
+        }
+      ]);
+    });
+  });
+
+  describe('getImportBatch', () => {
```
adrianaluchian, line 344: 'getImportBatchDetails'

src/services/import-service.spec.js
```diff
@@ -311,4 +311,100 @@ describe('import-service', () => {
...
+          createdOn: expect.any(Date),
+          completedOn: null,
+          batchType: BATCH_TYPE.importDocuments,
+          batchParams: {
+            name: 'source-1',
+            hostName: 'source1.com',
+            allowUnsecure: false
+          },
+          errors: [],
+          progress: 0.5
```
adrianaluchian, line 338: For this value I suggest to not test it indirectly and do a mock of what `getProgressForBatch` returns and just check that whatever the mock returns is put in here.

I would find super useful to actually have a direct test for `getProgressForBatch` (with mocking what the DB returns, of course).

src/server/client-data-mapper.js
laurentiu-ilici: It's more about not having the user :)

src/utils/urls.js
laurentiu-ilici: Solved

src/ui/default-prop-types.js
laurentiu-ilici: Done, thanks!

src/components/pages/import.js
laurentiu-ilici: Done, thanks!

src/server/client-data-mapper.js
ahelmberger: True, we always check in the mapper if the number matches, so it doesn't hurt to have it here.

src/services/import-service.spec.js
laurentiu-ilici: This one actually tests that our aggregation pipeline works correctly. I could make it a separate assertion though, to make that more clear.

src/components/pages/import.js
laurentiu-ilici: I went for:
import-batches
import-batch-view
create-import-batch (this one follows our other "create-something" pattern in the codebase)
Urls:
import-batches
import-batches/:batchId
import-batches/create?source=abc

I hope that's alright :).

src/resources/page-names.yml
laurentiu-ilici: Fixed :D

src/services/import-service.spec.js
laurentiu-ilici: Fixed, thanks, I used the wrong method name at first, then forgot to update the title of the describe :).

src/services/import-service.js
laurentiu-ilici: Fixed, it is indeed a bit confusing, I also needed a few seconds to remember what that is so :)).

src/components/pages/import.js
adrianaluchian: Hm, I still see the same here in `src/components/pages/import-batch-view.js`

src/components/pages/import-batch-view.yml
```diff
@@ -0,0 +1,60 @@
...
+  de: Erstellt am
+completedOn:
+  en: Completed on
+  de: Vervollständigt am
+createdBy:
+  en: Created by
+  de: Erstellt von
+importHeading:
+  en: Import overview
+  de: Import ubersicht
```
adrianaluchian, line 27: This guy is still wrong here.

src/services/import-service.spec.js
adrianaluchian: Or at least that 🙏 Cause actually that's the logic that needs testing, otherwise `getImportBatches` itself doesn't do much

adrianaluchian:
From UI perspective I would suggest:
- to make the source of the import on the `/import-batches/create` page displayed not as a disabled button but as a nice text: "**Source:** NameOfSource"
- to remove the padding of the table in each collapsible, if possible

src/components/pages/import-batch-view.yml
laurentiu-ilici: What the heck, I copy pasted it, hmmmmmmmmmm

src/server/client-data-mapper.js
laurentiu-ilici: Added

src/components/pages/import-batch-view.yml
laurentiu-ilici: Fixed

laurentiu-ilici:
> From UI perspective I would suggest:
> 
> * to make the source of the import on the `/import-batches/create` page displayed not as a disabled button but as a nice text: "**Source:** NameOfSource"
> * to remove the padding of the table in each collapsible, if possible

The second request is fixed, I'm not sure about the first one. I don't understand the requirement.

src/components/pages/import.js
adrianaluchian: Now it's `rel="noreferrer noreferrer"` 😂 So, it should be `rel="noreferrer noopener"`

src/components/pages/import-batch-view.js
```diff
@@ -0,0 +1,171 @@
...
+
+    return isTaskSuccessful(task) ? t('common:yes') : t('common:no');
+  };
+
+  const taskTableColumns = [
+    { title: t('importType'), dataIndex: 'taskParams', width: '50px', render: renderImportType },
+    { title: t('documentTitle'), key: '_id', dataIndex: 'taskParams', width: '150px', render: renderTitle },
+    { title: t('taskStatus'), dataIndex: 'processed', width: '150px', render: renderStatus, sorter: task => task.processed ? 1 : -1 },
+    { title: t('hasErrors'), width: '150px', render: renderHasErrrors, sorter: task => doesTaskHaveErrors(task) ? 1 : -1 },
+    { title: t('isSuccessful'), width: '150px', render: renderIsSuccessful, sorter: task => isTaskSuccessful(task) ? 1 : -1 }
```
laurentiu-ilici, line 107: These sorters aren't real sorters, they just do a grouping. One caveat of this particular one is that it doesn't distinguish between unsuccessful tasks and tasks that were not processed yet. If we want that I'd rather make another ticket for "proper sorters" and extend the functionality since this ticket has already a lot of new code.

src/components/pages/import-batch-view.js
```diff
@@ -0,0 +1,171 @@
...
+            rowExpandable: task => task.attempts?.length
+          }}
+          />
+      </div>
+    </Page>);
+}
+
+ImportBatchView.propTypes = {
+  initialState: PropTypes.shape({
+    batch: importBatchDetailsShape.isRequired
```
laurentiu-ilici, line 167: There's a console warning about this line which I cannot figure out why it's there. The warning says that this is required but an undefined value was given. I'm not sure how that happens since we do server side rendering and the batch is always there when I send the page. The same happens on the import-batches.js

src/components/pages/import.js
laurentiu-ilici: Fixed, sorry, got confused several times here it seems. I should pay more attention.

src/services/import-service.spec.js
laurentiu-ilici: I'm not sure what the point of mocking would be here, the heart of the logic is in testing that aggregation pipeline, if we mock that out what is the benefit?

laurentiu-ilici:
> From UI perspective I would suggest:
> 
> * to make the source of the import on the `/import-batches/create` page displayed not as a disabled button but as a nice text: "**Source:** NameOfSource"
> * to remove the padding of the table in each collapsible, if possible

Fixed both

src/components/pages/import-batch-view.js
ahelmberger: Untitled

laurentiu-ilici: Okay, I see, I don't know how to read, sorry.

src/components/pages/import-batch-view.js
ahelmberger: Sounds good. We could make a ticket to make all tables (that are not yet sortable) in the admin area sortable.

src/components/pages/import-batch-view.yml
```diff
@@ -0,0 +1,67 @@
...
+  de: Erstellt von
+batchStatus:
+  en: Batch status
+  de: Chargenstatus
+batchStatusPending:
+  en: Pending
+  de: Ausstehend
+batchStatusDone:
+  en: Done
+  de: Vertig
```
ahelmberger, line 33: Abgeschlossen

src/components/pages/import-batch-view.yml
```diff
@@ -0,0 +1,67 @@
...
+  de: Ausstehend
+taskStatusDone:
+  en: Done
+  de: Vertig
+createdOn:
+  en: Created on
+  de: Erstellt am
+completedOn:
+  en: Completed on
+  de: Vervollständigt am
```
ahelmberger, line 21: Abgeschlossen am

src/components/pages/import-batch-view.yml
```diff
@@ -0,0 +1,67 @@
...
+  de: Tasktyp
+taskStatus:
+  en: Task status
+  de: Taskstatus
+taskStatusPending:
+  en: Pending
+  de: Ausstehend
+taskStatusDone:
+  en: Done
+  de: Vertig
```
ahelmberger, line 15: Abgeschlossen

src/components/pages/import-batch-view.yml
```diff
@@ -0,0 +1,67 @@
...
+  de: Fehler
+errorsCount:
+  en: Errors count
+  de: Fehleranzahl
+attemptsHeader:
+  en: Attempt details
+  de: Versuchsdetails
+batchErrors:
+  en: Batch errors
+  de: Chargenfelher
```
ahelmberger, line 60: Chargenf**eh**ler

src/components/pages/import-batch-view.js
ahelmberger: It might be confusing for the user (at least I was) to have two columns "hasErrors" and "isSuccessful". Would any one of those ever be anything else than the opposite of the other? Maybe we can merge this into one column (yay/nay)?

src/resources/page-names.yml
```diff
@@ -10,9 +10,15 @@ account:
 settings:
   en: Settings
   de: Einstellungen
-import:
```
ahelmberger, line 13: This breaks line 126 in `page.js`

src/components/pages/import-batch-view.js
ahelmberger: :laughing:

src/components/pages/import-batch-view.js
laurentiu-ilici: I've done the sorting better cause I was getting bored. Please look at the task-utils.

It's useful to see if you had attempts that had errors (I think for us it will be easier to follow that). The logic is that a task can have errors on one attempt but succeed on a subsequent one. The first column tells you if there were any errors at all, the last one is the final state (success or failure). We could drop the one that says "have we had errors at all" if you think it's too much.

adrianaluchian: Can we please do this separately? 🙏 This PR has so many corrections that it reached 38 commits and 85 comments, I think the focus should be on stabilising it and getting it merged without risking side-effects.

ahelmberger: I agree, just saying that the column names are confusing, I think we shouldn't have two boolean columns with labels that are opposite to each other. But we could do one of the following:
* Rename "hasErrors" to "errors" and actually show the errors in that column
* Merge the two columns into one that can have one of the three values "succeeded" / "succeeded with errors" / "failed"
I would like to see the actual errors, so the first option would be my favorite

src/components/pages/import-batch-view.js
laurentiu-ilici: Hmm, this seems to have been another one that I didn't see, I still get the original one. I'll try to learn to read now :)). I think I mislead you with this line, it was on the /import-batches/ page. I'm getting more and more confused with all the renaming and the code sorry.

src/services/import-service.spec.js
adrianaluchian: Please read the comments in order.
You said "...  I could make it a separate assertion though, to make that more clear."
And I said "Or at least that ..."

adrianaluchian: It's true, DB should not be mocked regardless of testing `getProgressForBatch` directly or indirectly via `getImportBatches`. But since you didn't want to go for testing it directly I had agreed to a clear separate assertion.

laurentiu-ilici: Then I don't understand your initial comment. That getProgress for batch would actually be with an _ it was meant to be "private" (same as what we have in the document-service) since by itself it's pretty useless. I'll rename because I guess that is why we have this confusion.

adrianaluchian: I don't know what you mean by `meant to be "private"` but not worth dragging this so much anymore, I'm fine if you just have a specific assert for the progression value alone.

laurentiu-ilici: Added the tests :)

src/server/client-data-mapper.js
laurentiu-ilici: This was fixed, I forgot to comment on it

src/ui/default-prop-types.js
```diff
@@ -16,14 +17,17 @@ export const sectionEditorProps = {
   onContentChanged: PropTypes.func.isRequired
 };
 
+export const importSourceShape = PropTypes.shape({
+  name: PropTypes.string.isRequired,
+  hostName: PropTypes.string.isRequired,
+  allowUnsecure: PropTypes.bool
+});
```
ahelmberger, line 24: There is no need to inject these through initial state as it is already present in the client config.

src/ui/default-prop-types.js
```diff
@@ -173,3 +177,44 @@ export const formItemLayoutShape = PropTypes.shape({
...
+
+export const importBatchShape = PropTypes.shape({
+  _id: PropTypes.string.isRequired,
+  createdBy: userInDocShape.isRequired,
+  createdOn: PropTypes.string.isRequired,
+  completedOn: PropTypes.string,
+  batchType: PropTypes.oneOf(Object.values(BATCH_TYPE)),
+  batchParams: PropTypes.shape({
+    ...importSourceShape
```
ahelmberger, line 212: You can't destructure this shape as it is a data structure created by the `prop-types` package, not a plain JS object (note that in this file there are variables called `...Props` (JS objects) and `...Shape` (objects created by the library))

src/ui/default-prop-types.js
```diff
@@ -173,3 +177,44 @@ export const formItemLayoutShape = PropTypes.shape({
...
+  }),
+  errors: PropTypes.arrayOf(PropTypes.any).isRequired
+});
+
+export const importBatchDetailsShape = PropTypes.shape({
+  ...importBatchShape,
```
ahelmberger, line 218: Same issue here (see my comment above)

src/ui/default-prop-types.js
ahelmberger: Sorry, that's already the case, please ignore the comment !

src/ui/default-prop-types.js
laurentiu-ilici: Interesting, I thought you could because I saw further up the 'languageProps" but I see that is not a shape. Maybe this is why I have the warnings. I will read a bit about these shapes to see what is the benefit of declaring them as a "shape" instead of just props like we have the language.

src/ui/default-prop-types.js
laurentiu-ilici: Solved both of them, that took a while to figure out :).

ahelmberger (approved):

adrianaluchian (approved):

PR #57, opened 2021-12-02
EDU-89 page templating

ahelmberger:
Resolves https://educandu.atlassian.net/browse/EDU-89


src/components/pages/article.js
```diff
@@ -49,11 +58,12 @@ function Article({ initialState }) {
       <aside className="Content">
         <CreditsFooter documentOrRevision={documentOrRevision} type={type} />
       </aside>
-    </Page>
+    </PageTemplate>
   );
 }
 
 Article.propTypes = {
+  PageTemplate: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
ahelmberger: Nope, it should be the Component (Class/Function) for the template. The page itself will then instantiate the React element from it.

src/components/repository-browser.js
```diff
@@ -207,10 +206,6 @@ class RepositoryBrowser extends React.Component {
       return;
     }
 
-    if (!browserHelper.isIE()) {
```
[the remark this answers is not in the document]
ahelmberger: :laughing:

src/domain/page-resolver.js
```diff
@@ -0,0 +1,81 @@
...
+
+async function resolveAll(promiseCreators) {
+  // eslint-disable-next-line no-process-env
+  if (!isBrowser() && process.env.EDUCANDU_TEST === true.toString()) {
+    // Resolve sequentially in the test runner,
+    // because otherwise Jest bursts into flames:
+    // https://github.com/facebook/jest/issues/11434
+    const results = [];
+    for (const promiseCreator of promiseCreators) {
+      // eslint-disable-next-line no-await-in-loop
```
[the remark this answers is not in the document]
ahelmberger: Mostly in script files like gulpfile.js or migrations, not really in application code (I think there was just one occurrence when we started). Also, I hope that Jest will fix this bug at some point in the future so we can remove the workaround.

src/domain/page-resolver.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+    return results;
+  }
+
+  // Otherwise we can resolve everything in parallel:
+  return Promise.all(promiseCreators.map(promiseCreator => promiseCreator()));
+}
+
+export default class PageResolver {
+  constructor(bundleConfig) {
```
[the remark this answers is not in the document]
ahelmberger: Because I think it is not to be decided by the person that configures the app, rather internal mechanics. The server caches (because it renders a lot of pages), the client not (there is only one page there, caching would make it even slower).

test-app/bundles/educandu-page-template.js
```diff
@@ -0,0 +1,203 @@
...
+              <CountryFlagAndName code={lang.flag} name={lang.name} flagOnly />
+            </Button>
+          </React.Fragment>
+        ))}
+      </div>),
+    permission: null
+  });
+
+  return (
+    <div className="EducanduPageTemplate">
```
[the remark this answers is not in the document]
ahelmberger: Hm, but this is not a page (like in "DocsPage" etc.), it's the template that gets fed to the page (for example `<Docs pageTemplate={EducanduPageTemplate} />`). Wouldn't that be a little bit confusing?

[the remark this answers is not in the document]
laurentiu-ilici: True, when testing the code locally I also saw that this is in the test app, it's fine.

test-app/bundles/educandu-page-template.less
```diff
@@ -0,0 +1,126 @@
...
+
+.EducanduPageTemplate-header {
+  display: flex;
+  align-items: center;
+  min-height: @educandu-page-template-header-content-min-height + (2* @global-content-padding);
+  background-color: @educandu-blue;
+  color: #fff;
+
+  a {
+    color: white;
```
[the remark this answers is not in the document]
laurentiu-ilici: Yes you can with this color scheme

test-app/bundles/educandu-page-template.less
```diff
@@ -0,0 +1,126 @@
...
+  padding: @global-content-padding;
+  position: relative;
+  z-index: @educandu-page-template-footer-z-index;
+}
+
+.EducanduPageTemplate-footerLink {
+  display: inline-block;
+
+  &:not(:last-child)::after {
+    content: "\00A0\00A0|\00A0\00A0";
```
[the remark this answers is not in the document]
ahelmberger: Yeah, CSS doesn't let me write `&nbsp;` :-)

test-app/bundles/educandu-page-template.less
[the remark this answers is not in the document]
ahelmberger: If on a dark-blue background, sure :laughing: I changed it to `#fff` to align with line 32 though

src/components/page.js
```diff
@@ -198,12 +179,11 @@ function Page({ children, disableProfileWarning, fullScreen, headerActions, cust
 }
 
 Page.propTypes = {
-  children: PropTypes.node,
-  customAlerts: PropTypes.arrayOf(PropTypes.shape({
-    message: PropTypes.string.isRequired,
+  alerts: PropTypes.arrayOf(PropTypes.shape({
+    message: PropTypes.node.isRequired,
     type: PropTypes.oneOf(['success', 'info', 'warning', 'error'])
```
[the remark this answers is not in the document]
ahelmberger: Done!

test-app/bundles/educandu-page-template.js
```diff
@@ -0,0 +1,203 @@
...
+    'EducanduPageTemplate-contentArea--fullScreen': fullScreen
+  });
+
+  const contentClasses = classNames({
+    'EducanduPageTemplate-content': true,
+    'EducanduPageTemplate-content--fullScreen': fullScreen
+  });
+
+  let headerActionComponents = null;
+  if (headerActions && headerActions.length) {
```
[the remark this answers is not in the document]
ahelmberger: Yep, done, thanks!

laurentiu-ilici (approved):

laurentiu-ilici (approved):

src/components/page.js
```diff
@@ -198,12 +179,11 @@ function Page({ children, disableProfileWarning, fullScreen, headerActions, cust
 }
 
 Page.propTypes = {
-  children: PropTypes.node,
-  customAlerts: PropTypes.arrayOf(PropTypes.shape({
-    message: PropTypes.string.isRequired,
-    type: PropTypes.oneOf(['success', 'info', 'warning', 'error'])
+  alerts: PropTypes.arrayOf(PropTypes.shape({
```
adrianaluchian, line 182: Glad you reduced this to one set 🙏

src/components/pages/doc.js
```diff
@@ -179,13 +181,20 @@ class Doc extends React.Component {
...
 
-    const customAlerts = [];
+    const alerts = [];
     const headerActions = [];
 
+    if (isProfileInsufficient(user)) {
+      alerts.push({
+        message: <InsufficientProfileWarning />,
+        type: ALERT_TYPE.info
+      });
```
adrianaluchian, line 191: Since this is repeated selectively on some pages, maybe we can have an alerts module that gets the array of alerts by passing to it all it needs (`user`, `revision` etc.) plus the list of alerts types it wants to get for the current page (`['insufficient-profile', 'archived-document']).
Or by an object with properties being the alert type and values being the required data:
```
getAlerts({ insufficientProfile: { user }, archived: { revision } });
```
This would also insure the same priority/sorting of alerts.

Does that sound ok or an overkill? For a separate ticket of course.

test-app/bundles/educandu-page-template.js
```diff
@@ -0,0 +1,203 @@
...
+};
+
+EducanduPageTemplate.defaultProps = {
+  alerts: [],
+  children: null,
+  fullScreen: false,
+  headerActions: []
+};
+
+export default EducanduPageTemplate;
```
adrianaluchian, line 203: Man this is really cool, full control now over the layout

test-app/bundles/educandu-page-template.less
```diff
@@ -0,0 +1,126 @@
...
+  z-index: 1000;
+  position: sticky;
+  z-index: @educandu-page-template-header-z-index;
+  margin-bottom: @global-content-padding;
+}
+
+.EducanduPageTemplate-header {
+  display: flex;
+  align-items: center;
+  min-height: @educandu-page-template-header-content-min-height + (2* @global-content-padding);
```
adrianaluchian, line 30: Small thing: I would also extract this to the top in a variable, I feel it's easier to adjust/"debug" when you see all involved values defined right there

src/components/pages/doc.js
ahelmberger: We do have the insufficient profile warning an mostly all pages, so that one could be centralized, but most other alerts will probably be very specific to a single page, so I'd prefer to keep them in the page. Maybe we could do this only for the "global" alerts?

src/server/page-renderer.js
```diff
@@ -91,10 +62,7 @@ class PageRenderer extends PageRendererBase {
 
     const styles = [{ href: '/main.css' }];
 
-    const scripts = [
-      { content: inlineScript },
-      { src: `/${bundleName}.js` }
-    ];
+    const scripts = [{ content: inlineScript }, { src: '/main.js' }];
```
adrianaluchian, line 65: I won't pretend I'm completely clear on the bundle unification 😅 
For one I presume it was some overhead to keep them with the PageTemplate setup, but I guess it also makes sense since we're going in the direction of inline editing?
Or was there yet another thought behind this?

adrianaluchian: Or actually it has to do with the `page-resolver.js`? 🤔 I will study it for some clarity here.

ahelmberger: We have now automatic code splitting, so it's not necessary anymore to manually create separate bundles, also it's getting harder and harder with each new page. It's still the case that each page uses different chunks, they just start all with `main.js` now.

test-app/bundles/educandu-page-template.less
ahelmberger: Sure, done.

test-app/bundles/educandu-page-template.js
ahelmberger: I guess it's unfortunately still not enough to completely drive elmu and oma, but a firts step for now. We have to sit together and think of a UI strategy (also including the future merge of doc and edit-doc pages).

src/components/pages/doc.js
adrianaluchian: True, or then keep it simple then and extract an `alerts` catalog-type module with all alerts e.g.
```
const ALERT: {
   name: {
        message: ...,
        type: ...
   }
```
?

ahelmberger: Should I do something regarding this discussion on this MR?

adrianaluchian: If you feel it's easy and can do it here, yes, if not let's make a ticket.

ahelmberger: It's at least not hard :-) Then I will extract the global alerts (i.e. insufficient profile) to a separate module.

ahelmberger: I've extracted the global alerts into a module, please re-check.

adrianaluchian (approved):

PR #59, opened 2021-12-02
EDU-140 Show external users in separate tab

ahelmberger:
Resolves https://educandu.atlassian.net/browse/EDU-140

### DIFF
--- a/src/components/pages/users.js
+++ b/src/components/pages/users.js
@@ -3,38 +3,59 @@ import firstBy from 'thenby';
 import Page from '../page.js';
 import autoBind from 'auto-bind';
 import PropTypes from 'prop-types';
-import { Table, Popover } from 'antd';
-import { ROLE } from '../../domain/role.js';
 import Logger from '../../common/logger.js';
+import { ROLE } from '../../domain/role.js';
+import { Table, Popover, Menu } from 'antd';
 import { withTranslation } from 'react-i18next';
 import { inject } from '../container-context.js';
 import errorHelper from '../../ui/error-helper.js';
 import UserRoleTagEditor from '../user-role-tag-editor.js';
 import UserApiClient from '../../services/user-api-client.js';
-import { SaveOutlined, CloseOutlined } from '@ant-design/icons';
 import CountryFlagAndName from '../localization/country-flag-and-name.js';
 import UserLockedOutStateEditor from '../user-locked-out-state-editor.js';
 import { userShape, translationProps } from '../../ui/default-prop-types.js';
 
 const logger = new Logger(import.meta.url);
 
-const availableRoles = [
-  { name: ROLE.user, isReadonly: false },
-  { name: ROLE.admin, isReadonly: true }
-];
+const MenuItem = Menu.Item;
+
+const availableRoles = Object.values(ROLE);
+
+const TABS = {
+  internalUsers: 'internal-users',
+  externalUsers: 'external-users'
+};
+
+function separateInternalAndExternalUsers(allUsers) {
+  const internalUsers = [];
+  const externalUsers = [];
+  for (const user of allUsers) {
+    const matches = (/^external\/(.+)$/).exec(user.provider);
+    if (matches) {
+      externalUsers.push({
+        ...user,
+        importSource: matches[1]
+      });
+    } else {
+      internalUsers.push(user);
+    }
+  }
+
+  return { internalUsers, externalUsers };
+}
 
 class Users extends React.Component {
   constructor(props) {
     super(props);
     autoBind(this);
+
     this.state = {
-      users: props.initialState,
-      changedLockOutStates: {},
-      changedRoles: {},
-      isDirty: false
+      ...separateInternalAndExternalUsers(props.initialState),
+      currentTab: TABS.internalUsers,
+      isSaving: false
     };
 
-    this.columns = [
+    this.internalUserTableColumns = [
       {
         title: () => this.props.t('username'),
         dataIndex: 'username',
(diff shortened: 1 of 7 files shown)


src/components/pages/users.js
```diff
@@ -69,6 +90,31 @@ class Users extends React.Component {
...
+        key: 'username',
+        sorter: firstBy('username'),
+        render: this.renderUsername
+      }, {
+        title: () => this.props.t('email'),
+        dataIndex: 'email',
+        key: 'email',
+        sorter: firstBy('email')
+      }, {
+        title: () => this.props.t('id'),
```
laurentiu-ilici, line 107: Why do we need this column? I see no value for the user.

src/components/pages/users.js
```diff
@@ -69,6 +90,31 @@ class Users extends React.Component {
...
+
+    this.externalUserTableColumns = [
+      {
+        title: () => this.props.t('username'),
+        dataIndex: 'username',
+        key: 'username',
+        sorter: firstBy('username'),
+        render: this.renderUsername
+      }, {
+        title: () => this.props.t('email'),
```
laurentiu-ilici, line 102: This column is always empty (for now, I guess it will have data once we have Shibolleth), is it worth keeping?

src/components/pages/users.js
```diff
@@ -3,38 +3,59 @@ import firstBy from 'thenby';
...
+const MenuItem = Menu.Item;
+
+const availableRoles = Object.values(ROLE);
+
+const TABS = {
+  internalUsers: 'internal-users',
+  externalUsers: 'external-users'
+};
+
+function separateInternalAndExternalUsers(allUsers) {
```
laurentiu-ilici, line 29: You usually separate "something from something" :).

laurentiu-ilici (approved):

src/components/pages/users.js
adrianaluchian: That would sound to me more like extracting, when you separate A from B you keep/favor A.
Separate A and B sounds correct to me, you put in a list and you return it as split/separated in 2.

src/components/pages/users.js
```diff
@@ -121,116 +167,106 @@ class Users extends React.Component {
...
 
-    try {
-      for (const [userId, newLockedOut] of Object.entries(changedLockOutStates)) {
-        /* eslint-disable-next-line no-await-in-loop */
-        await userApiClient.saveUserLockedOutState({ userId, lockedOut: newLockedOut });
-      }
+    this.setState(prevState => ({
+      ...prevState,
+      internalUsers: prevState.internalUsers.map(usr => usr._id === user._id ? { ...user, roles: newRoles } : usr),
+      isSaving: true
```
adrianaluchian, line 198: Extracting 
`prevState.internalUsers.map(usr => usr._id === user._id ? { ...user, roles: newRoles } : usr)`
in a const would make this a bit more readable I find. Maybe also less confusing between `usr` and `user` to call `usr` -> `internalUser` then the mapping gets simpler to understand from one look.

src/components/pages/users.js
```diff
@@ -121,116 +167,106 @@ class Users extends React.Component {
...
-        await userApiClient.saveUserRoles({ userId, roles: newRoles });
-      }
+    try {
+      await userApiClient.saveUserRoles({ userId: user._id, roles: newRoles });
     } catch (error) {
       errorHelper.handleApiError({ error, logger, t });
+      this.setState(prevState => ({
+        ...prevState,
+        internalUsers: prevState.internalUsers.map(usr => usr._id === user._id ? { ...user, roles: oldRoles } : usr)
+      }));
```
adrianaluchian, line 208: Gleichfalls hier ☺️

src/components/pages/users.js
```diff
@@ -121,116 +167,106 @@ class Users extends React.Component {
...
-  async handleCancelClick() {
-    const { userApiClient } = this.props;
+  async handleLockedOutStateChange(user, newLockedOut) {
+    const { userApiClient, t } = this.props;
+    const oldLockedOut = user.lockedOut;
 
-    const { users } = await userApiClient.getUsers();
+    this.setState(prevState => ({
+      ...prevState,
+      internalUsers: prevState.internalUsers.map(usr => usr._id === user._id ? { ...user, lockedOut: newLockedOut } : usr),
```
adrianaluchian, line 220: Now I'm thinking a generic `extractInternalUsers` function might help 😅

src/components/pages/users.js
ahelmberger: True, removed

src/components/pages/users.js
ahelmberger: Removed it, we can add it again after introducing Shibboleth (so maybe never)

src/components/pages/users.js
ahelmberger: I renamed it to `split...`, maybe then it's clearer that both subsets are equally important

src/components/pages/users.js
laurentiu-ilici: I actually hope we get that feature, because it's something that can help us strategically to get other Unis interested.

src/components/pages/users.js
ahelmberger: I tried several variations and ended up with a helper function `replaceUserById(users, newUser)` that makes the line (and the other three occurrences) less cryptic.

src/components/pages/users.js
ahelmberger: Siehe oben

src/components/pages/users.js
adrianaluchian: Sounds good!

adrianaluchian (approved):

PR #60, opened 2021-12-02
EDU-139 implement document import task processor part 2

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-139

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,6 +9,9 @@ export default {
     '<rootDir>/coverage/',
     '<rootDir>/dist/',
     '<rootDir>/test-app/',
+    '<rootDir>/src/components/',
+    '<rootDir>/src/plugins/',
+    '<rootDir>/src/stores/collection-specs/',
     '<rootDir>/src/test-helper\\.js',
     '<rootDir>/src/test-setup\\.js'
   ],
--- a/src/repositories/cdn.js
+++ b/src/repositories/cdn.js
@@ -33,6 +33,10 @@ class Cdn {
     return buffer.toString(encoding);
   }
 
+  objectExists(objectName) {
+    return this.s3Client.objectExists(this.bucketName, objectName);
+  }
+
   uploadObject(objectName, filePath, metadata) {
     const defaultMetadata = this._getDefaultMetadata();
     const stream = fs.createReadStream(filePath);
--- a/src/repositories/cdn.js
+++ b/src/repositories/cdn.js
@@ -43,9 +43,8 @@ class Cdn {
 
   async uploadObjectFromUrl(objectName, url) {
     const body = await superagent.get(url).then(res => res.body);
-    const sanitizedObjectName = objectName.replace(/\\/g, '/');
-    const contentType = mime.getType(sanitizedObjectName) || defaultContentType;
-    return this.s3Client.upload(this.bucketName, sanitizedObjectName, body, contentType);
+    const contentType = mime.getType(objectName) || defaultContentType;
+    return this.s3Client.upload(this.bucketName, objectName, body, contentType);
   }
 
   uploadEmptyObject(objectName, metadata) {
--- a/src/repositories/cdn.js
+++ b/src/repositories/cdn.js
@@ -76,12 +76,12 @@ class Cdn {
     return { createdon: new Date().toISOString() };
   }
 
-  static create({ endpoint, region, accessKey, secretKey, bucketName, rootUrl, httpClient }) {
+  static create({ endpoint, region, accessKey, secretKey, bucketName, rootUrl }) {
     const s3Client = endpoint.includes('.amazonaws.com')
       ? new AwsSdkS3Client({ endpoint, region, accessKey, secretKey })
       : new MinioS3Client({ endpoint, region, accessKey, secretKey });
 
-    return Promise.resolve(new Cdn(s3Client, bucketName, region, rootUrl, httpClient));
+    return Promise.resolve(new Cdn(s3Client, bucketName, region, rootUrl));
   }
 }
 
--- a/src/repositories/cdn.js
+++ b/src/repositories/cdn.js
@@ -1,6 +1,7 @@
 import fs from 'fs';
 import mime from 'mime';
 import Stream from 'stream';
+import superagent from 'superagent';
 import Logger from '../common/logger.js';
 import MinioS3Client from './minio-s3-client.js';
 import AwsSdkS3Client from './aws-sdk-s3-client.js';
@@ -40,6 +41,13 @@ class Cdn {
     return this.s3Client.upload(this.bucketName, sanitizedObjectName, stream, contentType, { ...defaultMetadata, ...metadata });
   }
 
+  async uploadObjectFromUrl(objectName, url) {
+    const body = await superagent.get(url).then(res => res.body);
+    const sanitizedObjectName = objectName.replace(/\\/g, '/');
+    const contentType = mime.getType(sanitizedObjectName) || defaultContentType;
+    return this.s3Client.upload(this.bucketName, sanitizedObjectName, body, contentType);
+  }
+
   uploadEmptyObject(objectName, metadata) {
     const defaultMetadata = this._getDefaultMetadata();
     const sanitizedObjectName = objectName.replace(/\\/g, '/');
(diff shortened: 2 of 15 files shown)


src/services/document-import-task-processor.js
```diff
@@ -30,10 +39,43 @@ export class DocumentImportTaskProcessor {
...
+          originUrl: `${baseUrl}${docUrl}`,
+          appendTo: previousRevision
+            ? {
+              key: revision.key,
+              ancestorId: previousRevision._id
+            }
+            : null
+        };
+
+        await this.documentService.copyDocumentRevision({ doc: mappedRevision, user, databaseSession: session });
```
laurentiu-ilici, line 75: We could consider a cancellation after this step. You basically created a full new revision.

laurentiu-ilici (approved):
Works nicely! I've merged the main into this branch so to test the UI as well :).

src/services/document-import-task-processor.js
adrianaluchian: Cancelling here would make this into partial import that affects the user.
The import will not have been finalised but the user will already see new document revisions.
This is the reason why I've wrapped this iteration in a transaction. And if the graceful shutdown period would not give it enough time to finish the loop, then we would not persist a partial state and fully resume processing.

laurentiu-ilici: I know the user would see a partial import, mainly I'm thinking that we saw documents with 100+ revisions and that could take a while to process. If you import 50 of them now and mark the import as failed, the user can still see that it was failed and resume importing from where it left off.

adrianaluchian: By user here you mean just the admin. I mean all users, which is the most common case, the people that see the documents. I would not go for a partial state that was not specifically intended by the importer 🤷‍♀️ 
The case we are talking about is an edge case either way, so I'd treat it as one and opt for all-or-nothing.

adrianaluchian:
> Works nicely! I've merged the main into this branch so to test the UI as well :).

Ah you didn't just mean locally, you pushed. Better let the PR author do that, in case there are local changes to still be pushed. Plus, autonomy of own work and all that 😊

jest.config.js
```diff
@@ -9,6 +9,9 @@ export default {
     '<rootDir>/coverage/',
     '<rootDir>/dist/',
     '<rootDir>/test-app/',
+    '<rootDir>/src/components/',
+    '<rootDir>/src/plugins/',
+    '<rootDir>/src/stores/collection-specs/',
```
ahelmberger, line 14: Love it

laurentiu-ilici:
> > Works nicely! I've merged the main into this branch so to test the UI as well :).
> 
> Ah you didn't just mean locally, you pushed. Better let the PR author do that, in case there are local changes to still be pushed. Plus, autonomy of own work and all that 😊

Ok, I won't push again, sorry.

src/repositories/cdn.js
```diff
@@ -40,6 +41,13 @@ class Cdn {
     return this.s3Client.upload(this.bucketName, sanitizedObjectName, stream, contentType, { ...defaultMetadata, ...metadata });
   }
 
+  async uploadObjectFromUrl(objectName, url) {
+    const body = await superagent.get(url).then(res => res.body);
+    const sanitizedObjectName = objectName.replace(/\\/g, '/');
```
ahelmberger, line 46: Should not the imported object have the exact same name as the source object? Also, in which case would we have a backslash as a separator? :thinking:

adrianaluchian: True, I ran the same sanitizing as for `uploadObject` but `uploadObjectFromUrl` is used for import only and should keep the name as is, though from the perspective of the cdn module the reason for this difference is not clear.
Will remove anyway.

adrianaluchian: Done!

src/services/document-import-task-processor.js
ahelmberger: Sorry to swoop in after this was already resolved, but I think we could find a secure middle ground here that also caters better to documents with a huge amount of revisions or files: We can split the whole document import process into three stages:

1. Ensure all users are created
    Doesn't take long, cancellation afterwards (or even in-between) is no problem, they will be just re-imported on the next attempt

2. Copy all CDN files
    Can potentially take very long, has to be interruptible! But cancellation in-between is no problem, they will be just re-imported on the next attempt, so we should ask if cancellation is requested after each individual file copy process

3. Insert all revisions into the db and create the new "latest" document. This has to be transactional and with a document lock, cancellation of this task in-between and afterwards must not be possible anymore.

The good news is: Until we reach stage 3 everything else is kind of idempotent (if you consider overwriting an existing file as idempotent).

So in my opinion, the code above should be rearranged regarding 2 aspects:
* First we import all the files (with option to cancel after each), then we do the final db transaction
* Check that the revision boundary between the revision returned by `getCurrentDocumentRevisionByKey` is the correct ancestor to the first newly imported revision (there could have been an import in between and they would not match anymore)

src/services/document-service.js
```diff
@@ -342,23 +262,133 @@ class DocumentService {
...
+    });
+  }
+
+  _copySections({ sections }) {
+    return sections.map(section => ({
+      revision: section.revision,
+      key: section.key,
+      deletedOn: null,
+      deletedBy: null,
+      deletedBecause: null,
```
ahelmberger, line 380: What happens in case we import a document with sections that have been deleted?

src/services/document-import-task-processor.js
adrianaluchian: Oh this sounds good, basically taking the CDN copy out of the transaction as it is repeatable and has no impact on the user. And keeping the revisions as a transaction (plus the doc lock, will add that).

Will make the changes, thanks!

src/services/document-service.js
adrianaluchian: D'oh, we loose this info, but of course.
Will change to do a `sections.map(section => { ...section })` then.
I'd still keep this method though so that `_createDocumentRevision` doesn't need to do in internal `if`-`else`

ahelmberger: or just `return sections.map(s => cloneDeep(s));`

ahelmberger: I agree on the no `if`-`else`

src/services/document-service.js
```diff
@@ -372,6 +402,7 @@ class DocumentService {
       tags: doc.tags || [],
       archived: doc.archived || false,
       origin: doc.origin || DOCUMENT_ORIGIN.internal,
+      originUrl: doc.originUrl || '',
```
ahelmberger, line 405: Shouldn't we have a migration for that (setting it to `null` for internal docs/revs)?

src/services/document-import-task-processor.js
adrianaluchian: Made the changes as you suggested, please double check 🙏

adrianaluchian: Discovered a conflict with the lock, I will fix that first and the ask you to check again.

adrianaluchian: Done, ready to be checked again.

src/services/document-service.js
adrianaluchian: Changed!

src/services/document-service.js
adrianaluchian: Yes we should, added.

src/services/document-service.js
```diff
@@ -248,135 +360,89 @@ class DocumentService {
...
+            userId: revision.createdBy,
+            order,
+            restoredFrom: revision.restoredFrom,
+            sections: cloneDeep(revision.sections)
+          });
 
-      const revisionsToUpdate = [];
+          logger.info(`Saving new document revision with id ${newDocumentRevision._id}`);
+          // eslint-disable-next-line no-await-in-loop
+          await this.documentRevisionStore.save(newDocumentRevision, { session });
```
ahelmberger, line 403: The only improvement I can still see here, would be to not save each revision within the loop, but to call `insertMany` after the loop is done (and the `newDocumentRevisions` array is filled), but I don't think it will make it that much slower, it's not a performance critical path anyway, so I'll approve regardless for what you decide.

ahelmberger (approved):

src/services/document-service.js
adrianaluchian: Yes we said we wanted to do that actually. I'll want to do that 👍

adrianaluchian: Changed.

PR #61, opened 2021-12-03
EDU-145 Add polling on batch view

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-145

### DIFF
--- a/src/components/pages/import-batch-view.js
+++ b/src/components/pages/import-batch-view.js
@@ -15,7 +15,7 @@ import { isTaskSuccessful, taskStatusSorter, doesTaskHaveErrors } from '../../ut
 import { CloudDownloadOutlined, CloudSyncOutlined, WarningOutlined, CheckOutlined, ExclamationCircleOutlined, SyncOutlined } from '@ant-design/icons';
 
 const { Panel } = Collapse;
-const POLL_INTERVAL_IN_MS = 1000;
+const POLL_INTERVAL_IN_MS = 1500;
 const logger = new Logger(import.meta.url);
 
 function ImportBatchView({ initialState }) {
--- a/src/components/pages/import-batch-view.js
+++ b/src/components/pages/import-batch-view.js
@@ -31,7 +31,7 @@ function ImportBatchView({ initialState }) {
 
   useEffect(() => {
     if (isCompletedImport) {
-      return () => {};
+      return;
     }
 
     const interval = setInterval(async () => {
@@ -46,6 +46,7 @@ function ImportBatchView({ initialState }) {
       }
     }, POLL_INTERVAL_IN_MS);
 
+    // eslint-disable-next-line consistent-return
     return () => {
       if (interval) {
         clearInterval(interval);
--- a/src/components/pages/import-batch-view.js
+++ b/src/components/pages/import-batch-view.js
@@ -51,8 +51,7 @@ function ImportBatchView({ initialState }) {
         clearInterval(interval);
       }
     };
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  });
 
   const renderTaskStatus = (processed, task) => {
     if (!processed) {
--- a/src/components/pages/import-batch-view.js
+++ b/src/components/pages/import-batch-view.js
@@ -1,23 +1,58 @@
-import React from 'react';
 import Page from '../page.js';
 import PropTypes from 'prop-types';
+import Logger from '../../common/logger.js';
 import { useTranslation } from 'react-i18next';
+import React, { useEffect, useState } from 'react';
+import { useService } from '../container-context.js';
 import { useDateFormat } from '../language-context.js';
+import { handleApiError } from '../../ui/error-helper.js';
 import { getImportedArticleUrl } from '../../utils/urls.js';
 import { DOCUMENT_IMPORT_TYPE } from '../../common/constants.js';
+import ImportApiClient from '../../services/import-api-client.js';
 import { Table, Row, Space, Tooltip, Collapse, List } from 'antd';
 import { importBatchDetailsShape } from '../../ui/default-prop-types.js';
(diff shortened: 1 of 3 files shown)


src/components/pages/import-batch-view.js
```diff
@@ -1,24 +1,58 @@
...
+
+  const [batch, setBatch] = useState(initialState.batch);
 
-  const { batch } = initialState;
+  const id = batch._id;
+  const isCompletedImport = batch.progress === 1;
   const { hostName, allowUnsecure } = batch.batchParams;
 
+  useEffect(() => {
+    if (isCompletedImport) {
```
adrianaluchian, line 33: I find this early return a bit counterintuitive here in this case, 
I'd suggest to turn it around and wrap the code block in `if (!isCompletedImport) { .. }` and avoid this fake cleanup return to just satisfy the linter.
Or just doing a `return;` and suppressing the linter, though I'd go for the first option personally.

src/server/import-controller.js
```diff
@@ -60,6 +60,14 @@ class ImportController {
...
 
+    router.get('/api/v1/imports/:batchId', [needsPermission(permissions.MANAGE_IMPORT)], async (req, res) => {
+      const { batchId } = req.params;
+      const rawBatch = await this.importService.getImportBatchDetails(batchId);
+      const batch = await this.clientDataMapper.mapImportBatch(rawBatch, req.user);
+
+      res.send({ batch });
+    });
+
     router.post('/api/v1/imports/batch', [jsonParser, needsPermission(permissions.MANAGE_IMPORT), validateBody(postImportBatchBodySchema)], async (req, res) => {
```
adrianaluchian, line 71: Between this `POST` and the above `GET` we should have gone with
`POST /api/v1/imports/batches`
`GET /api/v1/imports/batches/:batchId`
🤦‍♀️  
Oh well, I guess it's too much of a hassle now to change and update both other websites for this.

src/components/pages/import-batch-view.js
```diff
@@ -1,24 +1,58 @@
...
 import { DOCUMENT_IMPORT_TYPE } from '../../common/constants.js';
+import ImportApiClient from '../../services/import-api-client.js';
 import { Table, Row, Space, Tooltip, Collapse, List } from 'antd';
 import { importBatchDetailsShape } from '../../ui/default-prop-types.js';
-import { CloudDownloadOutlined, CloudSyncOutlined, WarningOutlined, CheckOutlined, ExclamationCircleOutlined, SyncOutlined } from '@ant-design/icons';
 import { isTaskSuccessful, taskStatusSorter, doesTaskHaveErrors } from '../../utils/task-utils.js';
+import { CloudDownloadOutlined, CloudSyncOutlined, WarningOutlined, CheckOutlined, ExclamationCircleOutlined, SyncOutlined } from '@ant-design/icons';
 
 const { Panel } = Collapse;
+const POLL_INTERVAL_IN_MS = 1000;
```
adrianaluchian, line 18: Maybe 1s is a bit too often here 🤔 For a big payload especially.

src/components/pages/import-batch-view.js
laurentiu-ilici: I wanted to avoid the unnecessary indentation, I'll do the first one.

src/server/import-controller.js
laurentiu-ilici: 🤦 I could make this one  GET /api/v1/imports/batch/:batchId for consistency, but it would be correct to have "batches"

src/components/pages/import-batch-view.js
laurentiu-ilici: It felt alright, I had it with 2 seconds and it felt like an eternity :)). I can put it higher again. Suggest a value (try it locally with "all of stage" data :)).

src/components/pages/import-batch-view.js
laurentiu-ilici: I went for the second option because the linter would complain anyway on the first option and we would get the indentation for nothing as well.

src/components/pages/import-batch-view.js
adrianaluchian: Then let's split the difference and make it 1.5s?

src/server/import-controller.js
laurentiu-ilici: We can do one further patch at the very end to just update to "batches" as a last thing in the epic. What do you think?

adrianaluchian: Or then let's correct both endpoints and do another educandu update on the websites. We would need to do that anyway soon.

src/components/pages/import-batch-view.js
laurentiu-ilici: Sure, done

src/server/import-controller.js
laurentiu-ilici: Done, shall we make the updates?

adrianaluchian: Um, that can only be done after we merge the PR that brings the update, right? 🙄 
Or did we misunderstand each other?

laurentiu-ilici: Well, yes, after this thing is merged. But I think we don't need to rush it since it works even if the systems are running different versions of educandu. It's a breaking change only for the "current system" not for an external API

adrianaluchian (approved):

src/components/pages/import-batch-view.js
```diff
@@ -1,24 +1,59 @@
...
+      }
+    }, POLL_INTERVAL_IN_MS);
+
+    // eslint-disable-next-line consistent-return
+    return () => {
+      if (interval) {
+        clearInterval(interval);
+      }
+    };
+  });
```
ahelmberger, line 55: Here's the (empty) dependency array missing, so this will be called on each update and create a new interval!

laurentiu-ilici: Ah, I deleted it because of reason... I'll add it back

PR #62, opened 2021-12-03
EDU-146 Fix payload too large issue

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-146

### DIFF
--- a/src/server/import-controller.js
+++ b/src/server/import-controller.js
@@ -12,7 +12,7 @@ import { getImportsQuerySchema, postImportBatchBodySchema } from '../domain/sche
 
 const { NotFound } = httpErrors;
 
-const jsonParser = express.json();
+const jsonParserLargePayload = express.json({ limit: '2MB' });
 
 class ImportController {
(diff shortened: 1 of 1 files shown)


src/server/import-controller.js
```diff
@@ -12,7 +12,7 @@ import { getImportsQuerySchema, postImportBatchBodySchema } from '../domain/sche
 
 const { NotFound } = httpErrors;
 
-const jsonParser = express.json();
+const jsonParserLargePayload = express.json({ limit: '2MB' });
```
adrianaluchian, line 15: Cool, what was the size of the batch that errored out in your case?

laurentiu-ilici: It was almost 200kb, I saw we have this 2 MB for the docs as well so I took that. We could lower it, but I wouldn't overfit.

adrianaluchian: No no, actually wanted to make sure we don't soon reach this limit again, all good.

adrianaluchian (approved):

PR #63, opened 2021-12-03
EDU-145 Reintroduce the dependency array

laurentiu-ilici:


src/components/pages/import-batch-view.js
```diff
@@ -52,7 +52,9 @@ function ImportBatchView({ initialState }) {
         clearInterval(interval);
       }
     };
-  });
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
ahelmberger, line 57: Actually, now I see that you are using multiple variable from the outside in this side effect, that's why the linter complains. Actually, `isCompletedImport` is not used outside and can be moved into the hook body. Then you have to specify the used dependencies (`[id, t, importApiClient]`). ESLint should have told you that :thinking:

laurentiu-ilici: That's why I ignore the linter, I don't want this to trigger each time the basch changes (that would create multiple intervals). I only want this effect to trigger when I first load the page. So the empty array should prevent multiple fires of the effect

ahelmberger: The batch is not part of the array, it's the id (never changes), the client (never changes), and `t` (changes when the user changes the language), so the linter is correct.

laurentiu-ilici: I'm not saying the linter is incorrect, I'm saying that in this particular case we don't want to trigger this effect again. The linter guards against us forgetting dependencies and not updating things when they change right? but here we are forgrtting it on purpose.

The isCompletedImport is inteffered from the progress, this one will change on each interval which will trigger the hook if we give it the dependencies.

ahelmberger: I'll try to rephrase because I think we have a misunderstanding here: If you don't provide `t` as a dependency, the user will get the error message in the language that was set when the page first loaded, even in case that it was changed in the meantime. `id` and `importApiClient` will never change and can therefore be safely also provided as dependencies (because they are). I understand now that `isCompletedImport` is used as a flag to trigger `setInterval` and therefore has to be declared outside of the effect (opposite to what I suggested in my first comment). Even so, you can provide `isCompletedImport` as a dependency, too, because a) it is a dependency and b) we actually do want the loop to start if it's true and nothing to happen when it's false. I checked out the branch and if I do these things in the code I don't have to disable linting. This linting rule is actually one of the most useful in my opinion and unless there is a scenario where it's really impossible to adhere to it, I think we should stick to it. Or am I overlooking something?

src/components/pages/import-batch-view.js
```diff
@@ -25,34 +25,41 @@ function ImportBatchView({ initialState }) {
 
   const [batch, setBatch] = useState(initialState.batch);
 
-  const id = batch._id;
-  const isCompletedImport = batch.progress === 1;
+  const { _id: id, progress } = batch;
```
ahelmberger, line 28: Variable `progress` is already declared in line 78

src/components/pages/import-batch-view.js
```diff
@@ -25,34 +25,41 @@ function ImportBatchView({ initialState }) {
...
         handleApiError({ error, logger, t });
+        if (nextTimeout) {
+          clearTimeout(nextTimeout);
+        }
       }
-    }, POLL_INTERVAL_IN_MS);
+    };
 
-    // eslint-disable-next-line consistent-return
+    if (!nextTimeout) {
```
ahelmberger, line 54: This condition can never be false as `nextTimeout` is set to `null` two statements above (line 32) with no async code and therefore no thread change in between.

ahelmberger (approved):

PR #65, opened 2021-12-06
EDU-141 Segregate manual migrations from automatic migrations

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-141


gulpfile.js
```diff
@@ -10,16 +10,20 @@ import less from 'gulp-less';
 import csso from 'gulp-csso';
```
adrianaluchian, line 10: Hmm, I thought we wanted to separate into schema and content migrations 🤔 That's also how we wrote the ticket initially. And of course each category can have manual migrations, though schema ones likely will never need manual run.

The idea was to have only the schema migrations yield the database schema hash.

adrianaluchian: My suggestion would be that we introduce a way (by file naming, or by flag in the `migrations` db collection) in which we mark the migration type. And that we generate the hash from the run migrations of type `schema`

laurentiu-ilici: Somehow now I have doubts about the splitting of schema vs content migrations. Manual migrations are just those that run for a long time (regardless if they do some content update or a schema change). For example now I have to do a migration that changes the schema and updates the content but it should be very quick since it just migrates the settings collection (I would not introduce 2 migrations for that). I think the way this is made might actually work well (although it's not what the ticket says).

adrianaluchian: The whole discussion started from the generated hash. We definitely need to generate that in a reliable way that ensures the DBs in the import process match 🤔

laurentiu-ilici: Yes, I'm just not sure that we need to separate schema changes from content as long as everything runs quickly. The migrations table should always register all ran migrations so it can generate the right hash.

adrianaluchian: Maybe it would have been good to discuss this ticket a bit in the office.
Shall we leave it for not and take the discussion live?

ahelmberger: I have reworked this branch according to the requirements we clarified, please have another look!

adrianaluchian: Looks very good to me 👍

adrianaluchian (approved):

PR #69, opened 2021-12-07
EDU-149 Improve UX on import pages

ahelmberger:
Unfortunately it's not possible to prevent antd table from a full re-render on each checkbox click (there are a lot of issues on their GH repo, but they're not fixing it). I managed to cut down the lag maybe by 50% but it's still a very bad experience. The only remedies would be either paging or virtualizing the table.

Closes https://educandu.atlassian.net/browse/EDU-149

### DIFF
--- a/src/components/pages/import-batches.js
+++ b/src/components/pages/import-batches.js
@@ -89,8 +89,17 @@ function ImportBatches({ initialState, PageTemplate }) {
       <div className="ImportBatchesPage">
         <h1>{t('pageNames:importBatches')}</h1>
         {sources.map(source => (
-          <Collapse key={source.importSourceName} className="ImportBatchesPage-sourceCollapse" defaultActiveKey={source.batches.length ? '1' : null}>
-            <Panel header={header(source)} className="ImportBatchesPage-batchTablePanel" extra={getExtra(source)} key="1">
+          <Collapse
+            key={`collapse-${source.importSourceName}`}
+            className="ImportBatchesPage-sourceCollapse"
+            defaultActiveKey={source.batches.length ? `panel-${source.importSourceName}` : null}
+            >
+            <Panel
+              header={header(source)}
+              className="ImportBatchesPage-batchTablePanel"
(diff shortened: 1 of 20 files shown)


laurentiu-ilici (commented):
Your changes seem to be fine but I think I found a problem with the user import. I'm still looking  into it and trying to understand. Please don't merge yet.

src/components/pages/import-batches.js
```diff
@@ -82,27 +82,25 @@ function ImportBatches({ initialState, PageTemplate }) {
...
   return (
     <PageTemplate alerts={alerts}>
       <div className="ImportBatchesPage">
         <h1>{t('pageNames:importBatches')}</h1>
-        <Collapse defaultActiveKey={defaultActiveKey}>
-          {sources.map(source => (
-            <Panel header={header(source)} className="ImportBatchesPage-batchTablePanel" key={source.importSourceName} extra={getExtra(source)} >
+        {sources.map(source => (
+          <Collapse key={source.importSourceName} className="ImportBatchesPage-sourceCollapse" defaultActiveKey={source.batches.length ? '1' : null}>
+            <Panel header={header(source)} className="ImportBatchesPage-batchTablePanel" extra={getExtra(source)} key="1">
```
adrianaluchian, line 93: The keys could be `<Collapse key=\`collapse-${source.importSourceName}\`` and `<Panel key=\`panel-${source.importSourceName}\``.

adrianaluchian:
> Your changes seem to be fine but I think I found a problem with the user import. I'm still looking into it and trying to understand. Please don't merge yet.
The problem is unrelated, I think we'll be able to merge.

adrianaluchian:
Works well from my perspective 👍  I think we can live with the checkbox delay now for this particular admin-facing table.

src/components/pages/import-batches.js
ahelmberger: OK, done

adrianaluchian (approved):

PR #75, opened 2021-12-09
EDU-151 Introduce minus operator in search

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-151

### DIFF
--- a/src/components/pages/search.js
+++ b/src/components/pages/search.js
@@ -18,7 +18,6 @@ function Search({ initialState, PageTemplate }) {
   const { language } = useLanguage();
   const { docs } = initialState;
   const { query } = useRequest();
-  const decodedQuery = decodeURIComponent(query.query);
 
   const languageNameProvider = useService(LanguageNameProvider);
   const languageData = languageNameProvider.getData(language);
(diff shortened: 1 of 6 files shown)


src/components/pages/search.js
```diff
@@ -28,28 +27,18 @@ function Search({ initialState, PageTemplate }) {
     () => docs
       .map(doc => ({
         ...doc,
+        tags: [...new Set(doc.tags)], // Some docs have duplicate tags we don't want to render
```
laurentiu-ilici, line 30: Why keep the code when we have this as a next thing?

src/utils/urls.js
```diff
@@ -127,20 +127,7 @@ export function createFullyQualifiedUrl(pathname) {
 }
 
 export function getSearchUrl(query) {
-  const url = new URL(document.location);
-  const keys = [];
-  for (const key of url.searchParams.keys()) {
```
laurentiu-ilici, line 132: We need this for the case when users copy/paste a query. This was an original requirement.

src/components/pages/search.js
ahelmberger: We don't have it yet, so currently we have duplicate key errors

src/services/document-service.js
```diff
@@ -96,35 +96,39 @@ class DocumentService {
...
-      .split(/\s+/)
-      .map(tag => escapeStringRegexp(tag.toLowerCase()))
-      .filter(tag => tag.length > 2));
+  async getDocumentsByTags(searchQuery) {
+    const tokens = searchQuery.trim().split(/\s+/);
 
-    if (!searchTags.size) {
+    const positiveTokens = new Set(tokens
+      .filter(token => !token.startsWith('-'))
+      .filter(token => token.length > 2)
```
adrianaluchian, line 104: We could put the `.filter(token => token.length > 2)` up in `tokens` since it filters out invalid tags for all further cases.

src/utils/urls.js
laurentiu-ilici: Never mind I think this was a problem when we had the exact tags, not the full text.

src/services/document-service.js
adrianaluchian: Ah never mind, does not cover the `-token` cases 🤦

ahelmberger: But the minus tags have to get the minus removed first before measuring the length?

laurentiu-ilici (approved):

src/utils/urls.js
ahelmberger: Ah, I didn't refresh :-)

src/services/document-service.spec.js
```diff
@@ -817,9 +817,10 @@ describe('document-service', () => {
         language: 'en'
       });
 
+      // This archived document should never be matched:
```
adrianaluchian, line 820: Could we turn this comment into an assert? 🙏

ahelmberger: True, that would be better, will do

src/utils/urls.js
laurentiu-ilici: yeah, I tried to reproduce the problem but I really think I set this up because of the tag search and the it got left over. Anyway on the search page if you come with a query string query=yourQuery&extraParam=something it returns a 400, I'm not sure what was the case back then where that could have happened when redirecting.

src/services/document-service.spec.js
ahelmberger: Done!

adrianaluchian (approved):

PR #78, opened 2021-12-10
EDU-161 Add message for missing import sources

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-161

### DIFF
--- a/src/components/pages/import-batches.yml
+++ b/src/components/pages/import-batches.yml
@@ -37,3 +37,6 @@ createImport:
 importsHeaderPrefix:
   en: Source
   de: Quelle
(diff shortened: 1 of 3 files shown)


src/components/pages/import-batches.yml
```diff
@@ -37,3 +37,6 @@ createImport:
 importsHeaderPrefix:
   en: Source
   de: Quelle
+noImportSourcesConfigures:
```
ahelmberger, line 40: Just a small typo: "Configures" instead of "Configured". I'll approve, so you can fix and merge immediately

ahelmberger (approved):

PR #80, opened 2021-12-10
EDU-150 fix image tiles

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-150


src/plugins/image-tiles/editing/tile-editor.js
```diff
@@ -99,17 +99,17 @@ class TileEditor extends React.Component {
         <FormItem label={t('linkSource')} {...formItemLayout}>
           <RadioGroup value={link.type} onChange={this.handleLinkTypeValueChanged}>
             <RadioButton value={LINK_TYPE.external}>{t('externalLink')}</RadioButton>
-            <RadioButton value={LINK_TYPE.article}>{t('articleLink')}</RadioButton>
+            <RadioButton value={LINK_TYPE.internal}>{t('articleLink')}</RadioButton>
```
ahelmberger, line 102: Let's also rename the translation key

migrations/manual/educandu-2021-12-10-01-migrate-image-tiles-to-use-docs.js
```diff
@@ -0,0 +1,74 @@
...
+    const doc = await cursor.next();
+    /* eslint-disable-next-line no-await-in-loop */
+    await updateFn(doc);
+    /* eslint-disable-next-line no-await-in-loop */
+    await collection.replaceOne({ _id: doc._id }, doc);
+  }
+}
+
+// eslint-disable-next-line camelcase
+export default class Educandu_2021_12_10_01_migrate_migrate_image_tiles_to_use_docs {
```
ahelmberger, line 16: migrate-migrate :smile:

migrations/manual/educandu-2021-12-10-01-migrate-image-tiles-to-use-docs.js
```diff
@@ -0,0 +1,74 @@
+async function updateAll(collection, query, updateFn) {
+  const cursor = collection.find(query);
+
+  /* eslint-disable-next-line no-await-in-loop */
+  while (await cursor.hasNext()) {
+    /* eslint-disable-next-line no-await-in-loop */
+    const doc = await cursor.next();
+    /* eslint-disable-next-line no-await-in-loop */
+    await updateFn(doc);
```
ahelmberger, line 9: Actually I think we could save a lot of time if we would update only documents that actually have changed. We could achieve that by returning it from `updateFunc`. So in case it returns `undefined` or `null` it would not call `replaceOne`. What do you think?

src/plugins/image-tiles/editing/tile-editor.js
laurentiu-ilici: Done (I think)

migrations/manual/educandu-2021-12-10-01-migrate-image-tiles-to-use-docs.js
laurentiu-ilici: Okay, done

ahelmberger (approved):

PR #81, opened 2021-12-10
EDU-166 Add Import indexes

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-166

I've added only this index since this is the only query we have so far that takes a long time. Other indexes I considered:
task.processed (no query so far)
task.taskType (no use case)
batches.batchType (we only have one type of batch so far)

### DIFF
--- a/migrations/automatic/educandu-2021-12-10-01-add-import-related-indexes.js
+++ b/migrations/automatic/educandu-2021-12-10-01-add-import-related-indexes.js
@@ -9,11 +9,21 @@ export default class Educandu_2021_12_10_01_add_import_related_indexes {
       {
         name: '_idx_batch_id_',
         key: { batchId: 1 }
+      },
+      {
+        name: '_idx_batch_id_processed_',
+        key: { batchId: 1, processed: 1 }
+      },
+      {
+        name: '_idx_task_id_processed_',
+        key: { taskId: 1, processed: 1 }
       }
     ]);
   }
 
   async down() {
     await this.db.collection('tasks').dropIndex('_idx_batch_id_');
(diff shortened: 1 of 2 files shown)


ahelmberger:
You've missed to add the index to the collection spec inside the application

migrations/automatic/educandu-2021-12-10-01-add-import-related-indexes.js
```diff
@@ -0,0 +1,19 @@
...
+        name: '_idx_batch_id_',
+        key: { batchId: 1 }
+      }
+    ]);
+  }
+
+  async down() {
+    await this.db.collection('tasks').dropIndex('_idx_batch_id_');
+  }
+}
```
adrianaluchian, line 19: Should we consider
1. a combined index `tasks._id: taskId` + `tasks.processed` ? The combined query is used in `task-processor.js`
2. a combined index `tasks.processed` + `tasks.batchId`? The combined query is used in `batch-processor.js`

Though I'm not sure such indexes would bring much in these usage cases 🤷‍♀️

ahelmberger: We would have to measure it, but maybe that's a good idea, it might speed up task processing.

laurentiu-ilici: Done

adrianaluchian (approved):

ahelmberger:
Did you miss my comment above?

laurentiu-ilici:
> 

Thanks, added. I had no idea about them, I'll remember from now on.

ahelmberger (approved):

PR #85, opened 2021-12-13
EDU-180 Fix viewing documents that do not have slugs

laurentiu-ilici:
Closes:  https://educandu.atlassian.net/browse/EDU-180

### DIFF
--- a/src/components/pages/docs.js
+++ b/src/components/pages/docs.js
@@ -246,13 +246,6 @@ class Docs extends React.Component {
         sorter: by(x => x.title)
       },
       {
-        title: t('slug'),
-        dataIndex: 'slug',
-        key: 'slug',
-        render: this.renderSlug,
-        sorter: by(x => x.slug)
-      },
-      {
         title: t('language'),
         dataIndex: 'language',
         key: 'language',
--- a/src/components/pages/docs.js
+++ b/src/components/pages/docs.js
@@ -180,16 +180,6 @@ class Docs extends React.Component {
     return <a href={urls.getDocUrl(doc.key, doc.slug)}>{doc.title}</a>;
   }
 
-  renderSlug(_value, doc) {
-    if (!doc.slug) {
-      const { t } = this.props;
-      return t('notAssigned');
-    }
-
-    const url = urls.getDocUrl(doc.key, doc.slug);
-    return <a href={url}>{url}</a>;
-  }
-
   renderUpdatedOn(_value, doc) {
     const { formatDate } = this.props;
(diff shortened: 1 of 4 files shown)


src/server/document-controller.js
```diff
@@ -60,7 +62,7 @@ class DocumentController {
         throw new NotFound();
       }
 
-      return res.redirect(301, urls.getDocUrl(doc.key, doc.slug));
+      return res.redirect(301, urls.getDocUrl(doc.key, doc.slug || UNDEFINED_SLUG));
```
ahelmberger, line 65: Rather than using this magic string you could do here:

~~~
if (!doc.slug) {
  const mappedDoc = await this.clientDataMapper.mapDocOrRevision(doc, req.user);
  return this.pageRenderer.sendPage(req, res, PAGE_NAME.doc, { currentDocOrRevision: mappedDoc, type: DOCUMENT_TYPE.document });
}

src/server/document-controller.js
```diff
@@ -72,7 +74,7 @@ class DocumentController {
         throw new NotFound();
       }
 
-      if (slug !== doc.slug) {
+      if (slug !== UNDEFINED_SLUG && slug !== doc.slug) {
```
ahelmberger, line 77: Can be reverted if you use my suggestion above

src/server/document-controller.js
adrianaluchian: Oh I second that!

laurentiu-ilici: Done. It's a bit inconsistent though.

src/server/document-controller.js
laurentiu-ilici: Reverted

src/components/pages/docs.js
```diff
@@ -180,16 +180,6 @@ class Docs extends React.Component {
     return <a href={urls.getDocUrl(doc.key, doc.slug)}>{doc.title}</a>;
   }
 
-  renderSlug(_value, doc) {
-    if (!doc.slug) {
-      const { t } = this.props;
-      return t('notAssigned');
```
adrianaluchian, line 186: You can probably remove the translation too, right?

laurentiu-ilici: I checked the other translations and forgot about this one. Thanks!

laurentiu-ilici: Removed

ahelmberger (approved):

adrianaluchian (approved):

PR #87, opened 2021-12-13
EDU-170 Ensure consistency in importing system

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-170


laurentiu-ilici:
Reviewing

adrianaluchian:
> Reviewing
No danger in this case, nobody approved already :)

src/components/pages/doc.js
```diff
@@ -157,7 +157,7 @@ function Doc({ initialState, PageTemplate }) {
   const isEditingDisabled = state.currentDocOrRevision.archived || isExternalDocument || state.type !== DOCUMENT_TYPE.document;
   const isViewRevisionsButtonVisible = state.type === DOCUMENT_TYPE.document;
   const isViewDocumentButtonVisible = state.type === DOCUMENT_TYPE.revision;
-  const isHardDeleteEnabled = state.type === DOCUMENT_TYPE.revision;
+  const disabledSectionActions = [...isExternalDocument ? [SECTION_ACTIONS.hardDelete] : []];
```
laurentiu-ilici, line 160: This construct seems backwards to me. I had to dig down 3 files from here to understand why you send "disabledActions" instead of "possible/enabledActions".

laurentiu-ilici: This also changes the fact that you can only hard delete when you see revisions. You can now delete on the document view, is this wanted? (I don't think it's a bad idea).

src/components/doc-view.js
```diff
@@ -2,8 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import SectionDisplay from './section-display.js';
 import { documentShape, documentRevisionShape } from '../ui/default-prop-types.js';
+import { SECTION_ACTIONS } from '../ui/section-actions.js';
```
ahelmberger, line 5: Ha, that sometimes happens to mee too with the auto-import :laughing:

src/services/document-service.js
```diff
@@ -246,6 +246,31 @@ class DocumentService {
...
+    }
+
+    let lock;
+    try {
+      lock = await this.documentLockStore.takeLock(documentKey);
+
+      logger.info(`Hard deleting external document '${documentKey}'`);
+
+      await this.transactionRunner.run(async session => {
+        await this.documentRevisionStore.deleteMany({ key: documentKey }, { session });
```
laurentiu-ilici, line 263: Hmm, what about the external users? I guess they could be used in other docs though.

src/components/pages/docs.yml
```diff
@@ -31,6 +28,12 @@ user:
 actions:
   en: Actions
   de: Aktionen
+clone:
+  en: Clone
+  de: Klonen
+delete:
```
ahelmberger, line 34: You might find those general button labels also already in `common.yml` (not "clone" though)

src/services/export-service.js
```diff
@@ -38,31 +38,22 @@ class ExportService {
     return this.documentStore.find(filter, { sort: lastUpdatedFirst, projection: exportableDocumentsProjection });
   }
 
-  async getDocumentExport({ key, afterRevision, toRevision }) {
+  async getDocumentExport({ key, toRevision }) {
     const revisions = (await this.documentService.getAllDocumentRevisionsByKey(key)).sort(by(d => d.order));
-
-    let nextRevisionIndex;
     const lastRevisionIndex = revisions.findIndex(revision => revision._id === toRevision);
```
laurentiu-ilici, line 43: I think the toRevision is also obsolete. 
The importer just says, "give me what you can export", and based on that it calculates what is to import, reimport and update. 
One caveat is that another revision can be inserted while the user is looking at the import screen, but I'm not sure that that just doesn't save him the time to do another import :). We currently don't allow you to import up to some revision as far as I know.

src/components/doc-view.js
adrianaluchian: 😱 Na das kann nicht sein!

src/services/export-service.js
```diff
@@ -38,31 +38,22 @@ class ExportService {
...
-    const revisionsToExport = revisions.slice(nextRevisionIndex, lastRevisionIndex + 1);
+    const revisionsToExport = revisions.slice(0, lastRevisionIndex + 1);
 
     const userIdSet = this.userService.extractUserIdSetFromDocsOrRevisions(revisionsToExport);
     const users = (await this.userService.getUsersByIds(Array.from(userIdSet)))
       .map(({ _id, username }) => ({ _id, username }));
 
     if (userIdSet.size !== users.length) {
-      throw new Error(`Was searching for ${userIdSet.size} users in document ${key} between revisions ${afterRevision} - ${toRevision}, but found ${users.length}`);
+      throw new Error(`Was searching for ${userIdSet.size} users in document ${key} before revision '${toRevision}', but found ${users.length}`);
```
ahelmberger, line 56: Maybe better "up to revision"

src/services/import-service.js
```diff
@@ -59,8 +59,12 @@ class ImportService {
       .map(exportableDocument => {
         const importedDocument = importedDocuments.find(document => document.key === exportableDocument.key);
 
+        let importType = DOCUMENT_IMPORT_TYPE.add;
+        if (importedDocument) {
+          importType = DOCUMENT_IMPORT_TYPE.update;
+        }
         if (importedDocument?.revision === exportableDocument.revision) {
```
laurentiu-ilici, line 66: I would move this to line 64 and make it a ternary between an update and reimport.

src/services/document-service.js
adrianaluchian: I thought about that but didn't want to take it too far into sanitizing here. If that becomes an issue with external users that are not referenced, it's something we can clean up when we do https://educandu.atlassian.net/browse/EDU-158

src/services/export-service.js
adrianaluchian: I specifically did not want to remove this because:
1. I see the import as a snapshot, "give me this instance of the documents"
2. We distinguish between "update" and "reimport" and this flag is what facilitates that

src/components/pages/doc.js
adrianaluchian: `disabledActions` is because that is the exception from the rule, and the rule is that all actions are enabled.

adrianaluchian: I will check your second remark/comment here, though unless I'm missing something I don't get why it would behave like you are mentioning 🤔

laurentiu-ilici: Try the same on staging. You don't see the "delete section" when you just view the document. If you look at the line above that is the difference
  const isHardDeleteEnabled = state.type === DOCUMENT_TYPE.revision
I think this got lost

adrianaluchian: Ok I checked and also better understood what you meant. The fact that you can **hard delete** a section from a revision is, from what I know, actually what is wanted. The reason for this functionality is to be able to delete content that should legally not be on the website, in whichever revision it is (not only the last). That's also why the dialog when you delete asks you if you want to delete the same section in all other revisions (backwards and forwards basically)

src/components/pages/docs.yml
adrianaluchian: True, reused translation for deletion, thank you

src/services/export-service.js
adrianaluchian: Yes, changed

src/services/import-service.js
adrianaluchian: Good idea, done and also now moved setting `importType = DOCUMENT_IMPORT_TYPE.add` on the `else` branch cause it was bothering me that I had to pre-set the variable to it.

src/components/pages/doc.js
laurentiu-ilici: I don't think you understood what I meant :), the bug is still there.

adrianaluchian: "Try the same on staging. You don't see the "delete section" when you just view the document"
I think this is something you introduced on staging with the previous ticket but that is a bug, it basically broke the intended functionality.
@ahelmberger , what's your take on this?

laurentiu-ilici: This wasn't a bug, it was because previously when you came to the doc.js you didn't see the "article" (aka the document) but rather all revisions (like you have when you now click the "view revisions"). Think of the doc.js page as the article page when you come to it the first time :) and try that on prod.

adrianaluchian: If you go on prod now, on a document page as is (view mode, not edit mode) and hover on any section (in any revision), you see the option to hard delete the section.

laurentiu-ilici: > `disabledActions` is because that is the exception from the rule, and the rule is that all actions are enabled.

To me it seems weird that the logic for what is enables / disables the actions now lives in 2 places:
 doc.js has:   const disabledSectionActions = [...isExternalDocument ? [SECTION_ACTIONS.hardDelete] : []];
 section-display:  in between that just passes the disabledActions 
 section-action-dropdown: again knows about all actions and adds some additional restraints on this particular action:
if (canHardDelete && !section.deletedOn && //the rest)

Would it not be more streamlined if the 
   section-action-dropdown just knew how to display in stuff as a dumb component that receives it's actions from someone who knows? 
Alternatively, it could keep all the conditions. It's the split that is a bit hard to maintain I think.

laurentiu-ilici: > If you go on prod now, on a document page as is (view mode, not edit mode) and hover on any section (in any revision), you see the option to hard delete the section.

And if you go to the articles page (which was merged in the doc page) you see you can't delete it :)

ahelmberger: In the former article page there was no option to delete a section, only in the docs view (with the revision slider), and @laurentiu-ilici kept this behavior also for the now combined pages. Now with this change also the article view mode allows deletions. I don't really have an opinion on what would be better.

ahelmberger: But it couldn't hurt to also be able to delete a section without going into revisions mode before, right?

src/services/import-service.js
```diff
@@ -59,8 +59,14 @@ class ImportService {
       .map(exportableDocument => {
         const importedDocument = importedDocuments.find(document => document.key === exportableDocument.key);
 
-        if (importedDocument?.revision === exportableDocument.revision) {
-          return null;
+        let importType;
+
+        if (importedDocument) {
```
laurentiu-ilici, line 64: Nice!

src/components/pages/doc.js
laurentiu-ilici: > If you go on prod now, on a document page as is (view mode, not edit mode) and hover on any section (in any revision), you see the option to hard delete the section.

Anyway, as I said we can keep this new way of doing things, but then the code needs to change a bit. If you now delete all revisions, you get a blank screen after saying "Ja". Then you need to refresh the page to get to a stable state.

adrianaluchian: Regarding hard deletion in various view modes, I will make the smallest change for now and allow hard-delete only for document view mode.

src/components/pages/doc.js
```diff
@@ -258,7 +258,8 @@ function Doc({ initialState, PageTemplate }) {
         { state.revisions.length > 0 && revisionPicker}
         <DocView
           documentOrRevision={state.currentDocOrRevision}
-          onAction={isHardDeleteEnabled ? handleAction : null}
+          disabledActions={disabledSectionActions}
```
adrianaluchian, line 261: Separating discussion in second thread.

**laurentiu-ilici**
This construct seems backwards to me. I had to dig down 3 files from here to understand why you send "disabledActions" instead of "possible/enabledActions".

**adrianaluchian**
`disabledActions` is because that is the exception from the rule, and the rule is that all actions are enabled.

**laurentiu-ilici**
To me it seems weird that the logic for what is enables / disables the actions now lives in 2 places:
doc.js has: const disabledSectionActions = [...isExternalDocument ? [SECTION_ACTIONS.hardDelete] : []];
section-display: in between that just passes the disabledActions
section-action-dropdown: again knows about all actions and adds some additional restraints on this particular action:
if (canHardDelete && !section.deletedOn && //the rest)

Would it not be more streamlined if the
section-action-dropdown just knew how to display in stuff as a dumb component that receives it's actions from someone who knows?
Alternatively, it could keep all the conditions. It's the split that is a bit hard to maintain I think.

adrianaluchian: What I meant by "the rule is that all actions are enabled" was that the components involved are reused in some cases, and I did not want to escalate into a big change with risk of side-bugs.

`section-action-dropdown` is already not a dumb component, given it decides `canHardDelete`. I agree that the conditions to show the delete menu come from 2 different levels. To me they should come from above, but I would not want to escalate into a refactoring.
So I will then change to pass `isExternalDocument` from the page down through the components instead of `disabledActions` and have the lowest component decide `canHardDelete` based on this, to keep the pattern.

We can definitely rethink these components if we want different modes from them and we will probably have to for introducing inline editing of documents.

adrianaluchian: Actually scratch that, since the condition to disable editing will no longer be `isExternalDocument` but `state.type === DOCUMENT_TYPE.revision && isExternalDocument` 😂 I wouldn't pass the document type to the lower components in this case... I'd rather keep `disabledActions`

src/components/pages/doc.js
laurentiu-ilici: You mean for revision view mode :)). I see this is very confusing. Okay, we will improve in a different ticket, the important thing is that it still works.

src/components/pages/doc.js
laurentiu-ilici: 😆

laurentiu-ilici (approved):

ahelmberger (approved):

src/components/pages/doc.js
adrianaluchian: Got one better:
I gave up fighting to keep the `onAction` generic and went back to not attaching an event handler to it at all when section deletion is disabled. Otherwise there were more changes to make so that the UI doesn't show a border on hovering the sections when `doc` page was not in `review` mode.

We'll redo this part with `doc` inline edit.

Please check the changes once more that I haven't missed something 🙏

laurentiu-ilici: Ah, I checked before and I saw that the border was there but decided to ignore it.
I'll check.

adrianaluchian: But Andreas pointed it out to me earlier and from then on I could no longer unsee it :D

laurentiu-ilici (approved):
Rechecked the section hard delete and reapproved. All good!

PR #90, opened 2021-12-14
EDU-176 error on safari

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-176

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -211,7 +211,7 @@ export async function buildTestAppJs() {
   } else {
     testAppBuildResult = await esbuild.build({
       entryPoints: ['./test-app/bundles/main.js'],
-      target: ['esnext', 'chrome95', 'firefox93', 'safari15', 'edge95'],
+      target: ['esnext', 'chrome95', 'firefox93', 'safari13', 'edge95'],
       format: 'esm',
       bundle: true,
       splitting: true,
--- a/src/components/pages/browser-not-supported.yml
+++ b/src/components/pages/browser-not-supported.yml
@@ -1,3 +1,6 @@
+headerText:
+ en: We have detected that your browser version is not supported by our website
+ de: Wir haben festgestellt, dass Ihre Browserversion von unserer Website nicht unterstützt wird
 mainText:
- en: We have detected that your browser version is not supported by our
- de: We have detected that your browser version is not supported by our
+ en: Please download a modern broswer of your choice, some examples below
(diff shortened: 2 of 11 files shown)


gulpfile.js
```diff
@@ -211,7 +211,7 @@ export async function buildTestAppJs() {
   } else {
     testAppBuildResult = await esbuild.build({
       entryPoints: ['./test-app/bundles/main.js'],
-      target: ['esnext', 'chrome95', 'firefox93', 'safari15', 'edge95'],
+      target: ['esnext', 'chrome95', 'firefox93', 'safari13', 'edge95'],
```
laurentiu-ilici, line 214: I'm not sure what to do with the CSS bundler

src/components/pages/browser-not-supported.yml
```diff
@@ -0,0 +1,7 @@
+headerText:
+ en: We have detected that your browser version is not supported by our website
+ de: Wir haben festgestellt, dass Ihre Browserversion von unserer Website nicht unterstützt wird
+mainText:
+ en: Please download a modern broswer of your choice, some examples below
+ de: Bitte laden Sie einen modernen Browser Ihrer Wahl herunter, einige Beispiele unten
+
```
laurentiu-ilici, line 7: This is weird. On local I have only one line break at the end which I uploaded when it was complaining about a missing line break. Now it looks like this, I don't know why....

gulpfile.js
ahelmberger: It would be cool if you could also make sure that Autoprefixer is configured to respect Safari >=13. Otherwise we need to make a second ticket for that.

src/components/pages/browser-not-supported.yml
ahelmberger: The last line is not empty (it contains a single space), but also missing the final newline. If you install the editorconfig plugin, it will make sure this doesn't happen.

laurentiu-ilici: Geinstalled, danke! It seems to work

gulpfile.js
laurentiu-ilici: I hope I did it right, I knew there was something with the autoprefixer but I'm not sure how to test it.

ahelmberger: Yes, it's right. You can always check at: https://browserslist.dev/

ahelmberger (approved):

PR #91, opened 2021-12-14
Feature/edu 184 disable archiving external docs

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-184

### DIFF
--- a/src/common/github-flavored-markdown.js
+++ b/src/common/github-flavored-markdown.js
@@ -35,7 +35,7 @@ class GithubFlavoredMarkdown {
       }
     });
     extractor.use(replaceLink);
-    extractor.parse(markdown);
+    extractor.render(markdown);
     return [...linkSet].map(link => link.substr(6));
   }
 }
--- a/src/services/document-service.js
+++ b/src/services/document-service.js
@@ -291,17 +291,22 @@ class DocumentService {
     }
 
     const latestRevision = await this.getCurrentDocumentRevisionByKey(documentKey);
-    const order = await this.documentOrderStore.getNextOrder();
 
-    const newRevision = this._buildDocumentRevision({ data: latestRevision, documentKey, userId: user._id, order, sections: latestRevision.sections });
-
-    newRevision.appendTo = {
-      key: documentKey,
-      ancestorId: latestRevision._id
+    const doc = {
+      title: latestRevision.title,
+      slug: latestRevision.slug,
+      namespace: latestRevision.namespace,
+      language: latestRevision.language,
+      sections: latestRevision.sections,
+      appendTo: {
+        key: documentKey,
+        ancestorId: latestRevision._id
+      },
+      tags: latestRevision.tags,
+      archived
     };
-    newRevision.archived = archived;
 
-    return this.createNewDocumentRevision({ doc: newRevision, user });
+    return this.createNewDocumentRevision({ doc, user });
   }
 
   async createNewDocumentRevision({ doc, user, restoredFrom = null }) {
--- a/src/ui/default-prop-types.js
+++ b/src/ui/default-prop-types.js
@@ -28,10 +28,11 @@ export const importSourceShape = PropTypes.shape(importSourceProps);
 
 export const clientConfigProps = {
   clientConfig: PropTypes.shape({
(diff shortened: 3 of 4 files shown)


ahelmberger:
I also fixed a few thing in this PR that are not actually part of the ticket, I will put comments here

src/common/github-flavored-markdown.js
```diff
@@ -35,7 +35,7 @@ class GithubFlavoredMarkdown {
       }
     });
     extractor.use(replaceLink);
-    extractor.parse(markdown);
+    extractor.render(markdown);
```
ahelmberger, line 38: `parse` threw some errors when importing documents, probably because of invalid Markdown syntax

src/services/document-service.js
```diff
@@ -291,17 +291,22 @@ class DocumentService {
...
+      slug: latestRevision.slug,
+      namespace: latestRevision.namespace,
+      language: latestRevision.language,
+      sections: latestRevision.sections,
+      appendTo: {
+        key: documentKey,
+        ancestorId: latestRevision._id
+      },
+      tags: latestRevision.tags,
+      archived
```
ahelmberger, line 306: `getNextOrder` is called by `createNewDocumentRevision` (as is `_buildDocumentRevision`). Also the check if it is an internal doc is done in `createNewDocumentRevision`, so all that remains to be done here is to set the `archived` prop on the last revision and save it again.

src/ui/default-prop-types.js
```diff
@@ -28,10 +28,11 @@ export const importSourceShape = PropTypes.shape(importSourceProps);
...
   clientConfig: PropTypes.shape({
-    cdnRootUrl: PropTypes.string.isRequired
-  }).isRequired,
-  importSources: PropTypes.arrayOf(importSourceShape),
-  cookieConsentName: PropTypes.string.isRequired
+    cdnRootUrl: PropTypes.string.isRequired,
+    disabledFeatures: PropTypes.arrayOf(PropTypes.string).isRequired,
+    importSources: PropTypes.arrayOf(importSourceShape).isRequired,
+    consentCookieName: PropTypes.string.isRequired
+  }).isRequired
```
ahelmberger, line 35: This data structure was just wrong and got me a red console in the browser dev tools. It should be correct now.

src/services/document-service.js
```diff
@@ -291,17 +291,22 @@ class DocumentService {
...
 
     const latestRevision = await this.getCurrentDocumentRevisionByKey(documentKey);
-    const order = await this.documentOrderStore.getNextOrder();
 
-    const newRevision = this._buildDocumentRevision({ data: latestRevision, documentKey, userId: user._id, order, sections: latestRevision.sections });
-
-    newRevision.appendTo = {
-      key: documentKey,
-      ancestorId: latestRevision._id
+    const doc = {
```
laurentiu-ilici, line 295: Much better :D

laurentiu-ilici (approved):

src/services/document-service.js
adrianaluchian: The reason for `_buildDocumentRevision` was that it unifies revision creation. What I dislike in composing it in 2 places is that it leaves room for error and inconsistency.

src/ui/default-prop-types.js
adrianaluchian: Indeed it was wrong 😶‍🌫️

src/services/document-service.js
ahelmberger: It's not really composing in 2 places, I would say. `_buildDocumentRevision` builds a data structure that is not really the one expected by `createNewDocumentRevision` (it puts much more properties onto the "doc"/"data" object). Even though that doesn't hurt in theory, I disliked calling `getNextOrder` twice. I don't know how to solve this, except maybe validating the params of `createNewDocumentRevision` with `joi`, so we can be sure the data is 100% correct.

adrianaluchian: Ok then maybe let's do that then, the joi validation in a separate ticket 🤔

adrianaluchian (approved):

src/services/document-service.js
ahelmberger: Great idea!

PR #93, opened 2021-12-15
EDU-191 Fix image width in markdown

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-191

### DIFF
--- a/src/plugins/annotation/markdown.less
+++ b/src/plugins/annotation/markdown.less
@@ -1,1 +1,0 @@
-.Annotation {}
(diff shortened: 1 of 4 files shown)


ahelmberger:
We've learned today that we also have markdown in the annotation plugin, so this PR currently only fixes half of the cases probably.

laurentiu-ilici:
> We've learned today that we also have markdown in the annotation plugin, so this PR currently only fixes half of the cases probably.

Gefixed, thanks, I already forgot

src/plugins/annotation/markdown.less
```diff
@@ -1 +0,0 @@
-.Annotation {}
```
ahelmberger, line 1: Interesting, file had the wrong name and was never used. Ts, ts, ts

ahelmberger (approved):

PR #94, opened 2021-12-15
EDU-186 Extract common dev tasks into reusable modules

ahelmberger:
This PR does a little bit more than in the ticket (sorry, again). The idea is to extract the tasks that are the same in elmu, oma and educandu (and other packages), which is probably 95% reusable (and our gulpfiles a little bit lighter). For now I extracted some stuff to a folder called `dev`, we can continue to do this and then create an npm package out of this directory.

Closes https://educandu.atlassian.net/browse/EDU-186

### DIFF
--- a/dev/process.js
+++ b/dev/process.js
@@ -1,0 +1,62 @@
+import { delay } from './helpers.js';
+import { spawn } from 'child_process';
+
+export const DEFAULT_ENV = { ...process.env };
+export const DEFAULT_STDIO = 'inherit';
+export const DEFAULT_STARTUP_GRACE_PERIOD = 250;
+
+export class Process {
+  constructor({
+    command, args = [],
+    env = DEFAULT_ENV,
+    stdio = DEFAULT_STDIO,
+    startupGracePeriod = DEFAULT_STARTUP_GRACE_PERIOD
+  }) {
+    this.command = command;
+    this.args = args;
+    this.env = env;
+    this.stdio = stdio;
+    this.startupGracePeriod = startupGracePeriod;
+    this._process = null;
+  }
+
+  async start(env = {}) {
+    if (!this._process) {
+      const options = { env: { ...this.env, ...env }, stdio: this.stdio };
+      this._process = spawn(this.command, this.args, options);
+      this._process.once('exit', () => {
+        this._process = null;
+      });
+    }
+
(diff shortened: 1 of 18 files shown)


laurentiu-ilici (approved):

package.json
```diff
@@ -109,7 +109,9 @@
   },
   "devDependencies": {
     "@jest/globals": "^27.3.1",
+    "@rauschma/stringio": "^1.4.0",
```
adrianaluchian, line 112: hehe, Mr 2ality 😎

gulpfile.js
```diff
@@ -523,17 +397,18 @@ export async function release() {
...
+        { responseType: 'json', auth: { username: cliArgs.jiraUser, password: cliArgs.jiraApiKey } }
       );
     } catch (error) {
       console.log(error);
     }
   }
 }
 
-export const up = gulp.series(mongoUp, minioUp, maildevUp);
+export const up = gulp.parallel(mongoUp, minioUp, maildevUp);
```
adrianaluchian, line 408: Nice!

dev/process.js
```diff
@@ -0,0 +1,62 @@
...
+
+  waitForExit() {
+    return new Promise(resolve => {
+      if (this._process) {
+        this._process.once('exit', async () => {
+          await delay(100);
+          resolve();
+        });
+      } else {
+        delay(0).then(resolve);
```
adrianaluchian, line 58: I'm missing a bit why this needs to be resolved on the next tick 🤔

gulpfile.js
ahelmberger: Yeah, no clue why it was `series` before :grin:

package.json
ahelmberger: Finally he's officially part of the project

adrianaluchian:
Super cool console when you run multiple, with the color coding distinction 👍 
I tested some imports with multiple instances and all worked fine on my end!

dev/process.js
ahelmberger: Actually no reason except that I don't like resolving `new Promise`s synchronously in the constructor (so it is resolved even before the promise is assigned to a variable). With `delay(0)` we're always waiting now (just shorter in one case). It's a little bit neurotic, I admit.

adrianaluchian (approved):

PR #96, opened 2021-12-16
EDU-193 add delete button to cdn browser

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-193


src/components/repository-browser.less
```diff
@@ -59,6 +59,10 @@
   }
 }
 
+.RepositoryBrowser-tableDeleteCell {
+  color: red;
```
adrianaluchian, line 63: We could/should use the defined color ranges we have (`@red-1`, `@red-2` etc)

src/server/cdn-controller.js
```diff
@@ -29,6 +30,18 @@ class CdnController {
       return res.send({ objects });
     });
 
+    router.delete('/api/v1/cdn/objects/:objectName', [needsPermission(permissions.DELETE_CDN_FILE), validateQuery(deleteObjectQuerySchema)], async (req, res) => {
+      const objectName = req.params.objectName;
+      const prefix = req.query.prefix;
+      if (!objectName) {
+        throw new BadRequest();
+      }
```
adrianaluchian, line 38: There could also be a `validateParams` for this, like for the query.

src/services/cdn-api-client.js
```diff
@@ -35,6 +35,15 @@ class CdnApiClient {
 
     return request.then(res => res.data);
   }
+
+  deleteFile(prefix, fileName) {
```
adrianaluchian, line 39: To reflect the URL, I think this would be better called "deleteCdnObject" or just "deleteObject".

src/components/repository-browser.js
```diff
@@ -264,16 +269,30 @@ class RepositoryBrowser extends React.Component {
...
   }
 
+  async handleDeleteFile(fileName) {
+    const { cdnApiClient, onSelectionChanged } = this.props;
+    const { currentPathSegments, selectedRowKeys } = this.state;
+    const prefix = pathHelper.getPrefix(currentPathSegments);
+    const objectName = `${prefix}${fileName}`;
+
+    await cdnApiClient.deleteFile(prefix, fileName);
+    if (selectedRowKeys.includes(objectName)) {
```
adrianaluchian, line 289: This is actually `if (selectedRowKeys === fileName)`, or?

src/components/confirmation-dialogs.yml
```diff
@@ -7,6 +7,9 @@ deleteSectionConfirmation:
 deleteDocumentConfirmation:
   en: Do you really want to delete document "{title}"?
   de: Möchten Sie das Dokument "{title}" wirklich löschen?
+deleteCdnDocumentConfirmation:
+  en: Do you really want to delete the cdn file "{fileName}"?
+  de: Möchten Sie das cdn Datei "{fileName}" wirklich löschen?
```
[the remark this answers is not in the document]
ahelmberger: Möchten Sie die Datei "{fileName}" wirklich vom CDN löschen?

src/components/repository-browser.js
```diff
@@ -106,6 +95,22 @@ class RepositoryBrowser extends React.Component {
         sorter: firstBy('lastModified')
       }
     ];
+
+    if (hasUserPermission(this.props.user, permissions.DELETE_CDN_FILE)) {
+      this.columns.unshift({
```
adrianaluchian, line 100: To me it would be nicer to have "delete" as the last column, unless I am the only one that things so :)

src/components/confirmation-dialogs.yml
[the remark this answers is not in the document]
ahelmberger: I would also use capitalized letters CDN in the english text

src/components/repository-browser.js
```diff
@@ -106,6 +95,22 @@ class RepositoryBrowser extends React.Component {
...
+        dataIndex: 'displayName',
+        key: 'displayName',
+        render: this.renderDeleteColumn,
```
adrianaluchian, line 103: Could we maybe only render the column if `!isDirectory`?

When I tested deleting a directory it did not work, but I guess we don't want "directory" deletion, or would be safer not to allow it?

laurentiu-ilici: We don't want that, I missed that, I'll try to fix it, thanks!

src/components/confirmation-dialogs.yml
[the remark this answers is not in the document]
adrianaluchian: It's irrespective of language, it's an acronym and those are always capitalized as far as I know 🤔 Or at least CDN should be so 😄

src/components/repository-browser.js
laurentiu-ilici: It will be hidden way at the end after those horribly long columns, like this you have the name of the file and the button in the same place.

src/components/repository-browser.js
laurentiu-ilici: That's what I thought as well, but that wasn't the case.

src/services/cdn-api-client.js
laurentiu-ilici: Renamed to deleteCdnObject

src/components/repository-browser.less
laurentiu-ilici: Fixed, thanks. I guess we should make a ticket, we have lots of "red" in the code :))

src/components/confirmation-dialogs.yml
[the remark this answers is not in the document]
laurentiu-ilici: Fixed both comments, thanks!

src/server/cdn-controller.js
laurentiu-ilici: Fixed, thanks!

src/components/repository-browser.less
adrianaluchian: Yeah we need to add an `@error-color: @red-x` that could then also be changed in the system that uses educandu

ahelmberger: From ant-design we have:

~~~
@primary-color          : @global-primary-color;
@info-color             : @blue-6;
@success-color          : @green-6;
@processing-color       : @blue-6;
@error-color            : @red-6;
@highlight-color        : @red-6;
@warning-color          : @gold-6;
@normal-color           : #d9d9d9;
~~~

but I think it's fine for now (we have to go over the "theming" variables once again when we polish OMA)

src/components/repository-browser.js
ahelmberger: On second thought, I like the idea of having it "less reachable" as it is a last resort admin action. So I'd also vote for last column (not very insisting though)

laurentiu-ilici: No problem, done.

src/components/repository-browser.js
laurentiu-ilici: This is fixed in the renderer, it's not really possible here because you don't have yet the knowledge if the row is a directory or not.

adrianaluchian: Oh of course in the renderer, I didn't imply to fix it **exactly here** 😊  I mentioned "only render the column if !isDirectory", since this is the place that sets up the column.

adrianaluchian (approved):

src/components/repository-browser.js
```diff
@@ -6,34 +7,22 @@ import classNames from 'classnames';
...
-  [mimeTypeHelper.CATEGORY_DOCUMENT]: 'mimeTypeCategoryDocument',
-  [mimeTypeHelper.CATEGORY_SPREADSHEET]: 'mimeTypeCategorySpreadsheet',
-  [mimeTypeHelper.CATEGORY_PRESENTATION]: 'mimeTypeCategoryPresentation',
-  [mimeTypeHelper.CATEGORY_PROGRAM]: 'mimeTypeCategoryProgram',
-  [mimeTypeHelper.CATEGORY_FOLDER]: 'mimeTypeCategoryFolder',
-  [mimeTypeHelper.CATEGORY_UNKNOWN]: 'mimeTypeCategoryUnknown'
-};
-
-const localizeCategory = (cat, t) => t(resourceKeyMap[cat]);
+const localizeCategory = (cat, t) => t(mimeTypeHelper.resourceKeyMap[cat]);
```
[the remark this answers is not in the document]
ahelmberger: We moved the translation keys into `mimeTypeHelper` (which is nice) but the translations are still in `repository-browser.yml`. Let's merge this but then also move those, because now they're separated.

ahelmberger (approved):

PR #98, opened 2021-12-16
EDU-159 Remove namespace field

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-159


migrations/automatic/educandu-2021-12-16-01-delete-namespaces.js
```diff
@@ -0,0 +1,23 @@
...
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    await this.db.collection('documents').dropIndex('_idx_namespace_slug_');
+    await this.db.collection('documents').createIndexes([
+      {
+        name: '_idx_slug_',
+        key: { slug: 1 }
```
laurentiu-ilici, line 12: This should actually only live in the elmu db...

src/services/document-service.spec.js
```diff
@@ -291,7 +288,8 @@ describe('document-service', () => {
       });
 
       it('creates the revisions', async () => {
-        const createdRevisions = await db.documentRevisions.find({ key: documentKey }).toArray();
+        const createdRevisions = await db.documentRevisions.find({ key: documentKey }).sort({ order: 1 }).toArray();
```
laurentiu-ilici, line 291: This test was unstable, it was getting the wrong ordering on the revisions and failing.

migrations/automatic/educandu-2021-12-16-01-delete-namespaces.js
ahelmberger: True, we have to discuss how we wanne do such things

src/services/document-service.spec.js
adrianaluchian: It actually had the sort there exactly cause it was unstable otherwise (plus it should have been there anyway, like we have it in the service when it returns documents). Don't get when it vanished..

ahelmberger: :+1:

adrianaluchian (approved):

ahelmberger (approved):

PR #99, opened 2021-12-16
EDU-188 Log skipped migrations

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-188

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -333,6 +333,15 @@ export async function migrate() {
     logger: console
   });
 
+  const executedMigrationNames = (await umzug.executed()).map(migration => migration.name);
+
+  migrationsToRun.forEach(migrationFullPath => {
+    const executedMigrationName = executedMigrationNames.find(migrationName => migrationFullPath.includes(migrationName));
+    if (executedMigrationName) {
+      console.log(`Migration ${executedMigrationName} was already run, skipping it.`);
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -333,6 +333,15 @@ export async function migrate() {
...
 
+  const executedMigrationNames = (await umzug.executed()).map(migration => migration.name);
+
+  migrationsToRun.forEach(migrationFullPath => {
+    const executedMigrationName = executedMigrationNames.find(migrationName => migrationFullPath.includes(migrationName));
+    if (executedMigrationName) {
+      console.log(`Migration ${executedMigrationName} was already run, skipping it.`);
+    }
+  });
+
```
ahelmberger, line 344: BTW. there's also `umzug.up({ rerun: true })` in case you want to give the user the chance to rerun. There's also a function on `umzug` that would tell you which migrations are already run (i.e. in the DB table). That means, we could change the logic to show the user in the list which migrations have already run (and would be rerun if selected) and which are still missing. Then I don't know if it's worth it, just saying. I'll approve anyway.

ahelmberger (approved):

gulpfile.js
adrianaluchian: That's cool, that you could rerun a migration. Actually I was also tempted to extract the migration logic in the `dev` folder, but then thought better to first see which direction we actually go with migrations.
Same with your suggestions, I guess, before spending too much time with this we should see how we deal with migrations next.

PR #100, opened 2021-12-17
EDU-192 Escape non-url-characters in file upload

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-192

### DIFF
--- a/src/utils/file-name-helper.js
+++ b/src/utils/file-name-helper.js
@@ -1,23 +1,16 @@
 import path from 'path';
+import urls from './urls.js';
 import uniqueId from './unique-id.js';
+import slugify from '@sindresorhus/slugify';
 
-function insertBeforeExtension(fileName, stringToInsert) {
-  const index = fileName.indexOf('.');
-  if (index === -1) {
-    return fileName + stringToInsert;
-  }
-
-  return fileName.substr(0, index) + stringToInsert + fileName.substr(index);
-}
-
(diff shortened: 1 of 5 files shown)


adrianaluchian (approved):

src/utils/file-name-helper.js
```diff
@@ -1,23 +1,18 @@
...
+export function buildCdnFileName(fileName, prefix = null) {
   const id = uniqueId.create();
-  const baseName = path.basename(fileName);
-  const fileNameWithoutBaseName = fileName.substr(0, fileName.length - baseName.length);
-  const uniqueBaseName = insertBeforeExtension(baseName, `_${id}`);
-  return fileNameWithoutBaseName + uniqueBaseName;
+  const extension = path.extname(fileName);
+  const baseName = fileName.substr(0, fileName.length - extension.length);
+  const slugifiedBaseName = slugify(baseName);
+  const uniqueBaseName = [slugifiedBaseName, id].filter(x => x).join('-');
```
laurentiu-ilici, line 11: Why do we need the filtering? The id should not be never be '' and the filename I don't believe can be '', or can it?

adrianaluchian: It's a nice scenario covered by tests, if you remove the filter you'll see it 😬

laurentiu-ilici (approved):

PR #102, opened 2021-12-17
EDU-201 Fix socket hangup

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-201

### DIFF
--- a/dev/helpers.js
+++ b/dev/helpers.js
@@ -11,7 +11,7 @@ export function delay(ms) {
 export const glob = promisify(globModule);
 
 export function isMac() {
-  return os.platform === 'darwin';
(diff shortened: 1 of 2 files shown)


dev/helpers.js
```diff
@@ -11,7 +11,7 @@ export function delay(ms) {
 export const glob = promisify(globModule);
 
 export function isMac() {
-  return os.platform === 'darwin';
+  return os.platform() === 'darwin';
```
adrianaluchian, line 14: This is the actual problem we got with refactoring.

The retries are just an extra logic introduce while chasing the bug, kept it as a good to have.

laurentiu-ilici (approved):

PR #104, opened 2021-12-20
EDU-167 create room data model

laurentiu-ilici:


migrations/automatic/educandu-2021-12-20-01-create-rooms-collection.js
```diff
@@ -0,0 +1,40 @@
+// eslint-disable-next-line camelcase
+export default class Educandu_2021_12_20_01_create_rooms_colletion {
```
ahelmberger, line 2: Typo collection

migrations/automatic/educandu-2021-12-20-01-create-rooms-collection.js
```diff
@@ -0,0 +1,40 @@
...
+    await this.db.createCollection('rooms');
+    await this.db.createCollection('roomInvitations');
+
+    await this.db.collection('rooms').createIndexes([
+      {
+        name: '_idx_owner_',
+        key: { owner: 1 }
+      },
+      {
+        name: '_idx_members_user_id',
```
ahelmberger, line 17: For consistency we need another underscore at the end

migrations/automatic/educandu-2021-12-20-01-create-rooms-collection.js
laurentiu-ilici: Fixed, thanks

migrations/automatic/educandu-2021-12-20-01-create-rooms-collection.js
laurentiu-ilici: Fixed in both files. Thanks!

ahelmberger:
There is room-service, but room**s**-controller, room**s**-store and room-invitation**s**-store. I think that all services, stores and controllers so far use the singular, while the collection specs (and the collection names) use the plural, maybe we should keep that consistent.

laurentiu-ilici:
> There is room-service, but room**s**-controller, room**s**-store and room-invitation**s**-store. I think that all services, stores and controllers so far use the singular, while the collection specs (and the collection names) use the plural, maybe we should keep that consistent.

Fixed

ahelmberger (approved):

PR #105, opened 2021-12-20
EDU-162 Change account page to my space

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-162


src/components/default-page-template.js
```diff
@@ -65,59 +68,74 @@ function DefaultPageTemplate({ children, fullScreen, headerActions, alerts }) {
...
-  }
-
-  pageMenuItems.push({
-    key: 'language',
-    onClick: () => setIsUiLanguageDialogVisible(true),
-    text: t('common:language'),
-    icon: GlobalOutlined,
-    permission: null
-  });
+      permission: permissions.EDIT_SETTINGS,
```
laurentiu-ilici, line 119: This is unrelated to the PR itself: Why do we only show a help page to users that have this permission? This means only admins will ever see it, shouldn't we have help pages for everyone?

ahelmberger:
While testing this I realized that in the educandu test app we do not show the insufficient profile warning (I wanted to test the link to what was the profile page before). It's not really related to this PR, but I think we should enable it. Something that might be a little bit of a downside is, that now when the user clicks on this link they get transferred to a page where the profile is shown in the second tab (and not immediately visible).

src/components/default-page-template.js
```diff
@@ -65,59 +68,74 @@ function DefaultPageTemplate({ children, fullScreen, headerActions, alerts }) {
       href: urls.getHomeUrl(),
       text: t('pageNames:home'),
       icon: HomeOutlined,
-      permission: null
+      permission: null,
+      showWhen: true
```
ahelmberger, line 72: Good idea!

adrianaluchian:
> While testing this I realized that in the educandu test app we do not show the insufficient profile warning (I wanted to test the link to what was the profile page before). It's not really related to this PR, but I think we should enable it. Something that might be a little bit of a downside is, that now when the user clicks on this link they get transferred to a page where the profile is shown in the second tab (and not immediately visible).

We could also make the profile tab as the first tab, as the simplest solution. 
Come to think of it, it probably does make sense to see this first.

Will check the insufficient profile warning case, see why it's not shown.

adrianaluchian:
> > While testing this I realized that in the educandu test app we do not show the insufficient profile warning (I wanted to test the link to what was the profile page before). It's not really related to this PR, but I think we should enable it. Something that might be a little bit of a downside is, that now when the user clicks on this link they get transferred to a page where the profile is shown in the second tab (and not immediately visible).
> 
> We could also make the profile tab as the first tab, as the simplest solution. Come to think of it, it probably does make sense to see this first.
> 
> Will check the insufficient profile warning case, see why it's not shown.

We do show the insufficient profile warning, to test please remove both first and last name from the profile and go to `docs` page for example. It is not shown on the `my-space` page.

src/components/profile-tab.yml
```diff
@@ -1,49 +1,34 @@
...
-profile.country:
+country:
   en: Country
   de: Land
-profile.updateSuccessMessage:
+updateSuccessMessage:
   en: Your profile was updated successfully.
   de: Ihr Profil wurde erfolgreich aktualisiert.
-profile.avatarDescription:
+avatarDescription:
```
laurentiu-ilici, line 31: This still says elmu, maybe we should make some ticket to make it overridable.

ahelmberger:
> > > While testing this I realized that in the educandu test app we do not show the insufficient profile warning (I wanted to test the link to what was the profile page before). It's not really related to this PR, but I think we should enable it. Something that might be a little bit of a downside is, that now when the user clicks on this link they get transferred to a page where the profile is shown in the second tab (and not immediately visible).
> > 
> > 
> > We could also make the profile tab as the first tab, as the simplest solution. Come to think of it, it probably does make sense to see this first.
> > Will check the insufficient profile warning case, see why it's not shown.
> 
> We do show the insufficient profile warning, to test please remove both first and last name from the profile and go to `docs` page for example. It is not shown on the `my-space` page.

True, thanks, I was also expecting it on the "fullScreen" pages like in the very old days :-) So yeah, I think the idea to make the profile the first tab would make a lot of sense.

adrianaluchian:
> > > > While testing this I realized that in the educandu test app we do not show the insufficient profile warning (I wanted to test the link to what was the profile page before). It's not really related to this PR, but I think we should enable it. Something that might be a little bit of a downside is, that now when the user clicks on this link they get transferred to a page where the profile is shown in the second tab (and not immediately visible).
> > > 
> > > 
> > > We could also make the profile tab as the first tab, as the simplest solution. Come to think of it, it probably does make sense to see this first.
> > > Will check the insufficient profile warning case, see why it's not shown.
> > 
> > 
> > We do show the insufficient profile warning, to test please remove both first and last name from the profile and go to `docs` page for example. It is not shown on the `my-space` page.
> 
> True, thanks, I was also expecting it on the "fullScreen" pages like in the very old days :-) So yeah, I think the idea to make the profile the first tab would make a lot of sense.

Done!

src/components/account-tab.js
```diff
@@ -0,0 +1,95 @@
...
+    forbiddenUsernames: []
+  });
+
+  const formRef = React.createRef();
+
+  const saveAccountData = async ({ username, email }) => {
+    try {
+      const { result, user: updatedUser } = await userApiClient.saveUserAccount({ username, email });
+
+      switch (result) {
```
laurentiu-ilici, line 37: Maybe we should consider for such cases using the useReducer hook. It would make this testable and a lot easier to read. @adrianaluchian @ahelmberger what do you think?

ahelmberger:
> > > > > While testing this I realized that in the educandu test app we do not show the insufficient profile warning (I wanted to test the link to what was the profile page before). It's not really related to this PR, but I think we should enable it. Something that might be a little bit of a downside is, that now when the user clicks on this link they get transferred to a page where the profile is shown in the second tab (and not immediately visible).
> > > > 
> > > > 
> > > > We could also make the profile tab as the first tab, as the simplest solution. Come to think of it, it probably does make sense to see this first.
> > > > Will check the insufficient profile warning case, see why it's not shown.
> > > 
> > > 
> > > We do show the insufficient profile warning, to test please remove both first and last name from the profile and go to `docs` page for example. It is not shown on the `my-space` page.
> > 
> > 
> > True, thanks, I was also expecting it on the "fullScreen" pages like in the very old days :-) So yeah, I think the idea to make the profile the first tab would make a lot of sense.
> 
> Done!

:tada:

ahelmberger (approved):

src/components/profile-tab.js
```diff
@@ -0,0 +1,146 @@
...
+function ProfileTab({ formItemLayout, tailFormItemLayout }) {
+  const user = useUser();
+  const { language } = useLanguage();
+  const { t } = useTranslation('profileTab');
+  const userApiClient = useService(UserApiClient);
+  const countryNameProvider = useService(CountryNameProvider);
+
+  const profile = user.profile || { country: '' };
+  const gravatarUrl = gravatar.url(user.email, { s: AVATAR_SIZE, d: 'mp' });
+  const gravatarRagistrationUrl = `https://${language}.gravatar.com/`;
```
laurentiu-ilici, line 40: Maybe we could move this to the urls package? urls.getGravatarRegistrationUrl(language)?

adrianaluchian: To me that's for composing internal/website URLs, I would not put this one there.

src/components/default-page-template.js
adrianaluchian: Yeah, interesting 🤔

src/components/profile-tab.yml
adrianaluchian: Also the browser tab title says ELMU, btw 😊 
The customization of educandu is not complete.

src/components/account-tab.js
ahelmberger: I don't think that it will be easier to read, wouldn't it be basically the same as we have here (just distributed into two places)?

src/components/default-page-template.js
ahelmberger: That is a very good question! Nobody has noticed yet :laughing:

src/components/account-tab.js
adrianaluchian: Good to keep in mind 🤔 
In this case though I believe it would not make sense as it's a one-trigger event and represents the form save

src/components/profile-tab.yml
ahelmberger: Resources are already overridable

adrianaluchian: But maybe we should make a config to pass in the project name 🤔 So that the implementing project does not have to overwrite each message.

laurentiu-ilici: I guess that will be a another ticket though, I'll resolve this comment.

src/components/default-page-template.js
laurentiu-ilici: I've made a ticket to reconsider, we can delete it later if we don't need it: https://educandu.atlassian.net/browse/EDU-203

laurentiu-ilici (approved):

src/components/default-page-template.js
adrianaluchian: Good idea, let's reconsider then.

PR #106, opened 2021-12-21
EDU-165 Add rooms tab

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-165

### DIFF
--- a/src/components/pages/import-batch-view.js
+++ b/src/components/pages/import-batch-view.js
@@ -143,7 +143,7 @@ function ImportBatchView({ initialState, PageTemplate }) {
         </Row>
         <Row>
           <Space>
-            <span>{t('createdOn')}:</span>
+            <span>{t('common:createdOn')}:</span>
             <span>{renderDate(batch.createdOn)}</span>
           </Space>
         </Row>
@@ -155,7 +155,7 @@ function ImportBatchView({ initialState, PageTemplate }) {
         </Row>
         <Row>
           <Space>
-            <span>{t('createdBy')}:</span>
+            <span>{t('common:createdBy')}:</span>
             <span>{batch.createdBy.username}</span>
           </Space>
         </Row>
--- a/src/components/pages/import-batch-view.js
+++ b/src/components/pages/import-batch-view.js
@@ -104,7 +104,7 @@ function ImportBatchView({ initialState, PageTemplate }) {
 
   const attemtpsTableColumns = [
     { title: t('startedOn'), dataIndex: 'startedOn', width: '50px', render: renderDate },
-    { title: t('completedOn'), key: '_id', dataIndex: 'completedOn', width: '150px', render: renderDate },
+    { title: t('common:completedOn'), key: '_id', dataIndex: 'completedOn', width: '150px', render: renderDate },
     { title: t('errorsCount'), width: '150px', render: renderErrorCount }
   ];
 
@@ -149,7 +149,7 @@ function ImportBatchView({ initialState, PageTemplate }) {
         </Row>
         <Row>
           <Space>
-            <span>{t('completedOn')}:</span>
+            <span>{t('common:completedOn')}:</span>
             <span>{renderDate(batch.completedOn)}</span>
           </Space>
         </Row>
--- a/src/components/pages/import-batch-view.yml
+++ b/src/components/pages/import-batch-view.yml
@@ -19,9 +19,6 @@ taskStatusFailed:
 taskStatusDoneWithWarnings:
   en: Done (with warnings)
   de: Abgeschlossen (mit Warnungen)
-completedOn:
-  en: Completed on
-  de: Abgeschlossen am
 batchStatus:
   en: Batch status
   de: Chargenstatus
--- a/src/components/pages/import-batch-view.yml
+++ b/src/components/pages/import-batch-view.yml
@@ -19,15 +19,9 @@ taskStatusFailed:
 taskStatusDoneWithWarnings:
   en: Done (with warnings)
   de: Abgeschlossen (mit Warnungen)
-createdOn:
-  en: Created on
-  de: Erstellt am
 completedOn:
   en: Completed on
   de: Abgeschlossen am
-createdBy:
-  en: Created by
-  de: Erstellt von
 batchStatus:
   en: Batch status
   de: Chargenstatus
--- a/src/components/rooms-tab.js
+++ b/src/components/rooms-tab.js
@@ -60,7 +60,7 @@ function RoomsTab() {
 
   const renderOwner = owner => {
     return owner.email
-      ? <span>{owner.username} | <a href={`mailto:${owner.email}`}>{t('email')}</a></span>
+      ? <span>{owner.username} | <a href={`mailto:${owner.email}`}>{t('owner')}</a></span>
       : <span>{owner.username}</span>;
   };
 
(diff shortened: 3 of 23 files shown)


src/components/pages/import-batch-view.js
```diff
@@ -155,7 +155,7 @@ function ImportBatchView({ initialState, PageTemplate }) {
         </Row>
         <Row>
           <Space>
-            <span>{t('createdBy')}:</span>
+            <span>{t('common:createdBy')}:</span>
```
ahelmberger, line 158: :+1:

src/components/rooms-tab.js
```diff
@@ -0,0 +1,153 @@
...
+      sorter: by(x => x.access),
+      width: '150px'
+    }
+  ];
+
+  return (
+    <div className="RoomsTab">
+      <section>
+        <h2 className="RoomsTab-sectionHeader">{t('ownedRoomsHeader')}</h2>
+        <Table dataSource={getOwnedRooms(user)} columns={ownedRoomsColumns} size="middle" />
```
laurentiu-ilici, line 143: I know I've used tables in the project, but we should probably stay away from them when it comes to pages where the normal users operate, they don't really work on mobile. Should we discuss this and what to actually use?

src/components/rooms-tab.js
```diff
@@ -0,0 +1,153 @@
...
+      : <span>{room.owner.username}</span>;
+  };
+
+  const renderJoinedOn = (_value, room) => {
+    const userAsMember = room.members.find(member => member.userId === user._id);
+    return <span>{formatDate(userAsMember.joinedOn)}</span>;
+  };
+
+  const renderAccess = (_value, room) => {
+    return <span>{t(`accessType_${room.access}`)}</span>;
```
laurentiu-ilici, line 73: The _value is already the access level as defined by the data key in the column below

src/components/rooms-tab.js
```diff
@@ -0,0 +1,153 @@
...
+      title: t('joinedOn'),
+      dataIndex: 'joinedOn',
+      key: 'joinedOn',
+      render: renderJoinedOn,
+      defaultSortOrder: 'descend',
+      sorter: by(x => x.joinedOn),
+      width: '150px'
+    },
+    {
+      title: t('access'),
```
ahelmberger, line 130: It's questionable if we wanna show this column as there is no membership in public rooms anyway. But we can keep it for now until we are further into the epic.

src/components/rooms-tab.js
```diff
@@ -0,0 +1,153 @@
...
+
+  const renderName = (_value, room) => {
+    return <a href={fakeGetRoomUrl(room._id)}>{room.name}</a>;
+  };
+
+  const renderCreatedOn = (_value, room) => {
+    return <span>{formatDate(room.createdOn)}</span>;
+  };
+
+  const renderOwner = (_value, room) => {
```
laurentiu-ilici, line 61: The _value is the owner object I believe.

src/components/pages/import-batch-view.yml
```diff
@@ -19,15 +19,9 @@ taskStatusFailed:
 taskStatusDoneWithWarnings:
   en: Done (with warnings)
   de: Abgeschlossen (mit Warnungen)
-createdOn:
-  en: Created on
-  de: Erstellt am
 completedOn:
```
ahelmberger, line 22: Maybe this is something that could go into `common` too, what do you think? It's very general.

src/components/rooms-tab.js
```diff
@@ -0,0 +1,153 @@
...
+    return room.owner.email
+      ? <span>{room.owner.username} | <a href={`mailto:${room.owner.email}`}>{t('email')}</a></span>
+      : <span>{room.owner.username}</span>;
+  };
+
+  const renderJoinedOn = (_value, room) => {
```
laurentiu-ilici, line 67: You could make the column's dataInde be "members" and ger skip the second param

src/components/rooms-tab.js
```diff
@@ -0,0 +1,153 @@
...
+    members: [{ userId: user._id, joinedOn: new Date().toISOString() }]
+  }
+];
+
+function RoomsTab() {
+  const user = useUser();
+  const { formatDate } = useDateFormat();
+  const { t } = useTranslation('roomsTab');
+
+  const renderName = (_value, room) => {
```
laurentiu-ilici, line 53: The _value is already the createdOn field

src/components/rooms-tab.js
ahelmberger: I think we can do a big table overhaul one day. Same goes for the search results. I wouldn't put too much thought into it right now, we don't know what else has to go on these pages yet.

src/components/rooms-tab.js
adrianaluchian: Yes it's hard to tell for me without real data so far, so I'd also keep this as a helper and reduce it later if needed.

src/components/rooms-tab.js
adrianaluchian: Agree, I wouldn't deviate from the pattern with this ticket. We already have the Search page which is all users facing and using a table.
We should revisit the layout choices independently so to do it consistently.

src/components/pages/import-batch-view.yml
adrianaluchian: True, it's used on 2 import pages.
Extracted.

ahelmberger (approved):

src/components/rooms-tab.js
adrianaluchian: True! I perpetuated the inconsistencies and confusions to this page 🤦‍♀️ 
Corrected, thanks!

src/components/rooms-tab.js
adrianaluchian: Changed

src/components/rooms-tab.js
adrianaluchian: Didn't get this comment at first 😬 
I did one better here actually, I mapped columns to rows, as that would be the proper way to do it anyway.

src/components/rooms-tab.js
adrianaluchian: Yup, changed.

src/components/rooms-tab.js
adrianaluchian: By the way, I streamlined and corrected the other tables too.

src/components/rooms-tab.yml
```diff
@@ -0,0 +1,24 @@
...
+  de: Name
+access:
+  en: Access
+  de: Zugang
+joinedOn:
+  en: Joined on
+  de: Eingetreten am
+owner:
+  en: Owner
+  de: Besitzer
```
adrianaluchian, line 24: @ahelmberger Would this be the proper translation for room owner?

laurentiu-ilici: I used Eigentümer, if so, I'll also change my PR to have consistency

src/components/rooms-tab.js
ahelmberger: :+1:

laurentiu-ilici (approved):

src/components/rooms-tab.yml
adrianaluchian: Extracted to `owner` to `common`

PR #107, opened 2021-12-21
EDU-168 Add endpoint for creating invitations and sending the link

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-168

### DIFF
--- a/src/domain/schemas/document-schemas.js
+++ b/src/domain/schemas/document-schemas.js
@@ -1,7 +1,6 @@
 import joi from 'joi';
+import { idOrKeySchema } from './shared-schemas.js';
 import { slugValidationPattern } from '../../common/validation-patterns.js';
-
-export const idOrKeySchema = joi.string().alphanum().min(15).max(30);
 
 const sectionSchema = joi.object({
   key: idOrKeySchema.required(),
--- a/src/domain/schemas/rooms-schemas.js
+++ b/src/domain/schemas/rooms-schemas.js
@@ -1,13 +1,13 @@
 import joi from 'joi';
+import { idOrKeySchema } from './shared-schemas.js';
 import { ROOM_ACCESS_LEVEL } from '../../common/constants.js';
 
-export const roomSchema = joi.object({
+export const postRoomBodySchema = joi.object({
   name: joi.string().required(),
-  owner: joi.string().required(),
   access: joi.string().valid(ROOM_ACCESS_LEVEL.private, ROOM_ACCESS_LEVEL.public).required()
 });
 
-export const roomInvitationSchema = joi.object({
-  roomId: joi.string().required(),
+export const postRoomInvitationBodySchema = joi.object({
+  roomId: idOrKeySchema.required(),
   email: joi.string().required()
 });
--- a/src/server/room-controller.js
+++ b/src/server/room-controller.js
@@ -1,20 +1,53 @@
+import express from 'express';
+import urls from '../utils/urls.js';
 import permissions from '../domain/permissions.js';
 import RoomService from '../services/room-service.js';
-import { roomSchema } from '../domain/schemas/rooms-schemas.js';
+import MailService from '../services/mail-service.js';
+import requestHelper from '../utils/request-helper.js';
 import { validateBody } from '../domain/validation-middleware.js';
 import needsPermission from '../domain/needs-permission-middleware.js';
+import { postRoomBodySchema, postRoomInvitationBodySchema } from '../domain/schemas/rooms-schemas.js';
+
+const jsonParser = express.json();
 
 export default class RoomController {
-  static get inject() { return [RoomService]; }
+  static get inject() { return [RoomService, MailService]; }
 
-  constructor(roomService) {
+  constructor(roomService, mailService) {
     this.roomService = roomService;
+    this.mailService = mailService;
+  }
+
+  async handlePostRoom(req, res) {
+    const { user } = req;
+    const { name, access } = req.body;
+    const newRoom = await this.roomService.createRoom({ name, access, user });
+    res.status(201).send(newRoom);
+  }
+
+  async handlePostRoomInvitation(req, res) {
+    const { user } = req;
+    const { roomId, email } = req.body;
+    const { room, owner, invitation } = await this.roomService.createOrUpdateInvitation({ roomId, email, user });
+
+    const { origin } = requestHelper.getHostInfo(req);
+    const invitationLink = urls.concatParts(origin, urls.getConfirmRoomMembershipUrl(invitation.token));
+    await this.mailService.sendRoomInvitation({ roomName: room.name, ownerName: owner.username, email, invitationLink });
+
+    res.status(201).send(invitation);
   }
 
   registerApi(router) {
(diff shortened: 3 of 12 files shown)


src/server/room-controller.js
```diff
@@ -1,20 +1,53 @@
+import express from 'express';
```
ahelmberger, line 1: I implemented the controller according to what we agreed on in the start of the project (registration of endpoint and handler separated but in one file)

src/server/room-controller.js
```diff
@@ -1,20 +1,53 @@
...
+
+    const { origin } = requestHelper.getHostInfo(req);
+    const invitationLink = urls.concatParts(origin, urls.getConfirmRoomMembershipUrl(invitation.token));
+    await this.mailService.sendRoomInvitation({ roomName: room.name, ownerName: owner.username, email, invitationLink });
+
+    res.status(201).send(invitation);
   }
 
   registerApi(router) {
-    router.post('/api/v1/rooms', [needsPermission(permissions.CREATE_ROOMS), validateBody(roomSchema)], async (req, res) => {
```
ahelmberger, line 15: BTW: I realized that the `validateBody` doesn't throw if there is no body (because there was no JSON registered here) :scream:

laurentiu-ilici: Hmm, didn't know that, sorry

src/server/room-controller.js
laurentiu-ilici: I'll switch my stuff to this style as well when we merge. I totally forgot about it

src/services/room-service.spec.js
```diff
@@ -1,32 +1,110 @@
...
+      expect(invitation.token).toBeDefined();
+    });
+
+    it('should update an invitation if it already exists', async () => {
+      const { invitation: originalInvitation } = await sut.createOrUpdateInvitation({ roomId: myPrivateRoom._id, email: 'invited-user@test.com', user: myUser });
+      const { invitation: updatedInvitation } = await sut.createOrUpdateInvitation({ roomId: myPrivateRoom._id, email: 'invited-user@test.com', user: myUser });
+      expect(updatedInvitation._id).toBe(originalInvitation._id);
+      expect(updatedInvitation.token).not.toBe(originalInvitation.token);
+      expect(updatedInvitation.sentOn).not.toBe(originalInvitation.sentOn);
+      expect(updatedInvitation.expires).not.toBe(originalInvitation.expires);
```
laurentiu-ilici, line 87: The assertion could be that the new expires needs to be higher than the old one. But I'm not insisting.

laurentiu-ilici (approved):

src/services/room-service.spec.js
ahelmberger: Good point, I'll do that!

ahelmberger: Done

src/domain/schemas/document-schemas.js
```diff
@@ -1,8 +1,7 @@
 import joi from 'joi';
+import { idOrKeySchema } from './shared-schemas.js';
```
adrianaluchian, line 2: Funny, I was introducing `generic-schemas` also in my current ticket 🧠 
Will use yours 👍

src/domain/schemas/rooms-schemas.js
```diff
@@ -1,13 +1,13 @@
 import joi from 'joi';
+import { idOrKeySchema } from './shared-schemas.js';
 import { ROOM_ACCESS_LEVEL } from '../../common/constants.js';
 
-export const roomSchema = joi.object({
+export const postRoomBodySchema = joi.object({
```
adrianaluchian, line 5: And made this exact change too 😂

ahelmberger: Gimme back ma brain :laughing:

src/services/room-service.js
```diff
@@ -1,21 +1,76 @@
...
+      owner: user._id,
       members: []
     };
 
-    await this.roomsStore.save(newRoom);
+    await this.roomStore.save(newRoom);
     return newRoom;
   }
+
+  async findOwnedRoomById({ roomId, ownerId }) {
```
adrianaluchian, line 38: What do you think about moving this logic directly in `createOrUpdateInvitation` and throwing a separate cleare error if `room.owner !== ownerId` ?

src/server/room-controller.js
adrianaluchian: > I implemented the controller according to what we agreed on in the start of the project (registration of endpoint and handler separated but in one file)

I think so far this is the first such pattern.
Cool, let's start changing like this then 👍

src/services/mail-service.js
```diff
@@ -71,7 +71,41 @@ class MailService {
...
+
+    const englishHtml
+      = '<p>Hello!<br/><br/>'
+      + `User <b>${ownerName}</b> has invited you to the room <b>${roomName}</b>.<br/>`
+      + `Please click here in order to confirm the invitation: <a href="${invitationLink}">Join room</a>.</p>`;
+
+    const message = {
+      from: this.emailSenderAddress,
+      to: email,
+      subject: `Sie wurden in den Raum ${roomName} eingeladen / You have been invited to room ${roomName}`,
```
adrianaluchian, line 108: Room names could be long, should we maybe have a generic subject?
`"Sie wurden in einen Raum eingeladen / You have been invited to a room"` ?

src/domain/schemas/rooms-schemas.js
adrianaluchian: It's such a big one, we could share 😄

src/services/room-service.js
ahelmberger: I had this first, then moved it here cause we're gonna need that logic with the owner check more often in the future, but I agree that this is premature optimization. What do you think about throwing the "NotFound" error directly here instead of in the consuming code, then we ensure consistent behavior?

src/services/mail-service.js
ahelmberger: Good idea, will do!

ahelmberger: Done!

src/domain/schemas/rooms-schemas.js
ahelmberger: No comment!

src/services/room-service.js
adrianaluchian: That's also a good option, yes.
We could still throw 2 errors, one for room not found and one for owner not matching 🤔 I think it would be a useful distinction.

src/domain/schemas/document-schemas.js
ahelmberger: Perfect

src/services/room-service.js
ahelmberger: Actually, let's just log that but throw 404 in both cases, if a user is not a room owner it's non of their business if the room exists or not (thus showing "does not exist").

adrianaluchian: That's true.

ahelmberger: Changed!

adrianaluchian (approved):

PR #108, opened 2021-12-21
EDU-196 room details page

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-196

### DIFF
--- a/src/components/pages/room.yml
+++ b/src/components/pages/room.yml
@@ -1,0 +1,15 @@
+ownerUsername:
+  en: Owner user name
+  de: Benutzername des Eigentümers
+ownerEmail:
+  en: Owner email
+  de: Email des Eigentümers
+memberUsername:
+  en: Member username
+  de: Mitglieder Benutzername
+memberEmail:
+  en: Member email
+  de: Mitglieder email
+joinedOn:
+  en: Member since
+  de: Mitglied seit
--- a/src/components/pages/room.yml
+++ b/src/components/pages/room.yml
@@ -1,15 +1,9 @@
 ownerUsername:
   en: Owner user name
   de: Benutzername des Eigentümers
-ownerEmail:
-  en: Owner email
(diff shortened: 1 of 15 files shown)


src/components/pages/room.yml
```diff
@@ -0,0 +1,15 @@
...
+  de: Benutzername des Eigentümers
+ownerEmail:
+  en: Owner email
+  de: Email des Eigentümers
+memberUsername:
+  en: Member username
+  de: Mitglieder Benutzername
+memberEmail:
+  en: Member email
+  de: Mitglieder email
```
ahelmberger, line 12: Miglieder-Email

laurentiu-ilici: Fixed

src/services/room-service.js
```diff
@@ -18,4 +20,32 @@ export default class RoomService {
...
+
+    roomDetails.members = roomDetails.members.map(member => {
+      const memberDetails = members.find(user => member.userId === user._id);
+      return {
+        ...member,
+        email: memberDetails.email,
+        username: memberDetails.username
+      };
+    });
+
```
ahelmberger, line 48: I'm a little bit unsure about this approach. Usually (and by "usually" I mean: in case of documents and revisions) the service returns more or less the raw data objects from the DB, and then the client data mapper resolves user IDs (`createdBy` etc.) to the actual user objects (with more or less user data being there based on the permissions of the current user). This approach now constructs the full data model for the UI inside the service. I'd like to know your opinion on whether it wouldn't make sense to do it here like we do with documents (given the fact that probably room members should not be able to see the email addresses of other room members).

laurentiu-ilici: Moved to client-data-mapper.

ahelmberger: :+1: thanks

ahelmberger (approved):

PR #110, opened 2021-12-22
EDU-198 add authorization endpoint

laurentiu-ilici:


src/server/room-controller.spec.js
```diff
@@ -29,13 +29,20 @@ describe('room-controller', () => {
           return Promise.resolve(room);
         }
         return Promise.resolve();
+      }),
+      isRoomMemberOrOwner: sandbox.stub().callsFake(roomId => {
```
ahelmberger, line 33: The `.callsFake(...` part handles a very specific scenario used in one `describe` block, so I think it would be better to move that into the `beforeEach` of that block, leaving here only `isRoomMemberOrOwner: sandbox.stub()`, so other tests can do other things with it.

BTW: I did that already for the function above (`getRoomById`) in my branch.

src/services/room-service.js
```diff
@@ -77,4 +77,8 @@ export default class RoomService {
   getRoomById(roomId) {
     return this.roomStore.findOne({ _id: roomId });
   }
+
+  isRoomMemberOrOwner(roomId, userId) {
+    return this.roomStore.findOne({ $or: [{ '_id': roomId, 'members.userId': userId }, { _id: roomId, owner: userId }] });
```
ahelmberger, line 82: Given the function name I'd expect a real boolean

src/services/room-service.spec.js
```diff
@@ -117,4 +121,37 @@ describe('room-service', () => {
...
+    beforeEach(async () => {
+      await roomStore.save({
+        _id: roomId,
+        name: 'my room',
+        access: ROOM_ACCESS_LEVEL.private,
+        owner: myUser._id,
+        members: [
+          {
+            userId: otherUser._id,
+            joinedOn: new Date().toISOString()
```
ahelmberger, line 136: No need for `.toISOString()`, we always save real dates

src/services/room-service.js
laurentiu-ilici: I had big trouble with naming this and that was my problem with this name as well, I'm just afraid we will end up having this same query in another context. I'll change it to returning a boolean.

src/services/room-service.spec.js
laurentiu-ilici: Oh, okay. fixed.

src/server/room-controller.js
```diff
@@ -61,6 +61,14 @@ export default class RoomController {
     return this.pageRenderer.sendPage(req, res, PAGE_NAME.room, { roomDetails });
   }
 
+  async handleAuthorizeResourceAccess(req, res) {
+    const { roomId } = req.params;
+    const { _id: userId } = req.user;
+
+    const result = await this.roomService.isRoomMemberOrOwner(roomId, userId);
+    return result ? res.sendStatus(200) : res.sendStatus(403);
```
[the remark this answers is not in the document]
laurentiu-ilici: Fixed

src/services/room-service.js
laurentiu-ilici: Fixed

src/server/room-controller.spec.js
laurentiu-ilici: Fixed

src/server/room-controller.js
```diff
@@ -77,6 +89,12 @@ export default class RoomController {
       [needsPermission(permissions.CREATE_ROOMS), jsonParser, validateBody(postRoomInvitationBodySchema)],
       (req, res) => this.handlePostRoomInvitation(req, res)
     );
+
+    router.get(
+      '/api/v1/rooms/:roomId/authorize-resource-access',
```
adrianaluchian, line 94: Small thing: maybe `authorize-resources-access`, since it's not for a specific resource.

src/services/room-service.js
```diff
@@ -77,4 +77,9 @@ export default class RoomService {
   getRoomById(roomId) {
     return this.roomStore.findOne({ _id: roomId });
   }
+
+  async isRoomMemberOrOwner(roomId, userId) {
+    const room = await this.roomStore.findOne({ $or: [{ '_id': roomId, 'members.userId': userId }, { _id: roomId, owner: userId }] });
```
adrianaluchian, line 82: To me it would be more logical to do 
```
{ $and: [{ _id: roomId }, { $or: [{ 'members.userId': userId }, { owner: userId }] }] }
``` 
and would also remove the duplication.

I would not insist though, but if you keep the duplication then I'd ask to please make it consistent and go with either `'_id': roomId` or `_id: roomId`

src/services/room-service.spec.js
```diff
@@ -117,4 +121,37 @@ describe('room-service', () => {
...
+        members: [
+          {
+            userId: otherUser._id,
+            joinedOn: new Date()
+          }
+        ]
+      });
+    });
+
+    it('should return truthy when the user is the owner', async () => {
```
adrianaluchian, line 142: I think we should check for `true` and not truthy.
To me truthy is when you don't care about the result, it just has to return something. Here we actually expect a boolean.

src/server/room-controller.js
```diff
@@ -77,6 +89,12 @@ export default class RoomController {
...
+      [needsPermission(permissions.AUTORIZE_ROOMS_RESOURCES), validateParams(getAuthorizeResourceAccessParamsSchema)],
```
adrianaluchian, line 95: If you do change this guy https://github.com/educandu/educandu/pull/110/files#r773989272 then please also change `getAuthorizeResourceAccessParamsSchema` to keep it consistent 🙏

adrianaluchian: And `handleAuthorizeResource**s**Access` 😅

src/services/room-service.spec.js
laurentiu-ilici: I changed the code and forgot about the assertion (it used to be a truthy). Changed thanks!

src/services/room-service.js
laurentiu-ilici: Changed

src/server/room-controller.js
laurentiu-ilici: Fixed, thanks!

src/services/room-service.js
ahelmberger: Thanks

ahelmberger (approved):

src/server/room-controller.js
laurentiu-ilici: Fixed 👍

adrianaluchian (approved):

PR #111, opened 2021-12-22
EDU-204 display real data on rooms tab

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-204

**For ease of testing**

- users that can be inserted in the DB (where password is "test")
```
{
    "_id" : "5ME6QbzSdygAbGSv98YTv8",
    "provider" : "educandu",
    "username" : "roomstest1",
    "passwordHash" : "$2b$04$B2mPmYMZQDVbMKnR1dNlJ.9BxeFonMus8B7FHpRe8h48QRVm8JYNy",
    "email" : "roomestest1@test.com",
    "roles" : [ 
        "user", 
        "admin"
    ],
    "expires" : null,
    "verificationCode" : null,
    "lockedOut" : false,
    "profile" : null
},
{
    "_id" : "4ME6QbzSdygAbGSv98YTv8",
    "provider" : "educandu",
    "username" : "roomstest2",
    "passwordHash" : "$2b$04$B2mPmYMZQDVbMKnR1dNlJ.9BxeFonMus8B7FHpRe8h48QRVm8JYNy",
    "email" : "roomestest2@test.com",
    "roles" : [ 
        "user"
    ],
    "expires" : null,
    "verificationCode" : null,
    "lockedOut" : false,
    "profile" : null
}
```
- rooms that can be inserted in the DB to cover all cases
```
[
    {
        "_id": "s6hueCbWrbP87U3CFzPg4F",
        "name": "Room 1",
        "createdOn": ISODate("2021-12-21T13:26:00.388Z"),
        "access": "public",
        "owner": "5ME6QbzSdygAbGSv98YTv8",
        "members": []
    },
    {
        "_id": "ae4iWz8HT5fMKV4BbxFPDH",
        "name": "Room 2",
        "createdOn": ISODate("2021-12-21T13:26:00.388Z"),
        "access": "private",
        "owner": "5ME6QbzSdygAbGSv98YTv8",
        "members": []
    },
    {
        "_id": "9ME6QbzSdygAbGSv98YTv8",
        "name": "Room 1",
        "createdOn": ISODate("2021-12-21T13:26:00.388Z"),
        "access": "public",
        "owner": "4ME6QbzSdygAbGSv98YTv8",
        "members": [
            {
                "userId": "5ME6QbzSdygAbGSv98YTv8",
                "joinedOn": ISODate("2021-12-21T13:26:00.388Z")
            }
        ]
    },
    {
        "_id": "8ME6QbzSdygAbGSv98YTv8",
        "name": "Room 2",
        "createdOn": ISODate("2021-12-21T13:26:00.388Z"),
        "access": "private",
        "owner": "4ME6QbzSdygAbGSv98YTv8",
        "members": [
            {
                "userId": "5ME6QbzSdygAbGSv98YTv8",
                "joinedOn": ISODate("2021-12-21T13:26:00.388Z")
            }
        ]
    }
]
```


src/services/room-service.js
```diff
@@ -22,12 +22,48 @@ export default class RoomService {
...
+    if (andFilters.length === 1) {
+      filter = andFilters[0];
+    }
+    if (andFilters.length > 1) {
+      filter = { $and: andFilters };
+    }
+
+    const rooms = await this.roomStore.find(filter);
+    return rooms;
+  }
```
adrianaluchian, line 57: Since adding this method initially, I ended up rethinking the logic and not needing `andFilters`.
However, I kept this method as is, since I thought it's a good base setup for the future and has proper tests.

What's your take on it? Should we reduce it to only have `{ ownerId, memberId }` logic, as `access` is not used so far?

src/components/document-import-table.js
```diff
@@ -1,7 +1,7 @@
 import by from 'thenby';
 import { Table } from 'antd';
 import PropTypes from 'prop-types';
-import { getDocUrl } from '../utils/urls.js';
+import urls from '../utils/urls.js';
```
adrianaluchian, line 4: Bonus commit here:
- changed `urls` usage towards most common usages and adjusted exports accordingly

src/server/user-controller.js
```diff
@@ -1,15 +1,21 @@
 import express from 'express';
```
adrianaluchian, line 1: Bonus commit here:

- moved the methods from the `user-request-handler` in here to follow the new controller pattern (since I had to touch this file anyway)

src/server/my-space-controller.js
```diff
@@ -0,0 +1,47 @@
...
+    this.roomService = roomService;
+    this.pageRenderer = pageRenderer;
+    this.clientDataMapper = clientDataMapper;
+  }
+
+  async handleGetMySpacePage(req, res) {
+    const { user } = req;
+
+    let rooms = [];
+    if (!this.serverConfig.disabledFeatures.includes(FEATURE_TOGGLES.rooms)) {
```
laurentiu-ilici, line 23: This would be better put into the "register pages". It was done like that the room-controller.

src/services/room-service.js
laurentiu-ilici: You can leave it, but it seems a bit involved int his way. I think we can write this more easily, we can take a look if you want.

src/utils/urls.js
```diff
@@ -33,117 +34,121 @@ function concatParts(...parts) {
     .reduce((prev, next) => `${removeTrailingSlash(prev)}/${removeLeadingSlash(next)}`);
 }
 
-export function createRedirectUrl(path, redirect) {
```
laurentiu-ilici, line 36: Why did we need this change?

src/services/room-service.js
adrianaluchian: "but it seems a bit involved int his way." what do you mean by this?

src/server/my-space-controller.js
adrianaluchian: It's not the same thing, we do want the registration to happen in this case, we just bypass the data fetching.

laurentiu-ilici: Ah, ok

src/components/pages/room.js
```diff
@@ -42,6 +42,6 @@ export default function Room({ PageTemplate, initialState }) {
 Room.propTypes = {
   PageTemplate: PropTypes.func.isRequired,
   initialState: PropTypes.shape({
-    roomDetails: roomDetailsShape
+    roomDetails: roomShape
```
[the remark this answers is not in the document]
adrianaluchian: Hmm, the room models have little data, I don't think there is more on the room details page to be added, right?
I would keep it simple as long as we don't deal with big data models.

src/domain/schemas/shared-schemas.js
```diff
@@ -1,3 +1,5 @@
 import joi from 'joi';
 
 export const idOrKeySchema = joi.string().alphanum().min(15).max(30);
+export const boolStringSchema = joi.any().allow('true', 'false');
```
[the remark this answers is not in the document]
adrianaluchian: I just moved this here, but I will check.

src/services/room-service.js
laurentiu-ilici: I think this should work in the same way:
```
const ownerFilter = {
      owner: ownerId,
      ...access && { access }
    };

    const memberFilter = {
      'members.userId': memberId,
      ...access && { access }
    };

    if (ownerId) {
      orFilters.push(ownerFilter);
    }
    if (memberId) {
      orFilters.push(memberFilter);
    }

    const filter = orFilters.length ? orFilters : {};

    const rooms = await this.roomStore.find(filter);
```

src/domain/schemas/shared-schemas.js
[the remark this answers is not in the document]
laurentiu-ilici: Ah, also it should be joi.string() since the name says "boolString". We discovered yesterday that allow('') adds on top of allowed things while "valid" restricts the possibilities to exactly that which is specified.

src/services/room-service.js
adrianaluchian: If you change the code and place yours instead, 6 tests fail 😬 

I think I might have not made it clear. `access` is not passed in with the current usage, it's `undefined`. So can all the others. I wanted to make a generic method that can be scaled up and can support lacking of any of the arguments.

src/domain/schemas/shared-schemas.js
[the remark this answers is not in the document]
adrianaluchian: Yup, the documentation regarding `allow` is clear

"Note that this list of allowed values is in addition to any other permitted values. To create an exclusive list of values, see any.valid(value)."

Thanks for pointing it out, changed.

src/services/room-service.js
laurentiu-ilici: But this is as generic, maybe you need to spread the  access also on the stand alone filter. 
The tests fail because they expect the filter to look in a certain way instead of actually querying the db or?
I'm currently checking with the code on local.

src/domain/schemas/shared-schemas.js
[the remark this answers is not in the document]
adrianaluchian: > Ah, also it should be joi.string() since the name says "boolString". We discovered yesterday that allow('') adds on top of allowed things while "valid" restricts the possibilities to exactly that which is specified.

I kept it `any` and `valid("true", "false", true, false)`, this way this will work regardless of json parsing having been performed before or not.

src/services/room-service.js
laurentiu-ilici: It works the same with your db setup and this code on local :).
```
 async getRooms({ ownerId, memberId, access }) {
    const orFilters = [];
    const ownerFilter = {
      owner: ownerId,
      ...access && { access }
    };

    const memberFilter = {
      'members.userId': memberId,
      ...access && { access }
    };

    if (ownerId) {
      orFilters.push(ownerFilter);
    }
    if (memberId) {
      orFilters.push(memberFilter);
    }

    const filter = orFilters.length ? orFilters : { ...access && { access } };

    const rooms = await this.roomStore.find(filter);
    return rooms;
  }
```

adrianaluchian: Some fail because the Mongo query is invalid.
In your code you combine `access` into the other 2 filters, but that's not what is needed. For example, you could use this method to simply `getRooms({ access: 'private' })`. With the current code, this works, with your change it doesn't.
This is what I mean by generic.

src/components/pages/room.js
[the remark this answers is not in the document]
laurentiu-ilici: I think there will be more coming from the server, we won't save it in the room collection, but there will be other stuff mapped on to the room details which is aggregated from other collections and used by the page.
But we will change the code, no problem. This is why I made this shape for the details page rather than a simple room shape (lazy loading when it comes to loading rooms without the details).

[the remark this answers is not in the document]
adrianaluchian: We don't usually aggregate stuff, we pass it as separate objects. I think the only thing we fill in is probably the user 🤔 
Yeah I would change it if and when we need to, so far I would really keep it simple with this data volume and considering so far both pages use the same data.

src/utils/urls.js
```diff
@@ -1,23 +1,24 @@
...
+const createImportPath = '/import-batches/create';
+
+const docsPrefix = '/docs/';
+const revisionPrefix = '/revs/';
+const editDocPrefix = '/edit/doc/';
+const roomDetailsPrefix = '/rooms/';
+const pluginApiPathPrefix = '/plugins/';
+const completeRegistrationPrefix = '/complete-registration/';
+const completePasswordResetPrefix = '/complete-password-reset/';
+const confirmRoomMembershipPrefix = '/confirm-room-membership/';
```
ahelmberger, line 21: Nice, soon we can enable the "don't double-export" eslint rule!

laurentiu-ilici (approved):

src/components/pages/my-space.js
```diff
@@ -70,7 +73,10 @@ function MySpace({ PageTemplate }) {
 }
 
 MySpace.propTypes = {
-  PageTemplate: PropTypes.func.isRequired
+  PageTemplate: PropTypes.func.isRequired,
+  initialState: PropTypes.shape({
+    rooms: PropTypes.arrayOf(roomShape)
```
ahelmberger, line 78: Should this not also be required (I mean the room array)?

adrianaluchian: True, even when we bypass fetching the rooms, the array is still set. Changed, thanks.

src/utils/urls.js
adrianaluchian: No clue how far we are from that rule yet, but one day maybe 🤞

src/server/my-space-controller.js
```diff
@@ -0,0 +1,47 @@
...
+      rooms = await this.roomService.getRooms({ ownerId: user._id, memberId: user._id });
+    }
+
+    const mappedRooms = [];
+    for (const room of rooms) {
+      // eslint-disable-next-line no-await-in-loop
+      mappedRooms.push(await this.clientDataMapper.mapRoom(room, user));
+    }
```
ahelmberger, line 31: Let's do them in parallel:

~~~
const mappedRooms = await Promise.all(rooms.map(...
~~~

src/server/user-controller.js
ahelmberger: :+1:

src/server/my-space-controller.js
adrianaluchian: Of course, no need to serialize this. Changed!

src/services/room-service.js
adrianaluchian: @ahelmberger, @laurentiu-ilici, I reduced `getRooms` how we discussed, please take a look.

src/server/my-space-controller.js
```diff
@@ -0,0 +1,47 @@
...
+
+    let rooms = [];
+    if (!this.serverConfig.disabledFeatures.includes(FEATURE_TOGGLES.rooms)) {
+      rooms = await this.roomService.getRooms({ ownerId: user._id, memberId: user._id });
+    }
+
+    const mappedRooms = [];
+    for (const room of rooms) {
+      // eslint-disable-next-line no-await-in-loop
+      mappedRooms.push(await this.clientDataMapper.mapRoom(room, user));
```
laurentiu-ilici: This could be a Promise.all(room.map(...)) so we get rid of the es lint directive and we do it in parallel.

laurentiu-ilici (approved):

src/services/room-service.js
ahelmberger: :+1: Thanks!

ahelmberger (approved):

PR #113, opened 2021-12-22
EDU-205 Add button to create room

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-205

### DIFF
--- a/src/components/rooms-tab.js
+++ b/src/components/rooms-tab.js
@@ -1,24 +1,59 @@
 import by from 'thenby';
 import PropTypes from 'prop-types';
 import urls from '../utils/urls.js';
-import { Button, Table } from 'antd';
-import React, { useState } from 'react';
+import Logger from '../common/logger.js';
 import Restricted from './restricted.js';
 import { useUser } from './user-context.js';
 import { useTranslation } from 'react-i18next';
+import errorHelper from '../ui/error-helper.js';
+import React, { useState, useRef } from 'react';
 import { PlusOutlined } from '@ant-design/icons';
 import permissions from '../domain/permissions.js';
 import { useDateFormat } from './language-context.js';
 import { roomShape } from '../ui/default-prop-types.js';
 import { ROOM_ACCESS_LEVEL } from '../common/constants.js';
+import { Form, Button, Table, Modal, Input, Radio } from 'antd';
+
+const FormItem = Form.Item;
+const RadioGroup = Radio.Group;
+const RadioButton = Radio.Button;
+
+const logger = new Logger(import.meta.url);
 
 function RoomsTab({ rooms }) {
   const user = useUser();
+  const newRoomFormRef = useRef(null);
   const { formatDate } = useDateFormat();
   const { t } = useTranslation('roomsTab');
-  const [setState] = useState({
+
+  const roomNameValidationRules = [
+    {
+      required: true,
+      message: t('roomNameRequired'),
+      whitespace: true
+    }
+  ];
+
+  const isNewRoomFormValid = async () => {
+    try {
+      await newRoomFormRef.current.validateFields(['name'], { force: true });
+      return true;
+    } catch {
+      return false;
+    }
+  };
+
+  const createNewRoomState = () => {
+    return {
+      name: t('newRoom'),
+      access: ROOM_ACCESS_LEVEL.private
+    };
+  };
+
+  const [state, setState] = useState({
     isNewRoomModalVisible: false,
-    newRoom: null
+    isNewRoomBeingCreated: false,
+    newRoom: createNewRoomState()
   });
 
   const renderName = (name, room) => {
@@ -43,18 +78,42 @@ function RoomsTab({ rooms }) {
     return <span>{t(`accessType_${access}`)}</span>;
   };
 
-  const createNewRoomState = () => {
-    return {
-      name: t('defaultRoomName'),
-      access: ROOM_ACCESS_LEVEL.private
-    };
-  };
-
   const handleNewRoomClick = () => {
     setState({
       newRoom: createNewRoomState(),
-      isNewDocModalVisible: true
+      isNewRoomModalVisible: true
     });
+  };
+
+  const handleNewRoomOk = async () => {
+    try {
+      const isValid = await isNewRoomFormValid();
+      if (!isValid) {
+        return;
+      }
+
+      setState(prevState => ({ ...prevState, isNewRoomBeingCreated: true }));
+      // eslint-disable-next-line no-console
+      console.log(`In EDU-206 a new room ${JSON.stringify(state.newRoom)} will be created and redirected to`);
+      setState(prevState => ({ ...prevState, isNewRoomBeingCreated: false, isNewRoomModalVisible: false }));
+    } catch (error) {
+      errorHelper.handleApiError({ error, logger, t });
+      setState(prevState => ({ ...prevState, isNewRoomBeingCreated: false }));
+    }
+  };
+
+  const handleNewRoomCancel = () => {
+    setState(prevState => ({ ...prevState, isNewRoomModalVisible: false }));
+  };
+
+  const handleNewRoomNameChange = event => {
+    const { value } = event.target;
+    setState(prevState => ({ ...prevState, newRoom: { ...prevState.newRoom, name: value } }));
+  };
+
+  const handleRoomAccessChange = event => {
+    const { value } = event.target;
+    setState(prevState => ({ ...prevState, newRoom: { ...prevState.newRoom, access: value } }));
   };
 
   const ownedRooms = rooms.filter(room => room.owner._id === user._id);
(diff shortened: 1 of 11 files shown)


src/components/rooms-tab.js
```diff
@@ -127,11 +208,37 @@ function RoomsTab({ rooms }) {
...
+        <Restricted to={permissions.CREATE_ROOMS}>
+          <Button className="RoomsTab-newRoomButton" type="primary" shape="circle" icon={<PlusOutlined />} size="large" onClick={handleNewRoomClick} />
+        </Restricted>
       </section>
       <section className="RoomsTab-section">
         <h2>{t('joinedRoomsHeader')}</h2>
         <Table dataSource={membershipRoomsRows} columns={membershipRoomsColumns} size="middle" />
       </section>
+
+      {state.isNewRoomModalVisible && (
```
adrianaluchian, line 220: I added this so that the modal "name" input field (value + validation message) gets reset when you:
- click the add room button
- invalidate the name field
- click 'cancel'
- click the add room button again

Due to the form validation, I found no other way for this. Maybe someone has a nicer solution?

adrianaluchian:
Declining so that I call the existing controller endpoint for adding the room and the redirect.

src/resources/resources.json
```diff
@@ -820,7 +822,9 @@
       "accessType_private": "privat",
       "name": "Name",
       "access": "Zugang",
-      "joinedOn": "Eingetreten am"
+      "joinedOn": "Eingetreten am",
+      "newRoom": "Neuer Raum",
+      "roomNameRequired": "Bitte geben Sie einen Raumnamen"
```
ahelmberger, line 827: Bitte geben Sie einen Raumnamen an

ahelmberger: Oh, closed :laughing:

src/components/rooms-tab.js
ahelmberger: Yes, there is a nicer way by using the functions (like `resetFields` etc.) on the form instance (the ref you created), see also here: https://ant.design/components/form/#FormInstance

src/resources/resources.json
adrianaluchian: angeben! Of course, changed, thank you

src/components/rooms-tab.js
adrianaluchian: Since only one field needed resetting (the other one functioned just fine with the `setState`) I checked the `Form.Item` documentation, but it makes sense for this to be on the `Form` level only 🤦‍♀️ 
Thanks, changed!
Kept it in the ok handler thought, feelt it's more future proof, for example if we'd change from redirecting after creation to just updating the rooms list.

ahelmberger (approved):

src/components/rooms-tab.js
ahelmberger: :+1:

adrianaluchian: Had to make a small change to use `setFieldsValue` instead of `resetFields`, to reliably reset the `name` field.

Since the initial value is taken from the state, and the state change is async, sometimes the `name` was not changed when opening the modal.

ahelmberger: Alternatively you could not do anything when the modal opens, but reset the fields after each OK/Cancel click (on OK only after room creation, of course)

src/components/rooms-tab.js
```diff
@@ -35,6 +81,46 @@ function RoomsTab({ rooms }) {
...
+      }
+
+      setState(prevState => ({ ...prevState, isNewRoomBeingCreated: true }));
+      const newRoom = await roomApiClient.addRoom({ name: state.newRoom.name, access: state.newRoom.access });
+      setState(prevState => ({ ...prevState, isNewRoomBeingCreated: false, isNewRoomModalVisible: false }));
+
+      window.location = urls.getRoomDetailsUrl(newRoom._id);
+    } catch (error) {
+      errorHelper.handleApiError({ error, logger, t });
+      setState(prevState => ({ ...prevState, isNewRoomBeingCreated: false }));
```
laurentiu-ilici, line 106: This one can go in a finally block so we never have a chance to freeze the UI.

src/components/rooms-tab.js
```diff
@@ -127,11 +213,35 @@ function RoomsTab({ rooms }) {
...
+        <Restricted to={permissions.CREATE_ROOMS}>
+          <Button className="RoomsTab-newRoomButton" type="primary" shape="circle" icon={<PlusOutlined />} size="large" onClick={handleNewRoomClick} />
+        </Restricted>
       </section>
       <section className="RoomsTab-section">
         <h2>{t('joinedRoomsHeader')}</h2>
         <Table dataSource={membershipRoomsRows} columns={membershipRoomsColumns} size="middle" />
       </section>
+
+      <Modal
```
laurentiu-ilici, line 225: I would like it if we start splitting these modals into their own components, these files where we define them are getting too  long. What do you guys think?

laurentiu-ilici:
When I create a new room I get: 
TypeError
urls_default.getRoomDetailsUrl is not a function
This function doesn't seem to exist in the urls

laurentiu-ilici:
We could also consider using the useReloadPersistedWindow on this page, otherwise a browser back will display an incomplete list of pages if you have a new room created and click back.

src/components/rooms-tab.js
adrianaluchian: I think I would not not extract these, feels like there would be alot of props to pass down to them.

However maybe we can have model mappers for the pages/components that have tables, and extract the mapping between the DB data and the table data (columns and rows).

adrianaluchian:
> We could also consider using the useReloadPersistedWindow on this page, otherwise a browser back will display an incomplete list of pages if you have a new room created and click back.

This is the behavior on all our pages, as this is not a SPA 😬

adrianaluchian:
> When I create a new room I get: TypeError urls_default.getRoomDetailsUrl is not a function This function doesn't seem to exist in the urls

Interesting, readded it, thanks!

src/components/rooms-tab.js
adrianaluchian: The UI won't freeze, this is set in the `try` block:

`setState(prevState => ({ ...prevState, isNewRoomBeingCreated: false, isNewRoomModalVisible: false }));`

laurentiu-ilici:
> > We could also consider using the useReloadPersistedWindow on this page, otherwise a browser back will display an incomplete list of pages if you have a new room created and click back.
> 
> This is the behavior on all our pages, as this is not a SPA 😬

I know, that's why I said we can consider it like we did on other pages where we thought it makes sense. I'm not insisting on this though, it was just confusing while testing.

src/components/rooms-tab.js
laurentiu-ilici: If the error helper throws, it will (I've had the situation happen with some mismatched parameters). But okay, it's overkill.

adrianaluchian:
> > > We could also consider using the useReloadPersistedWindow on this page, otherwise a browser back will display an incomplete list of pages if you have a new room created and click back.
> > 
> > 
> > This is the behavior on all our pages, as this is not a SPA 😬
> 
> I know, that's why I said we can consider it like we did on other pages where we thought it makes sense. I'm not insisting on this though, it was just confusing while testing.

Got it, I see we use that just on import pages now. I'll made a ticket, cause maybe we should do it consistently on all pages listing content that needs re-fetching, and place this logic in a common place maybe.
https://educandu.atlassian.net/browse/EDU-216

src/components/rooms-tab.js
laurentiu-ilici: I think if you would in this example split the Modal + its form into its own component, every  state property and event handler for the form can go into the modal. Then the rooms tab becomes the tab + the tables + a button that opens the modal. 
We have the same situation on creating docs. We don't need to do it now, I was just wanting to bring it up.

laurentiu-ilici (approved):

src/components/rooms-tab.js
adrianaluchian: Ok, good point. Let's consider this then, if we'll find it decongests the pages enough, we could do it on all of them.

adrianaluchian: I decided to try this extraction out in this PR 😊 Please check.

laurentiu-ilici (approved):

src/components/rooms-tab.js
laurentiu-ilici: Cool, and you chose a better name :D

src/utils/urls.js
```diff
@@ -144,6 +144,10 @@ function getRoomUrl(roomId) {
   return concatParts(roomsPrefix, encodeURIComponent(roomId));
 }
 
+function getRoomDetailsUrl(roomId) {
+  return concatParts(roomsPrefix, roomId);
+}
```
ahelmberger, line 149: I renamed this to `getRoomUrl` (the function above) :laughing: (too late, I know).

PR #114, opened 2021-12-28
EDU-213 Streamline controllers

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-213

### DIFF
--- a/src/server/i18n-controller.js
+++ b/src/server/i18n-controller.js
@@ -7,7 +7,7 @@ const logger = new Logger(import.meta.url);
 
 class I18nController {
 
-  handleUseMiddleware(req, res, next) {
+  handleDetectUiLanguage(req, res, next) {
     const language = this._detectLanguageFromRequest(req);
     req.language = language;
     res.cookie(UI_LANGUAGE_COOKIE_NAME, language, { expires: UI_LANGUAGE_COOKIE_EXPIRES });
@@ -54,7 +54,7 @@ class I18nController {
   }
 
   registerMiddleware(router) {
-    router.use((req, res, next) => this.handleUseMiddleware(req, res, next));
+    router.use((req, res, next) => this.handleDetectUiLanguage(req, res, next));
   }
 }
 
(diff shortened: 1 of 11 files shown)


src/server/i18n-controller.js
```diff
@@ -54,6 +52,10 @@ class I18nController {
 
     return foundLanguage;
   }
+
+  registerMiddleware(router) {
+    router.use((req, res, next) => this.handleUseMiddleware(req, res, next));
```
ahelmberger, line 57: Maybe more specificly: `handleDetectUiLanguage`

src/server/user-controller.js
```diff
@@ -228,6 +181,53 @@ class UserController {
     return res.send({ lockedOut: newLockedOutState });
   }
 
+  registerMiddleware(router) {
```
ahelmberger, line 184: You extracted the registered middleware in `i18n-controller.js` but not here. It's not a huge inconsistency that would bother me, I just wanted to mention it. Maybe there's also no real benefit in extracting the registration of middlewares into their own `handlers` unless we want to test it with mocked `req`s and `res`s and `next`s (which we don't do in this case here).

src/server/i18n-controller.js
adrianaluchian: Thanks, I better named all extracted middleware handlers

ahelmberger (approved):

src/server/user-controller.js
adrianaluchian: Here I didn't want to overcomplicate it, because the handlers are basically already either `session` or the `passport` methods. Also did not want to risk introducing bugs here. However I wanted to have the middleware handler registration at least in the same spot as in the other controllers.

For `user-agent-controller` and `i18n-controller` I think we could use some middleware handlers tests.
`settings-controller` I just extracted for consistency with the others..

ahelmberger: I see, just wanted to mention it. I think it's fine.

PR #118, opened 2021-12-28
EDU-203 Show help page to all users

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-203

### DIFF
--- a/src/components/default-page-template.js
+++ b/src/components/default-page-template.js
@@ -116,7 +116,7 @@ function DefaultPageTemplate({ children, fullScreen, headerActions, alerts }) {
       href: helpPage ? urls.getDocUrl(helpPage.documentKey, helpPage.documentSlug) : '',
       text: helpPage?.linkTitle,
       icon: QuestionOutlined,
-      permission: permissions.EDIT_SETTINGS,
+      permission: null,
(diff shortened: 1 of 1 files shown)


src/components/default-page-template.js
```diff
@@ -116,7 +116,7 @@ function DefaultPageTemplate({ children, fullScreen, headerActions, alerts }) {
       href: helpPage ? urls.getDocUrl(helpPage.documentKey, helpPage.documentSlug) : '',
       text: helpPage?.linkTitle,
       icon: QuestionOutlined,
-      permission: permissions.EDIT_SETTINGS,
+      permission: null,
```
laurentiu-ilici, line 119: It might be that we will need specific pages for different types of users (registered and unregistered). I think the current ones are just serving registered users, so maybe we should put there permissions.EDIT_DOC or VIEW_DOCS?

ahelmberger: This is an error. You can always reach the help page from the footer link regardless if you are logged in or out.

laurentiu-ilici: Oh, okay :).

laurentiu-ilici (approved):

PR #119, opened 2021-12-28
EDU-208 room details part 2

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-208


src/components/pages/room.js
```diff
@@ -1,14 +1,59 @@
...
+import { Row, Space, List, Collapse, Button } from 'antd';
+import { invitationShape, roomShape } from '../../ui/default-prop-types.js';
+import RoomInvitationCreationModal from '../room-invitation-creation-modal.js';
 
 export default function Room({ PageTemplate, initialState }) {
   const { t } = useTranslation('room');
   const { formatDate } = useDateFormat();
-  const { roomDetails } = initialState;
+  const { roomDetails, invitations } = initialState;
+  const [isInvitationModalOpen, setIsRoomInvitationModalOpen] = useState(false);
```
adrianaluchian, line 13: Usually with `useState` the naming pattern is `state`, `setState`.
In this case the pattern is broken.

laurentiu-ilici: I would like to correct this pattern.

In previous projects I was using the state to set different state related fields (usually our states are quite small).
For example here if we would have a second thing that we would need on the state we would have:
const [secondThing, setSecondThing] = useState(...);
This allows us to setSecondThing without doing the
setState(prevState => ({...preveState, secondThing: {}));

This way of working is also in tune with what the documentation of useState says:

What do we pass to useState as an argument? The only argument to the useState() Hook is the initial state. Unlike with classes, the state doesn’t have to be an object. We can keep a number or a string if that’s all we need. In our example, we just want a number for how many times the user clicked, so pass 0 as initial state for our variable. **(If we wanted to store two different values in state, we would call useState() twice.)**

[Link to the documentation](https://reactjs.org/docs/hooks-state.html)

What do you guys thing?

laurentiu-ilici: If you agree, I would also change the pattern on the room invitation modal and we can change the rest as we go.

adrianaluchian: Ok, sorry, I see you took my remark quite literally, as I used the documentation terminology. I didn't literally mean `state`, just the actual pattern of `something` and `setSomething`.
And in your code this is not correct.

adrianaluchian: Regarding splitting up a state in multiple states, I would not like to be black or white on it, as we should use our best judgement depending on case. In some cases, the components should stay together (even if they are not always set at the same time) because, due to the async manner of the state being set, it is a way to guarantee state consistency between events.

src/components/room-invitation-creation-modal.js
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <Modal
+      title={t('newRoomInvitation')}
+      onCancel={handleCancel}
+      onOk={submitForm}
+      maskClosable={false}
+      visible={isVisible}
+      okButtonProps={{ loading: loading.loading, htmlType: 'submit' }}
+      >
+      <Form name="new-room-invitation-form" initialValues={initialFormValues} onFinish={onFinish} ref={formRef} layout="vertical">
```
laurentiu-ilici, line 63: I've gone for a programatic way to use the form here, if you like the approach this could be done for other forms that are used within modals. This has the advantage of getting rid of handling the state and form validation outside the form.

src/components/pages/room.yml
```diff
@@ -7,3 +7,18 @@ memberUsername:
...
+  de: Einladung läuft ab am
+createInvitationButton:
+  en: Create an invitation
+  de: Einladung erstellen
+roomMembersHeader:
+  en: Room members ({count})
+  de: Raummitglieder ({count})
+invitationsHeader:
+  en: Outstanding invitations ({count})
+  de: Herausragende Einladungen ({count})
```
ahelmberger, line 24: Ausstehende

src/components/room-invitation-creation-modal.yml
```diff
@@ -0,0 +1,6 @@
+newRoomInvitation:
+  en: New invitation
+  de: Neuer Einladung
+emailRequired:
+  en: Please specify an Email
+  de: Bitte geben Sie einen E-Mail an
```
ahelmberger, line 6: eine E-Mail-Adresse

src/components/room-invitation-creation-modal.yml
```diff
@@ -0,0 +1,6 @@
+newRoomInvitation:
+  en: New invitation
+  de: Neuer Einladung
+emailRequired:
+  en: Please specify an Email
```
ahelmberger, line 5: an email address

src/components/room-invitation-creation-modal.yml
```diff
@@ -0,0 +1,6 @@
+newRoomInvitation:
+  en: New invitation
+  de: Neuer Einladung
```
ahelmberger, line 3: Neue Einladung

src/server/room-controller.js
```diff
@@ -86,6 +73,34 @@ export default class RoomController {
...
+      }
+
+      if (room.owner === userId) {
+        invitations = await this.roomService.getRoomInvitations(roomId);
+      }
+    }
+
+    const roomDetails = await this.clientDataMapper.mapRoom(room);
+
+    return this.pageRenderer.sendPage(req, res, PAGE_NAME.room, { roomDetails, invitations });
```
ahelmberger, line 101: Can we just call it `room`? I don't think that name is already taken by a different data structure.

src/server/room-controller.spec.js
```diff
@@ -158,42 +159,139 @@ describe('room-controller', () => {
...
+        it('should not call getRoomInvitations', () => {
+          sinon.assert.notCalled(roomService.getRoomInvitations);
+        });
+
+        it('should call pageRenderer with the right parameters', () => {
+          sinon.assert.calledWith(pageRenderer.sendPage, request, {}, PAGE_NAME.room, { roomDetails: privateRoom, invitations: null });
+        });
+      });
+
+      describe('when some random person calls', () => {
```
ahelmberger, line 247: `random` could be also a room member. To make it clear I'd say `when the user is not authorized for room access` or similar.

src/server/room-controller.spec.js
```diff
@@ -158,42 +159,139 @@ describe('room-controller', () => {
...
       it('should call pageRenderer with the right parameters', () => {
-        sinon.assert.calledWith(pageRenderer.sendPage, request, {}, PAGE_NAME.room, { roomDetails: room });
+        sinon.assert.calledWith(pageRenderer.sendPage, request, {}, PAGE_NAME.room, { roomDetails: privateRoom, invitations: null });
       });
     });
 
     describe('when the room does not exist', () => {
       it('should throw a not found exception', () => {
-        expect(() => sut.handleGetRoomPage({ params: { roomId: 'abc' } }).rejects.toThrow(NotFound));
+        expect(() => sut.handleGetRoomPagePage({ params: { roomId: 'abc' } }).rejects.toThrow(NotFound));
```
ahelmberger, line 294: PagePage?

src/server/room-controller.js
adrianaluchian: Wanted to suggest the same actually.

laurentiu-ilici: I wanted to keep a distinction between the room without any other data aggregated to it (as it comes from the db) and a room that had more data loaded.
Changed.

src/components/room-invitation-creation-modal.yml
laurentiu-ilici: Fixed

src/components/room-invitation-creation-modal.yml
laurentiu-ilici: Fixed

src/components/room-invitation-creation-modal.yml
laurentiu-ilici: Fixed, I left the "an", not sure if that's okay

src/components/pages/room.yml
laurentiu-ilici: Fixed

src/server/room-controller.spec.js
laurentiu-ilici: Fixed

src/server/room-controller.spec.js
laurentiu-ilici: Yeah... I'm not sure how that happened... Fixed, thanks!

src/components/pages/room.js
```diff
@@ -1,47 +1,99 @@
...
 export default function Room({ PageTemplate, initialState }) {
   const { t } = useTranslation('room');
   const { formatDate } = useDateFormat();
-  const { roomDetails } = initialState;
+  const { room, invitations } = initialState;
+  const [isRoomInvitationModalOpen, setIsRoomInvitationModalOpen] = useState(false);
+
+  const handleOpenInvitationModalClick = event => {
+    setIsRoomInvitationModalOpen(true);
+    event.stopPropagation();
```
adrianaluchian, line 17: I think this might not be needed here, right? 🤔

src/components/pages/room.js
```diff
@@ -1,47 +1,99 @@
...
+    setIsRoomInvitationModalOpen(false);
+
+    if (!wasNewInvitationCreated) {
+      return;
+    }
+    window.location.reload();
+  };
+
+  const displayInvitations = () => (
+    <Collapse className="Room-invitationsCollapse">
```
adrianaluchian, line 30: This sounds like a movie, "The collapse of the room invitations - Part 1 🔥 " 😄

src/components/pages/room.less
```diff
@@ -0,0 +1,9 @@
+.Room {}
+
+.Room-membersCollapse {
+  margin-top: 16px;
+}
+
+.Room-invitationsCollapse {
```
adrianaluchian, line 7: You could make just one class ("Room-sectionCollapse" or something like this) and use for both.

src/components/room-invitation-creation-modal.js
```diff
@@ -0,0 +1,78 @@
...
+      message: t('emailRequired'),
+      whitespace: true
+    }
+  ];
+  const initialFormValues = { email: null };
+  const [loading, setIsLoading] = useState(false);
+
+  const handleFormFinish = async values => {
+    try {
+      setIsLoading(prevState => ({ ...prevState, loading: true }));
```
adrianaluchian, line 31: Probably a leftover, in all the `setIsLoading` you can just say `setIsLoading(true/false)`

src/server/room-controller.js
```diff
@@ -86,6 +73,34 @@ export default class RoomController {
...
+    if (!room) {
+      throw new NotFound();
+    }
+
+    const { _id: userId } = req.user;
+
+    let invitations = null;
+
+    if (room.access === ROOM_ACCESS_LEVEL.private) {
+      const isOwnerOrMember = await this.roomService.isRoomMemberOrOwner(roomId, userId);
```
adrianaluchian, line 89: Tiny thing, but I find it unnecessary to reverse the orders or "owner" and "member" (`isOwnerOrMember` vs `isRoomMemberOrOwner`).
Could we please streamline? I'd vote for renaming `isRoomOwnerOrMember` ✋

src/server/room-controller.js
```diff
@@ -86,6 +73,34 @@ export default class RoomController {
...
+    const { roomId } = req.params;
+    const room = await this.roomService.getRoomById(roomId);
+
+    if (!room) {
+      throw new NotFound();
+    }
+
+    const { _id: userId } = req.user;
+
+    let invitations = null;
```
adrianaluchian, line 86: Another small thing:
For me this would make more sense to be an empty array by default. Then the page can make `invitations` required and knows it always has to deal with an array and can check for empty array when it needs a special case.

src/server/room-controller.spec.js
```diff
@@ -158,36 +159,133 @@ describe('room-controller', () => {
...
+      roomService.getRoomById.callsFake(roomId => {
+        if (roomId === privateRoom.roomId) {
+          return Promise.resolve(privateRoom);
+        }
+        return Promise.resolve();
+      });
+    });
+
+    describe('when the private room exists', () => {
+      describe('when the room owner calls', () => {
```
adrianaluchian, line 176: Small hint: this would read better in the test report as
```
when the private room exists 
   and the room owner calls
   and a member calls
```

src/components/pages/room.js
laurentiu-ilici: It is, otherwise the collapse also collapses when you click the button.

adrianaluchian: Got it 👍

src/server/room-controller.spec.js
```diff
@@ -158,36 +159,133 @@ describe('room-controller', () => {
...
+          user: {
+            _id: 'owner'
+          }
+        };
+
+        const invitations = [{ email: 'lae@bucalae.com', sentOn: new Date() }];
+
+        beforeEach(async () => {
+          roomService.getRoomInvitations = sandbox.stub().resolves(invitations);
+          roomService.isRoomMemberOrOwner = sandbox.stub().resolves(true);
```
adrianaluchian, line 191: I see stubbing is now implemented in 2 ways in this file.
You could use the already existing pattern and place all `roomService` stubs together with the others in the out-most `beforeEach` and then stub the return as in the other usages.

Example:
```
roomService.getRoomInvitations.resolves(invitations);
```

src/server/room-controller.js
laurentiu-ilici: The invitations are not required by the room page. They are only loaded when you are the room owner, thus preventing members from ever seeing invitations. Here are the propTypes:
 invitations: PropTypes.arrayOf(invitationShape) //no isRequired added
I've put  room: roomShape.isRequired, that was forgotten and I think this caused the confusion, thanks!

src/components/pages/room.less
laurentiu-ilici: Done

src/components/room-invitation-creation-modal.js
laurentiu-ilici: Yes, thanks, I also renamed to isLoading so we have the pattern x setX

src/server/room-controller.spec.js
laurentiu-ilici: Done, thanks for the hint.

src/server/room-controller.js
adrianaluchian: Yes I know that they are, as data, currently not required, there was no confusion 👍 

What I said was that the page could simply always require an `invitations` array there and display what it gets (an empty array being, still, a valid array and fulfilling the requirement).
Because the room page itself does not care about the logic of room owner or not when it comes to invitations, it just displays. The business logic of populating `invitations` or not is strictly in the controller.

src/server/room-controller.spec.js
adrianaluchian: Cool 👍

src/server/room-controller.js
laurentiu-ilici: The UI is a presentation layer and it should present the data sent by the controller to the user. 
The controller decides if you deliver invitations or not based on the type of user (member or owner).
The UI just checks if there are invitations delivered or not and displays them if they are (it doesn't care about owner or member as you said).

If we want to deliver an empty array we need to deliver an extra bit of information (or make the UI smarter) to know whether it should display the invitations section or not. Maybe something like:
{ 
  isAllowedToViewInvitations: true/false,
  invitations: [...]
}

Would this be better?

src/server/room-controller.spec.js
laurentiu-ilici: Fixed, thanks.

src/components/pages/room.less
```diff
@@ -0,0 +1,5 @@
+.Room {}
```
adrianaluchian, line 1: This class needs to use `.m-content` mixin 
```
.Room {
   .m-content;
}
```
so that it has the same page styling as the others.

Example:
Screen Shot 2021-12-30 at 10 20 48
vs
Screen Shot 2021-12-30 at 10 20 56

src/server/room-controller.js
laurentiu-ilici: Fixed and renamed everything to isRoomOwnerOrMember, it sounds better indeed.

src/components/pages/room.less
laurentiu-ilici: hmm, will remember this, thanks!

src/server/room-controller.js
adrianaluchian: I think this complicates things. Why not simply have the controller always send an array and have the UI always display the collapsible invitations section, which in some cases will simply be empty. But this way the UI is also consistent and the logic is as simple as can be.

laurentiu-ilici: Currently the invitations section has also the send invite button, we'd need to hide that for members. Plus it displays an empty section that will always be empty for a normal member.

For me right now the UI is the simplest it can be: if there is data display it, if not, then don't.

adrianaluchian:
To me the current UI is pretty hard to read, with all the text crammed up together

Screen Shot 2021-12-30 at 10 40 45

How about something like this? I even think the translated labels are not needed because the reorganizing of the data within the rows (and having the consistency between the sections) makes it clear.

Screen Shot 2021-12-30 at 10 40 05


What do you guys think?

src/server/room-controller.js
adrianaluchian: Yeah it makes more sense, from order of interest/priority, Owner before Member.
Cool, thanks!

laurentiu-ilici:
> To me the current UI is pretty hard to read, with all the text crammed up together
> 
> Screen Shot 2021-12-30 at 10 40 45
> 
> How about something like this? I even think the translated labels ("Einladung gesendet am:" etc.. ) are not needed because the reorganizing of the data within the rows (and having the consistency between the sections) makes it clear.
> 
> Screen Shot 2021-12-30 at 10 40 05
> 
> What do you guys think?

I'm also not happy with the way it looks, I wanted to not block the other stories for too long and I wanted to try to avoid tables, this was a try out. I'm not sure how to design this better at this point. I could go for tables and be done with it until we want to get rid of tables. Would that be better?

src/server/room-controller.js
adrianaluchian: Ok then let's just leave this part as is now and not drag this thread longer 😬

laurentiu-ilici:
> > To me the current UI is pretty hard to read, with all the text crammed up together
> > Screen Shot 2021-12-30 at 10 40 45
> > How about something like this? I even think the translated labels ("Einladung gesendet am:" etc.. ) are not needed because the reorganizing of the data within the rows (and having the consistency between the sections) makes it clear.
> > Screen Shot 2021-12-30 at 10 40 05
> > What do you guys think?
> 
> I'm also not happy with the way it looks, I wanted to not block the other stories for too long and I wanted to try to avoid tables, this was a try out. I'm not sure how to design this better at this point. I could go for tables and be done with it until we want to get rid of tables. Would that be better?

Yeah sorry, I'll try to mimic your design.

src/server/room-controller.spec.js
adrianaluchian: Can you please fix **all** such cases that are not streamlined?
Meaning everything with `= sandbox.stub()` in the inner `beforeEach` blocks? It's the same case for all, you don't need re-stubbing, just setting the resolve.

src/components/room-invitation-creation-modal.yml
ahelmberger: Ah yes, we need the "an" :-)

src/server/room-controller.js
ahelmberger: We'll have to change this pattern anyway (probably), as there will be much more data in the room admin area (that is only visible to owners).

adrianaluchian:
> 
> I'm also not happy with the way it looks, I wanted to not block the other stories for too long and I wanted to try to avoid tables, this was a try out. I'm not sure how to design this better at this point. I could go for tables and be done with it until we want to get rid of tables. Would that be better?

Who said we need tables 🤷‍♀️ It's as simple as a flex container.

**room.js** for invitations:
```
            <List.Item class="Room-row">
              <Space>
                <span>{formatDate(invitation.sentOn)}</span>
                <Space>
                  <span className="Room-invitationEmail">{invitation.email}</span>
                </Space>
              </Space>

              <Space>
                <span>{t('expires')}:</span>
                <span>{formatDate(invitation.expires)}</span>
              </Space>
            </List.Item>
```

**room.less**
```
.Room-row {
  display: flex;
  justify-content: space-between;
}
```

For room members it's even as simple as:

```
              <List.Item>
                <Space>
                  <Space>
                    <span>{formatDate(member.joinedOn)}</span>
                  </Space>
                  <Space>
                    <span>{member.username}</span>
                  </Space>
                </Space>
              </List.Item>

```

src/server/room-controller.spec.js
laurentiu-ilici: Somehow I tried this before and it didn't work, now it does, I must have done something wrong back then, thanks.
I think it's fixed now.

adrianaluchian: You probably confused the methods, cause one was already stubbed in the outer most `beforeEach` and one not. And of course it would not work for the one that was not.

src/server/room-controller.js
laurentiu-ilici: I've changed it so that the room page decides what to show. I realized I needed also to not display this section for the public rooms (invitations make no sense there since everyone is a "member").

Sorry for the bid discussion, my head was not in the right place this morning.

laurentiu-ilici: Hmm, actually I will hide the members list as well, it makes no sense on the public rooms. We can revisit this later.

laurentiu-ilici:
> 

Fixed, looks much better, thanks!

src/components/pages/room.js
```diff
@@ -73,28 +93,8 @@ export default function Room({ PageTemplate, initialState }) {
...
-                    <span>{formatDate(member.joinedOn)}</span>
-                  </Space>
-                </Space>
-              </List.Item>)}
-            />
-        </Collapse.Panel>
-      </Collapse>
-      { shouldDisplayInivations && displayInvitations(invitations) }
+      { isPrivateRoom && displayMembers }
+      { shouldDisplayInivations && displayInvitations }
```
adrianaluchian, line 97: Small thing: typo "Inivations"

laurentiu-ilici: Hi, thanks! Solved

adrianaluchian (approved):

ahelmberger (approved):

PR #121, opened 2021-12-30
EDU-189 Add document validation upon saving to DB

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-189

As we have several places that call `documentStore.save` I opted for placing the validation in the store.


src/domain/schemas/document-schemas.js
```diff
@@ -32,7 +32,7 @@ export const createRevisionBodySchema = joi.object({
   language: joi.string().case('lower').required(),
   sections: joi.array().items(sectionSchema).required(),
   appendTo: documentRevisionAppendToSchema.optional(),
-  tags: joi.array(),
+  tags: joi.array().items(joi.string()).required(),
```
ahelmberger, line 35: :+1:

src/domain/schemas/document-schemas.js
```diff
@@ -52,3 +52,55 @@ export const hardDeleteSectionBodySchema = joi.object({
 export const hardDeleteDocumentBodySchema = joi.object({
   documentKey: idOrKeySchema.required()
 });
+
+const sectionDBSchema = joi.object({
+  revision: joi.string().required(),
```
ahelmberger, line 57: You can use the `idOrKeySchema` also here

src/domain/schemas/document-schemas.js
```diff
@@ -52,3 +52,55 @@ export const hardDeleteSectionBodySchema = joi.object({
...
+  _id: idOrKeySchema.required(),
+  key: idOrKeySchema.required(),
+  order: joi.number().required(),
+  createdOn: joi.date().required(),
+  createdBy: idOrKeySchema.required(),
+  title: joi.string().required(),
+  slug: slugSchema,
+  language: joi.string().case('lower').required(),
+  sections: joi.array().items(sectionDBSchema).required(),
+  restoredFrom: joi.string().allow(null).allow('').required(),
```
ahelmberger, line 79: Is there a case where we have empty string? I think it should be either `null` or not-empty :thinking:

src/domain/schemas/document-schemas.js
```diff
@@ -52,3 +52,55 @@ export const hardDeleteSectionBodySchema = joi.object({
...
+  updatedBy: idOrKeySchema.required(),
+  title: joi.string().required(),
+  slug: slugSchema,
+  language: joi.string().case('lower').required(),
+  sections: joi.array().items(sectionDBSchema).required(),
+  contributors: joi.array().items(joi.string()).required(),
+  tags: joi.array().items(joi.string()).required(),
+  archived: joi.bool().required(),
+  origin: joi.string().required(),
+  originUrl: joi.string().allow(null).allow(''),
```
ahelmberger, line 104: Here I also think we do not need the empty string

adrianaluchian:
I shall fix all tests once I make all schema corrections base on your comments here.

ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-189
> 
> As we have several places that call `documentStore.save` I opted for placing the validation in the store.

Even if that's a new "concept" you're introducing, it's probably the best place. At least I can't think of a better way. We should do similar for other "data integrity checks" in the future then.

src/domain/schemas/document-schemas.js
adrianaluchian: Double check I shall. There might be cases in the DB already that will cause issues on import.

ahelmberger:
> I shall fix all tests once I make all schema corrections base on your comments here.

I'm done with suggestions :-)

src/domain/schemas/document-schemas.js
laurentiu-ilici: Cool, we can also put a min(3) and max(30) so it matches more closely to the FE validation on tags 
.items(joi.string().min(3).max(30))

laurentiu-ilici:
> > Closes https://educandu.atlassian.net/browse/EDU-189
> > As we have several places that call `documentStore.save` I opted for placing the validation in the store.
> 
> Even if that's a new "concept" you're introducing, it's probably the best place. At least I can't think of a better way. We should do similar for other "data integrity checks" in the future then.

" We should do similar for other "data integrity checks" in the future then."
Totally agreed and we need to think about how to keep these checks consistent between the front-end and the backend (some are not at the moment, e.g. tags).

src/domain/schemas/document-schemas.js
adrianaluchian: True, used!

src/domain/schemas/document-schemas.js
adrianaluchian: To me client/server side validation should insure **business** validation, the store should insure **data/model consistency**. For this reason I would not add this extra restriction.

src/domain/schemas/document-schemas.js
adrianaluchian: Both `restoredFrom` and `originUrl` currently have in the DB empty string and `null` values.
So we either accept both, or we need to migrate both.
Due to the import feature, is not enough to ensure new documents consistently default these properties to `null`.

src/domain/schemas/document-schemas.js
adrianaluchian: See https://github.com/educandu/educandu/pull/121#discussion_r777383608

src/domain/schemas/document-schemas.js
adrianaluchian: To be honest, I would not do a migration at this point, though I would like the consistency I think we can live with this slight extra tolerance.

src/domain/schemas/document-schemas.js
laurentiu-ilici: Ok

src/domain/schemas/document-schemas.js
ahelmberger: Ok. If it's already inconsistent, we can fix it also later.

ahelmberger (approved):

PR #123, opened 2022-01-03
EDU-222 repair image tiles

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-222

We will have to recreate the documents after running this script otherwise we will see no benefits.


migrations/manual/educandu-2021-12-29-01-repair-image-tiles.js
```diff
@@ -0,0 +1,62 @@
...
+      } else if (!doc) {
+        console.log('Could not find document for slug: ', documentLinkInfo.url);
+        imageTilesSections[0].content.tiles[documentLinkInfo.tileIndex].link.url = '';
+      } else {
+        imageTilesSections[0].content.tiles[documentLinkInfo.tileIndex].link.url = doc.key;
+      }
+
+      // eslint-disable-next-line no-await-in-loop
+      await this.db.collection('documentRevisions').replaceOne({ _id: revision._id }, revision);
+    }
```
ahelmberger, line 44: As we don't have a way to trigger re-creation of documents yet, maybe it would be the fastest way to extract the content of the `for` above into a function that can deal with an individual document revision or also a document and then also update the corresponding document in this migration?

laurentiu-ilici: Ok, will do, I thought we already have it somehow and renounced the docs part when I did it.

laurentiu-ilici: Added and tested with the prod version from 20-12-2021 it seems to work fine.

migrations/manual/educandu-2021-12-29-01-repair-image-tiles.js
```diff
@@ -0,0 +1,62 @@
...
+
+      if (imageTilesSections[0].content.tiles.length < documentLinkInfo.tileIndex + 1) {
+        console.log(documentLinkInfo);
+        console.log('Number of tile sections', imageTilesSections[0].content.tiles.length, 'tile index: ', documentLinkInfo.tileIndex, 'revision', revision._id);
+      }
+      // eslint-disable-next-line no-await-in-loop
+      const doc = await this.db.collection('documents').findOne({ slug: documentLinkInfo.url });
+
+      if (imageTilesSections[0].content.tiles[documentLinkInfo.tileIndex]?.link.type !== 'internal') {
+        console.log(`Document info tile is not of expected type, but of type ${imageTilesSections[0].content.tiles[documentLinkInfo.tileIndex].link.type}`);
```
adrianaluchian, line 35: Maybe better to extract `imageTilesSections[0].content.tiles[documentLinkInfo.tileIndex]?.link.type` in a variable, because in the second usage on line 33 the optional chaining operator is missing, but the `if` block could be entered with the condition evaluating to `if  (null !== 'internal')`

migrations/manual/educandu-2021-12-29-01-repair-image-tiles.js
ahelmberger: Looks good

migrations/manual/educandu-2021-12-29-01-repair-image-tiles.js
laurentiu-ilici: Done, thanks! I also extracted some of the other long names to a variable for better readability.

ahelmberger (approved):

PR #128, opened 2022-01-04
EDU-232 Add lessons collection

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-232

### DIFF
--- a/src/domain/schemas/document-schemas.js
+++ b/src/domain/schemas/document-schemas.js
@@ -1,6 +1,5 @@
 import joi from 'joi';
-import { idOrKeySchema } from './shared-schemas.js';
-import inputValidators from '../../utils/input-validators.js';
+import { idOrKeySchema, slugSchema, sectionDBSchema } from './shared-schemas.js';
 
 const sectionSchema = joi.object({
   key: idOrKeySchema.required(),
@@ -23,8 +22,6 @@ export const getSearchDocumentsByTagsSchema = joi.object({
 export const getRevisionsByKeyQuerySchema = joi.object({
   key: idOrKeySchema.required()
 });
-
-export const slugSchema = joi.string().pattern(inputValidators.slugValidationPattern).allow('').required();
 
 export const createRevisionBodySchema = joi.object({
   title: joi.string().required(),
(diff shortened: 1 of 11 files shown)


src/domain/schemas/document-schemas.js
```diff
@@ -53,7 +53,7 @@ export const hardDeleteDocumentBodySchema = joi.object({
   documentKey: idOrKeySchema.required()
 });
 
-const sectionDBSchema = joi.object({
+export const sectionDBSchema = joi.object({
```
laurentiu-ilici, line 56: We probably should not use this, but for now it's okay. It has some properties which I think are debatable if we need such as "deleteBecause"

src/domain/schemas/lesson-schemas.js
```diff
@@ -0,0 +1,16 @@
+import joi from 'joi';
+import { idOrKeySchema } from './shared-schemas.js';
+import { slugSchema, sectionDBSchema } from './document-schemas.js';
```
ahelmberger, line 3: We should move those two schemas into `shared-schemas.js` as they are used in multiple places

src/domain/schemas/document-schemas.js
ahelmberger: True, that's a good question. See also my other comment regarding this.

src/domain/schemas/lesson-schemas.js
laurentiu-ilici: Done, thanks!

ahelmberger: When you move the `slugSchema` from `document-schemas.js` to `shared-schemas.js`, you also have to move the tests for `slugSchema` from `document-schemas.spec.js` to `shared-schemas.spec.js`.

laurentiu-ilici: True, moved and added some section schema db tests.

ahelmberger: Cool, thanks

ahelmberger (approved):

PR #129, opened 2022-01-04
EDU-230 Remove default document from landing page

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-230

Due to the fact that `global-alerts.js` was exposing both the helper method `getGlobalAlerts` and the hook `useGlobalAlerts` and due to the need of the extra dependency on `settings`, I decided to unify in the direction of using the hook, reason why I changed all pages that did not use the hook into functional components.

**Update**
Changed logic to display this alert banner on fullscreen pages as well, thus displaying it on all pages.


laurentiu-ilici:
I don't see the banner on the home page, it seems to work on other pages, is this intentional? The story says "all pages"

adrianaluchian:
> I don't see the banner on the home page, it seems to work on other pages, is this intentional? The story says "all pages"
There are no banners on the home page of any kind, all our banners are on all other pages.

ahelmberger:
> > I don't see the banner on the home page, it seems to work on other pages, is this intentional? The story says "all pages"
> > There are no banners on the home page of any kind, all our banners are on all other pages.

It's just important to not forget to make the home page show the banner, because that's the most important place we want to have it. We have to adjust the page template in the consuming systems (maybe it would make sense to also change it in the test app in this repo?

adrianaluchian:
> > > I don't see the banner on the home page, it seems to work on other pages, is this intentional? The story says "all pages"
> > > There are no banners on the home page of any kind, all our banners are on all other pages.
> 
> It's just important to not forget to make the home page show the banner, because that's the most important place we want to have it. We have to adjust the page template in the consuming systems (maybe it would make sense to also change it in the test app in this repo?

Ok my understanding was that we need to add this banner the same way as the other global banner 😬 
In this case I will decline the PR, make the changes and reopen.

adrianaluchian:
Closing due to pending changes.

laurentiu-ilici:
> > > I don't see the banner on the home page, it seems to work on other pages, is this intentional? The story says "all pages"
> > > There are no banners on the home page of any kind, all our banners are on all other pages.
> 
> It's just important to not forget to make the home page show the banner, because that's the most important place we want to have it. We have to adjust the page template in the consuming systems (maybe it would make sense to also change it in the test app in this repo?

Okay, then there are 3 others (which probably don't make a lot of sense to have the banner),  but I'll ask: login page, registration form page and password forgot.

ahelmberger:
> > > > I don't see the banner on the home page, it seems to work on other pages, is this intentional? The story says "all pages"
> > > > There are no banners on the home page of any kind, all our banners are on all other pages.
> > 
> > 
> > It's just important to not forget to make the home page show the banner, because that's the most important place we want to have it. We have to adjust the page template in the consuming systems (maybe it would make sense to also change it in the test app in this repo?
> 
> Okay, then there are 3 others (which probably don't make a lot of sense to have the banner), but I'll ask: login page, registration form page and password forgot.

To be honest, I would just show it on **all** pages, basta :-)

src/components/default-page-template.js
```diff
@@ -137,6 +137,25 @@ function DefaultPageTemplate({ children, fullScreen, headerActions, alerts }) {
     }
   ].filter(item => item.showWhen);
 
+  const renderAlert = alert => {
+    const shouldRenderAlert = !fullScreen || alert.showInFullScreen;
```
ahelmberger, line 141: Nice :+1:

ahelmberger (approved):

PR #133, opened 2022-01-05
EDU-216 Add cache busting everywhere

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-216

I'm not so sure this is exactly what we need though since it can also be an annoying feature. 
Please try the following:
 Go to the rooms section
 Make a  new room
 Click the back button
 What I would like to see is the updated list of rooms
 What you see is a flicker with this and the page going back to the first tab

Most pages will be okay with this feature, but some might become bad for the user experience

### DIFF
--- a/src/components/pages/import-batch-creation.js
+++ b/src/components/pages/import-batch-creation.js
@@ -7,12 +7,10 @@ import { useService } from '../container-context.js';
 import { handleApiError } from '../../ui/error-helper.js';
 import ClientConfig from '../../bootstrap/client-config.js';
 import { useGlobalAlerts } from '../../ui/global-alerts.js';
-import { useReloadPersistedWindow } from '../../ui/hooks.js';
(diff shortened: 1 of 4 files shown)


ahelmberger:
> Closes: https://educandu.atlassian.net/browse/EDU-216
> 
> I'm not so sure this is exactly what we need though since it can also be an annoying feature. Please try the following: Go to the rooms section Make a new room Click the back button What I would like to see is the updated list of rooms What you see is a flicker with this and the page going back to the first tab
> 
> Most pages will be okay with this feature, but some might become bad for the user experience

Yeah, the tabs are a problem with this, but in my opinion I would rather (in the future) add a URL query param on tab change, so each tab has its own URL, so we can "deeplink". But we have to discuss this. For now, I think we can live with this flicker.

ahelmberger (approved):

PR #134, opened 2022-01-05
EDU-237 Fix validation error

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-237

### DIFF
--- a/src/services/document-service.js
+++ b/src/services/document-service.js
@@ -364,7 +364,10 @@ class DocumentService {
           // If not changed, re-use existing revision:
           if (deepEqual(ancestorSection.content, section.content)) {
             logger.info(`Section has not changed compared to ancestor section with revision ${ancestorSection.revision}, using the existing`);
-            return cloneDeep(ancestorSection);
+            const clonedSection = cloneDeep(ancestorSection);
+            clonedSection.deletedOn = clonedSection.deletedOn && new Date(clonedSection.deletedOn);
+
+            return clonedSection;
           }
         }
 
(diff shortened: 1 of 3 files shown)


src/services/document-service.js
```diff
@@ -364,7 +364,10 @@ class DocumentService {
           // If not changed, re-use existing revision:
           if (deepEqual(ancestorSection.content, section.content)) {
             logger.info(`Section has not changed compared to ancestor section with revision ${ancestorSection.revision}, using the existing`);
-            return cloneDeep(ancestorSection);
+            const clonedSection = cloneDeep(ancestorSection);
```
adrianaluchian, line 367: Maybe it would be good if we looked into another cloning method or make our own, so that the types are persisted. Now dates become strings after cloning.

FYI, we could use a custom parsing method (`JSON.parse(value, customParsing)`), in which we could say:
- if a value is a string date, then cast it to `Date`
**but** it might be that in some case the original value was string and we want string in the clone too.

ahelmberger: Oh, I see. Luckily it's encapsulated, so we just have to change one place. Maybe something like https://www.npmjs.com/package/clone-deep would be a good choice?

ahelmberger: I'm fine with the workaround for now, but, as you said, let's fix this globally soon.

ahelmberger (approved):

src/services/document-service.js
laurentiu-ilici: I'm thinking that maybe we can use joi to somehow do the transform before checking/saving, I know yup had this option.

laurentiu-ilici (approved):

src/services/document-service.js
ahelmberger: Maybe, although I find this to sloppy. IMO this joi type coercion is a good feature to consolidate user input, but the application business logic should know what types it deals with at all times and throw if something is off.

PR #135, opened 2022-01-05
EDU-236 Use Lambda@Edge dev server locally

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-236

### DIFF
--- a/src/server/room-controller.js
+++ b/src/server/room-controller.js
@@ -103,14 +103,18 @@ export default class RoomController {
 
   async handleAuthorizeResourcesAccess(req, res) {
     const { roomId } = req.params;
-    const { _id: userId } = req.user;
+
+    const userId = req.user?._id || null;
+    if (!userId) {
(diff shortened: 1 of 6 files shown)


laurentiu-ilici (approved):

src/server/room-controller.js
```diff
@@ -103,14 +103,18 @@ export default class RoomController {
 
   async handleAuthorizeResourcesAccess(req, res) {
     const { roomId } = req.params;
-    const { _id: userId } = req.user;
+
+    const userId = req.user?._id || null;
```
adrianaluchian, line 107: Small thing, due to the early return and "falsy" check, `|| null` is not needed.

adrianaluchian (approved):

src/server/room-controller.js
ahelmberger: Fixed, thanks

PR #136, opened 2022-01-05
EDU-238 change clone deep

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-238

### DIFF
--- a/src/utils/clone-deep.js
+++ b/src/utils/clone-deep.js
@@ -15,7 +15,7 @@ function isDate(value) {
 }
 
 function isPrimitiveType(value) {
-  return ['string', 'number', 'boolean', 'undefined'].includes(typeof value);
+  return ['undefined', 'string', 'number', 'bigint', 'boolean'].includes(typeof value);
 }
 
 function cloneDeep(value) {
--- a/src/utils/clone-deep.spec.js
+++ b/src/utils/clone-deep.spec.js
@@ -3,14 +3,15 @@ import cloneDeep from './clone-deep.js';
 describe('clone-deep', () => {
 
   const original = {
-    stringProp: 'a',
-    numberProp: 1,
     nullProp: null,
     // eslint-disable-next-line no-undefined
     undefinedProp: undefined,
+    stringProp: 'a',
+    numberProp: 1,
+    bigintProp: 10n,
     booleanProp: true,
-    dateAsStringProp: new Date().toISOString(),
     dateProp: new Date(),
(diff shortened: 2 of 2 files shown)


src/utils/clone-deep.js
```diff
@@ -1,5 +1,44 @@
-function cloneDeep(obj) {
-  return JSON.parse(JSON.stringify(obj));
+function isObject(value) {
+  return Object.prototype.toString.call(value) === '[object Object]';
+}
+
+function isArray(value) {
+  return value instanceof Array;
```
ahelmberger, line 6: Better: `Array.isArray(value)`

adrianaluchian: Yes of course, thank you, changed.

Would have been nice to have this on `Date` as well 😊

src/utils/clone-deep.js
```diff
@@ -1,5 +1,44 @@
...
+  }
+
+  if (isObject(value)) {
+    return Object.entries(value).reduce((clone, entry) => {
+      clone[entry[0]] = cloneDeep(entry[1]);
+      return clone;
+    }, {});
+  }
+
+  return value;
```
ahelmberger, line 41: It doesn't work with `BigInt` and `Symbol` types but we don't use them. Should we add `BigInt`, just in case (I think there will never be any symbols in pure data structures)?

adrianaluchian: I considered `Symbol` but also thought it's not the case for us here.
I did not think of `BigInt` since I never used it before, but I can surely add it.

ahelmberger: It won't hurt

ahelmberger (approved):

src/utils/clone-deep.js
adrianaluchian: Done!
Actually we could simplify the method to the following, since the check for `isObject` is strict and would not include `null`, `Date`, `Array` (so no danger if the `if` statements would be switched around):

```
function cloneDeep(value) {
  if (isDate(value)) {
    return new Date(value.getTime());
  }

  if (isArray(value)) {
    return value.map(item => cloneDeep(item));
  }

  if (isObject(value)) {
    return Object.entries(value).reduce((clone, entry) => {
      clone[entry[0]] = cloneDeep(entry[1]);
      return clone;
    }, {});
  }

  return value;
}
```

Or would we rather specifically see the `isPrimitiveType` and `isFunction` checks?
Thoughts?

PR #137, opened 2022-01-05
EDU-209 prevent some room membership contradictions

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-209

### DIFF
--- a/src/services/room-service.js
+++ b/src/services/room-service.js
@@ -4,6 +4,7 @@ import Logger from '../common/logger.js';
 import UserService from './user-service.js';
 import uniqueId from '../utils/unique-id.js';
 import RoomStore from '../stores/room-store.js';
+import RoomLockStore from '../stores/room-lock-store.js';
 import { ROOM_ACCESS_LEVEL } from '../domain/constants.js';
 import TransactionRunner from '../stores/transaction-runner.js';
 import RoomInvitationStore from '../stores/room-invitation-store.js';
@@ -22,10 +23,11 @@ const roomInvitationProjection = {
 };
 
 export default class RoomService {
-  static get inject() { return [RoomStore, RoomInvitationStore, UserService, TransactionRunner]; }
+  static get inject() { return [RoomStore, RoomLockStore, RoomInvitationStore, UserService, TransactionRunner]; }
 
-  constructor(roomStore, roomInvitationStore, userService, transactionRunner) {
+  constructor(roomStore, roomLockStore, roomInvitationStore, userService, transactionRunner) {
     this.roomStore = roomStore;
+    this.roomLockStore = roomLockStore;
     this.roomInvitationStore = roomInvitationStore;
     this.userService = userService;
     this.transactionRunner = transactionRunner;
@@ -144,13 +146,28 @@ export default class RoomService {
         joinedOn: new Date()
       };
 
-      await this.roomStore.updateOne(
-        { _id: invitation.roomId },
-        { $push: { members: newMember } },
-        { session }
-      );
+      let lock;
 
-      await this.roomInvitationStore.deleteOne({ _id: invitation._id }, { session });
+      try {
+        try {
+          lock = await this.roomLockStore.takeLock(invitation.roomId);
+        } catch {
+          throw new BadRequest();
+        }
+
+        const room = await this.roomStore.findOne({ '_id': invitation.roomId, 'members.userId': newMember.userId });
+        if (!room) {
+          await this.roomStore.updateOne(
+            { _id: invitation.roomId },
+            { $push: { members: newMember } },
+            { session }
+          );
+        }
+
+        await this.roomInvitationStore.deleteOne({ _id: invitation._id }, { session });
+      } finally {
+        this.roomLockStore.releaseLock(lock);
+      }
     });
   }
 
--- a/src/services/room-service.js
+++ b/src/services/room-service.js
@@ -80,7 +80,7 @@ export default class RoomService {
     return room;
   }
 
-  async createOrUpdateInvitation({ roomId, email, user }) {
+  async createOrUpdateInvitationIfNotOwner({ roomId, email, user }) {
     const now = new Date();
     const lowerCasedEmail = email.toLowerCase();
 
@@ -90,6 +90,10 @@ export default class RoomService {
     }
 
     const owner = await this.userService.getUserById(room.owner);
+
+    if (owner.email.toLowerCase() === lowerCasedEmail) {
+      return { room, owner, invitation: null };
+    }
 
     let invitation = await this.roomInvitationStore.findOne({ roomId, email: lowerCasedEmail });
     if (!invitation) {
--- a/src/services/room-service.js
+++ b/src/services/room-service.js
@@ -80,7 +80,7 @@ export default class RoomService {
     return room;
   }
 
-  async createOrUpdateInvitationIfNotOwner({ roomId, email, user }) {
+  async createOrUpdateInvitation({ roomId, email, user }) {
     const now = new Date();
     const lowerCasedEmail = email.toLowerCase();
 
@@ -92,7 +92,7 @@ export default class RoomService {
     const owner = await this.userService.getUserById(room.owner);
 
     if (owner.email.toLowerCase() === lowerCasedEmail) {
-      return { room, owner, invitation: null };
+      throw new BadRequest('The owner may not invite herself as a member');
     }
 
     let invitation = await this.roomInvitationStore.findOne({ roomId, email: lowerCasedEmail });
--- a/src/services/room-service.js
+++ b/src/services/room-service.js
@@ -94,7 +94,7 @@ export default class RoomService {
     const owner = await this.userService.getUserById(room.owner);
 
     if (owner.email.toLowerCase() === lowerCasedEmail) {
-      throw new BadRequest('The owner may not invite herself as a member');
+      throw new BadRequest('Invited user is the same as room owner');
     }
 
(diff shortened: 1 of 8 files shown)


src/services/room-service.js
```diff
@@ -140,13 +146,28 @@ export default class RoomService {
...
+
+      try {
+        try {
+          lock = await this.roomLockStore.takeLock(invitation.roomId);
+        } catch {
+          throw new BadRequest();
+        }
+
+        const room = await this.roomStore.findOne({ '_id': invitation.roomId, 'members.userId': newMember.userId });
+        if (!room) {
```
adrianaluchian, line 159: `if (!room) {` can create misunderstanding here, because the room itself dues actually exist.

I would maybe call the variable `roomContainingNewMember` to make it a bit clearer.

src/services/room-service.spec.js
```diff
@@ -167,17 +173,26 @@ describe('room-service', () => {
     let myPrivateRoom = null;
 
     beforeEach(async () => {
+      userService.getUserById = sandbox.stub().resolves(myUser);
```
adrianaluchian, line 176: A better practice is to let the sandbox do the stubbing instead of replacing the function implementation.

`sandbox.stub(userService, 'getUserById').resolves(myUser);`

It's also how we do it in all other tests.

src/services/room-service.spec.js
```diff
@@ -276,6 +302,33 @@ describe('room-service', () => {
...
+          await sut.confirmInvitation({ token: invitation.token, user: otherUser });
+        });
+
+        it('should not add the user user a second time', async () => {
+          const roomFromDb = await db.rooms.findOne({ _id: testRoom._id });
+          expect(roomFromDb.members).toHaveLength(1);
+          expect(roomFromDb.members[0]).toEqual({
+            userId: otherUser._id,
+            joinedOn: exisingMemberJoinDate
+          });
```
adrianaluchian, line 324: You can merge the expectations into one:

```
expect(roomFromDb.members).toEqual([{
    userId: otherUser._id,
    joinedOn: exisingMemberJoinDate
}]);
```

src/services/room-service.spec.js
```diff
@@ -276,6 +302,33 @@ describe('room-service', () => {
...
       });
+
+      describe('and the user gets invited a second time', () => {
+        let exisingMemberJoinDate;
+
+        beforeEach(async () => {
+          ({ invitation } = await sut.createOrUpdateInvitation({ roomId: testRoom._id, email: otherUser.email, user: myUser }));
+
+          const roomFromDb = await db.rooms.findOne({ _id: testRoom._id });
+          exisingMemberJoinDate = roomFromDb.members[0].joinedOn;
```
adrianaluchian, line 313: Typo in here.

Also I find it simpler to not stray with namings in too many directions, makes code harder to follow.
I would suggest `existingMemberJoinedOn`.

src/services/room-service.js
```diff
@@ -91,6 +93,10 @@ export default class RoomService {
 
     const owner = await this.userService.getUserById(room.owner);
 
+    if (owner.email.toLowerCase() === lowerCasedEmail) {
+      throw new BadRequest('The owner may not invite herself as a member');
```
adrianaluchian, line 97: I guess if the owner is male, they also cannon invite themselves 😅 

How about
`throw new BadRequest('Invited room member is the same as room owner');`
?

src/services/room-service.spec.js
```diff
@@ -167,17 +173,26 @@ describe('room-service', () => {
...
+    afterEach(() => {
+      sandbox.restore();
+    });
+
     it('should create a new invitation if it does not exist', async () => {
       const { invitation } = await sut.createOrUpdateInvitation({ roomId: myPrivateRoom._id, email: 'invited-user@test.com', user: myUser });
       expect(invitation.token).toBeDefined();
     });
 
+    it('should throw a bad request if the owner invites herself', () => {
```
adrianaluchian, line 192: `... if owner invites themselves`

laurentiu-ilici: But themselves is plural. Herself is the new polite way to treat someone when you don't know the gender.

src/services/room-service.spec.js
laurentiu-ilici: Fixed

src/services/room-service.spec.js
laurentiu-ilici: Adjusted both here and in older code.

src/stores/room-lock-store.js
```diff
@@ -0,0 +1,12 @@
+import Database from './database.js';
+import LockStoreBase from './lock-store-base.js';
+
+class RoomLockStore extends LockStoreBase {
+  static get inject() { return [Database]; }
+
+  constructor(db) {
+    super(db.batchLocks);
```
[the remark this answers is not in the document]
laurentiu-ilici: Thanks, good catch.

src/services/room-service.js
laurentiu-ilici: Fixed to the new message.

src/services/room-service.js
laurentiu-ilici: Renamed

src/services/room-service.spec.js
laurentiu-ilici: For some reason this doesn't work on this file. I've tried both here and on the room lock, any ideas?

src/services/room-service.spec.js
adrianaluchian: "Themselves" can be used in the singular non-gender specific case: 
"Both “themselves” and “themself” are acceptable as reflexive singular pronouns; however, “themselves” is currently the more common usage."
"Herself" is feminine, as "himself" is masculine. But anyway, I was not into discussing genders or political correctness, I just wanted to point out that I find the phrasing towards a gender confusing and unnecessary.

src/services/room-service.spec.js
adrianaluchian: I've done the investigation locally and found the causes.

**Case 1**
`sandbox.stub(userService, 'getUserById').resolves(myUser);` not working

This is because `setupTestUser` is called in `beforeAll`, but `pruneTestEnvironment` is called in `afterEach`.

What you need to do is:
- move both `setupTestUser` calls in a `beforeEach`
- remove `userService` and it's mocking altogether, as it is not needed (that's the whole point of `setupTestUser`)

**Case 2:**
`sandbox.stub(roomLockStore, 'takeLock').resolves(lock);` not working

The cause of this is that you did not add an equivalent `afterEach` for that particular test set, restoring the sandbox.

src/services/room-service.spec.js
laurentiu-ilici: Maybe it got updated in the meantime, when I was writing/reviewing research papers this was the rule, but it's been a few years. Anyway, I'll adjust.

src/services/room-service.spec.js
laurentiu-ilici: Fixed, thanks.

src/services/room-service.js
```diff
@@ -140,13 +146,24 @@ export default class RoomService {
...
-        { $push: { members: newMember } },
-        { session }
-      );
+      let lock;
+
+      try {
+        lock = await this.roomLockStore.takeLock(invitation.roomId);
 
-      await this.roomInvitationStore.deleteOne({ _id: invitation._id }, { session });
+        const roomContainingNewMember = await this.roomStore.findOne({ '_id': invitation.roomId, 'members.userId': newMember.userId });
```
ahelmberger, line 154: You have to put the session also here as a param in order to get the data that is valid for this transaction

src/services/room-service.spec.js
```diff
@@ -4,8 +4,10 @@ import RoomService from './room-service.js';
 import uniqueId from '../utils/unique-id.js';
 import Database from '../stores/database.js';
 import RoomStore from '../stores/room-store.js';
+import RoomLockStore from '../stores/room-lock-store.js';
 import { ROOM_ACCESS_LEVEL } from '../domain/constants.js';
 import { destroyTestEnvironment, setupTestEnvironment, pruneTestEnvironment, setupTestUser, createTestRoom } from '../test-helper.js';
+import { afterEach } from '@jest/globals';
```
ahelmberger, line 10: Not needed

laurentiu-ilici: Autoimport, I didn't see it sorry, I still haven't found a way to disable this one

src/services/room-service.spec.js
```diff
@@ -259,23 +272,71 @@ describe('room-service', () => {
...
 
+      afterEach(() => {
+        sandbox.restore();
+      });
+
+      it('should take a lock on the room', () => {
+        sinon.assert.calledWith(roomLockStore.takeLock, invitation.roomId);
+      });
+
+      it('should release thelock on the room', () => {
```
ahelmberger, line 292: Small typo, missing space

src/services/room-service.js
laurentiu-ilici: Hmmm, okay, I thought since it's a read it doesn't matter. Fixed, thanks!

src/services/room-service.spec.js
laurentiu-ilici: Fixed

src/services/room-service.spec.js
laurentiu-ilici: Fixed

adrianaluchian (approved):

ahelmberger (approved):

PR #138, opened 2022-01-06
EDU-231 Change room page layout

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-231

### DIFF
--- a/src/components/pages/room.js
+++ b/src/components/pages/room.js
@@ -86,16 +86,18 @@ export default function Room({ PageTemplate, initialState }) {
 
   return (
     <PageTemplate>
-      <h1> {t('pageNames:room', { roomName: room.name })}</h1>
-      <Row>
-        <Space>
-          <span>{t('ownerUsername')}:</span>
-          <span> {room.owner.username}</span>
-        </Space>
-      </Row>
-      { isPrivateRoom && displayMembers }
-      { shouldDisplayInvitations && displayInvitations }
-      <RoomInvitationCreationModal isVisible={isRoomInvitationModalOpen} onClose={handleInvitationModalClose} roomId={room._id} />
+      <div className="Room">
+        <h1> {t('pageNames:room', { roomName: room.name })}</h1>
+        <Row>
+          <Space>
+            <span>{t('ownerUsername')}:</span>
+            <span> {room.owner.username}</span>
+          </Space>
+        </Row>
+        { isPrivateRoom && displayMembers }
+        { shouldDisplayInvitations && displayInvitations }
+        <RoomInvitationCreationModal isVisible={isRoomInvitationModalOpen} onClose={handleInvitationModalClose} roomId={room._id} />
+      </div>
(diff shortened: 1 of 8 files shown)


src/components/pages/room.js
```diff
@@ -30,32 +34,35 @@ export default function Room({ PageTemplate, initialState }) {
     window.location.reload();
   };
 
-  const isPrivateRoom = room.access === ROOM_ACCESS_LEVEL.private;
-  const shouldDisplayInvitations = isPrivateRoom && isRoomOwner;
+  const renderLesson = lesson => {
+    const urlParts = ['lessons', encodeURIComponent(lesson._id)];
```
laurentiu-ilici, line 38: This can go to urls.js, see "getDocUrl", it already takes care of a missing slug. If the intention is to not put there a url that doesn't exist yet, you can still used the concatParts method I think.

src/server/room-controller.spec.js
```diff
@@ -221,8 +231,13 @@ describe('room-controller', () => {
           }
         };
 
+        const lessons = [];
+
         beforeEach(async () => {
           roomService.isRoomOwnerOrMember.resolves(true);
+          roomService.isRoomOwnerOrMember.resolves(true);
```
laurentiu-ilici, line 238: This just repeats the line above, typo I guess.

src/components/pages/room.js
```diff
@@ -30,32 +34,35 @@ export default function Room({ PageTemplate, initialState }) {
     window.location.reload();
```
laurentiu-ilici, line 34: This doesn't work so nicely anymore. After each invite, you get redirected to the lessons tab, Andreas had a suggestion that we deep link the tab key in the URL. I'm not sure if you want to go for that now.

src/components/pages/room.js
adrianaluchian: I actually intentionally didn't want to put it there yet, until we discuss with Uli and agree that we want a different page.

src/components/pages/room.js
adrianaluchian: I know we discussed that and we should definitely do it. I would not do that here though, but in a ticket that covers all such case.

src/server/room-controller.spec.js
adrianaluchian: Yup, leftover, removed, thanks!

laurentiu-ilici (approved):

src/components/pages/room.yml
```diff
@@ -13,3 +13,15 @@ roomMembersHeader:
 invitationsHeader:
   en: Outstanding invitations ({count})
   de: Ausstehende Einladungen ({count})
+lessonsTabTitle:
+  en: Lessons
+  de: Unterrichte
```
ahelmberger, line 18: Unterrichtseinheiten

ahelmberger (approved):

PR #139, opened 2022-01-06
EDU-199 Add room deletion

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-199

For testing I suggest that after going through the invitation process, you save your room collection and reintroduce it.

### DIFF
--- a/src/api-clients/room-api-client.js
+++ b/src/api-clients/room-api-client.js
@@ -40,7 +40,7 @@ class RoomApiClient {
   deleteRoom(roomId) {
     return this.httpClient
       .delete(
-        `/api/v1/rooms/${roomId}`,
+        `/api/v1/rooms/${encodeURIComponent(roomId)}`,
         { responseType: 'json' }
       )
       .then(res => res.data);
--- a/src/api-clients/room-api-client.js
+++ b/src/api-clients/room-api-client.js
@@ -36,6 +36,15 @@ class RoomApiClient {
       )
       .then(res => res.data);
   }
+
+  deleteRoom(roomId) {
+    return this.httpClient
+      .delete(
+        `/api/v1/rooms/${roomId}`,
+        { responseType: 'json' }
+      )
+      .then(res => res.data);
+  }
 }
 
 export default RoomApiClient;
--- a/src/components/confirmation-dialogs.yml
+++ b/src/components/confirmation-dialogs.yml
@@ -7,6 +7,9 @@ deleteSectionConfirmation:
 deleteDocumentConfirmation:
   en: Do you really want to delete document "{title}"?
   de: Möchten Sie das Dokument "{title}" wirklich löschen?
+deleteRoomConfirmation:
+  en: Do you really want to delete room "{roomName}"?
+  de: Möchten Sie der Raum "{roomName}" wirklich löschen?
 deleteCdnDocumentConfirmation:
   en: Do you really want to delete the CDN file "{fileName}"?
   de: Möchten Sie die Datei "{fileName}" wirklich vom CDN löschen?
--- a/src/components/confirmation-dialogs.yml
+++ b/src/components/confirmation-dialogs.yml
@@ -9,7 +9,7 @@ deleteDocumentConfirmation:
   de: Möchten Sie das Dokument "{title}" wirklich löschen?
 deleteRoomConfirmation:
   en: Do you really want to delete room "{roomName}"?
-  de: Möchten Sie der Raum "{roomName}" wirklich löschen?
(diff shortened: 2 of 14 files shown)


src/components/confirmation-dialogs.yml
```diff
@@ -7,6 +7,9 @@ deleteSectionConfirmation:
 deleteDocumentConfirmation:
   en: Do you really want to delete document "{title}"?
   de: Möchten Sie das Dokument "{title}" wirklich löschen?
+deleteRoomConfirmation:
+  en: Do you really want to delete room "{roomName}"?
+  de: Möchten Sie der Raum "{roomName}" wirklich löschen?
```
ahelmberger, line 12: ... de**n** Raum ...

src/server/room-controller.js
```diff
@@ -124,6 +144,12 @@ export default class RoomController {
       (req, res) => this.handlePostRoom(req, res)
     );
 
+    router.delete(
+      '/api/v1/rooms/:roomId',
+      [needsPermission(permissions.CREATE_ROOMS), validateParams(deleteRoomParamsSchema)],
```
ahelmberger, line 149: Should we maybe call this permission `CREATE_OR_DELETE_ROOMS`? :no_good:

laurentiu-ilici: I'm not sure actually why we need here create/edit/view since everyone can do it (as long as they have a user). Why not just MANAGE_ROOMS?

src/services/room-service.spec.js
```diff
@@ -331,4 +337,76 @@ describe('room-service', () => {
...
+          { name: cdnNameObject2 }
+        ]);
+
+        sandbox.stub(cdn, 'deleteObjects');
+        invitationDetails = await sut.createOrUpdateInvitation({ roomId, email: otherUser.email, user: myUser });
+
+        await sut.deleteRoom(roomId, myUser);
+      });
+
+      afterEach(() => {
```
ahelmberger, line 387: The sandbox is global, so this `afterEach` should better be global, otherwise the sandbox works differently in this `describe` block then elsewhere.

src/components/confirmation-dialogs.yml
laurentiu-ilici: Danke, fixed.

src/server/room-controller.js
ahelmberger: Good point, I'd be fine with that (or something like `OWN_ROOMS`)

laurentiu-ilici: Added a specific DELETE_ROOMS permission.

laurentiu-ilici: Fixed to have OWN_ROOMS permission. We will add specific rights when we need them for admins (for example to be able to view rooms that they don't own).

src/services/room-service.spec.js
laurentiu-ilici: Fixed

src/api-clients/room-api-client.js
```diff
@@ -36,6 +36,15 @@ class RoomApiClient {
       )
       .then(res => res.data);
   }
+
+  deleteRoom(roomId) {
+    return this.httpClient
+      .delete(
+        `/api/v1/rooms/${roomId}`,
```
adrianaluchian, line 43: Let's keep consistent and `encodeURIComponent(roomId)`, since we did this now in all api clients.

src/services/room-service.spec.js
```diff
@@ -391,4 +398,72 @@ describe('room-service', () => {
...
+      });
+    });
+
+    describe('when the room does not exist', () => {
+      it('should throw a not found exception', () => {
+        expect(() => sut.deleteRoom('badIdBadIdITellYou', myUser)).rejects.toThrow(NotFound);
+      });
+    });
+
+    describe('when another another user except the owner tries to delete', () => {
```
adrianaluchian, line 425: "another another" 😊

laurentiu-ilici: thanks thanks! :)

src/api-clients/room-api-client.js
laurentiu-ilici: Done

adrianaluchian (approved):

src/domain/permissions.js
```diff
@@ -65,9 +61,7 @@ export default {
   MANAGE_IMPORT,
   MANAGE_EXPORT,
   DELETE_CDN_FILE,
-  CREATE_ROOMS,
+  OWN_ROOMS,
```
ahelmberger, line 64: When you rename a constant, could you please do a full text search and rename everywhere. You missed one!

PR #140, opened 2022-01-07
EDU-240 Fix duplicate alerts on document pages

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-240

### DIFF
--- a/src/components/pages/doc.js
+++ b/src/components/pages/doc.js
@@ -1,12 +1,12 @@
 import PropTypes from 'prop-types';
 import DocView from '../doc-view.js';
 import urls from '../../utils/urls.js';
-import React, { useState } from 'react';
 import Restricted from '../restricted.js';
 import clipboardCopy from 'clipboard-copy';
 import Logger from '../../common/logger.js';
 import { Button, Slider, message } from 'antd';
 import CreditsFooter from '../credits-footer.js';
+import React, { useEffect, useState } from 'react';
 import { useService } from '../container-context.js';
 import permissions from '../../domain/permissions.js';
 import { Trans, useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


src/components/pages/edit-doc.js
```diff
@@ -333,13 +335,19 @@ function EditDoc({ initialState, PageTemplate }) {
...
-    alerts.push({
-      message: t('proposedSectionsAlert'),
-      type: ALERT_TYPE.info,
-      showInFullScreen: false
-    });
-  }
+  useEffect(() => {
+    const newAlerts = [...globalAlerts];
+
+    if (initialState.proposedSections?.length) {
```
adrianaluchian, line 341: I've changed that to rely on the `initialState.proposedSections` instead of `[state.]proposedSectionKeys.length` from before, which from a "business logic" perspective means that:
- before, the alert was shown for a clone document once a section was selected to be applied on the new document
- now, the alert is shown for a cloned document that has any sections

To me this makes more sense, given that the alert message is generic:
"Apply or discard the proposed sections. Unapplied sections are only visible until you leave the page."

What do you think?

ahelmberger: I didn't quite get the comment, but the behavior now - as far as I can see - is identical to back when the page was a class component (which is what we want).

src/components/pages/doc.js
```diff
@@ -199,27 +202,32 @@ function Doc({ initialState, PageTemplate }) {
...
+            i18nKey="common:externalDocumentWarning"
+            components={[<a key="external-document-warning" href={state.currentDocOrRevision.originUrl} />]}
+            />),
+        type: 'warning',
+        showInFullScreen: false
+      });
+    }
+
+    setAlerts(newAlerts);
+  }, [globalAlerts, state.currentDocOrRevision, isExternalDocument, t]);
```
ahelmberger, line 230: Good that you also catched that one!

ahelmberger (approved):

src/components/pages/doc.js
adrianaluchian: I checked all `globalAlert` cases 😬 They all would have had the same bug.

PR #142, opened 2022-01-07
EDU-224 create task processor to recreate all documents from revisions

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-224

### DIFF
--- a/migrations/automatic/educandu-2022-01-10-01-update-task-and-batch-type-for-imports.js
+++ b/migrations/automatic/educandu-2022-01-10-01-update-task-and-batch-type-for-imports.js
@@ -1,0 +1,18 @@
+import { BATCH_TYPE, TASK_TYPE } from '../../src/domain/constants.js';
+
+// eslint-disable-next-line camelcase
+export default class Educandu_2022_01_10_01_update_task_and_batch_type_for_imports {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    await this.db.collection('batches').updateMany({ batchType: 'import-documents' }, { $set: { batchType: BATCH_TYPE.documentImport } });
+    await this.db.collection('tasks').updateMany({ taskType: 'import-document' }, { $set: { taskType: TASK_TYPE.documentImport } });
+  }
+
+  async down() {
+    await this.db.collection('batches').updateMany({ batchType: BATCH_TYPE.documentImport }, { $set: { batchType: 'import-documents' } });
+    await this.db.collection('tasks').updateMany({ taskType: TASK_TYPE.documentImport }, { $set: { taskType: 'import-document' } });
+  }
+}
--- a/migrations/automatic/educandu-2022-01-10-01-update-task-and-batch-type-for-imports.js
+++ b/migrations/automatic/educandu-2022-01-10-01-update-task-and-batch-type-for-imports.js
@@ -1,5 +1,3 @@
-import { BATCH_TYPE, TASK_TYPE } from '../../src/domain/constants.js';
-
 // eslint-disable-next-line camelcase
 export default class Educandu_2022_01_10_01_update_task_and_batch_type_for_imports {
   constructor(db) {
@@ -7,12 +5,12 @@ export default class Educandu_2022_01_10_01_update_task_and_batch_type_for_impor
   }
 
   async up() {
-    await this.db.collection('batches').updateMany({ batchType: 'import-documents' }, { $set: { batchType: BATCH_TYPE.documentImport } });
-    await this.db.collection('tasks').updateMany({ taskType: 'import-document' }, { $set: { taskType: TASK_TYPE.documentImport } });
+    await this.db.collection('batches').updateMany({ batchType: 'import-documents' }, { $set: { batchType: 'document-import' } });
+    await this.db.collection('tasks').updateMany({ taskType: 'import-document' }, { $set: { taskType: 'document-import' } });
   }
 
   async down() {
-    await this.db.collection('batches').updateMany({ batchType: BATCH_TYPE.documentImport }, { $set: { batchType: 'import-documents' } });
-    await this.db.collection('tasks').updateMany({ taskType: TASK_TYPE.documentImport }, { $set: { taskType: 'import-document' } });
+    await this.db.collection('batches').updateMany({ batchType: 'document-import' }, { $set: { batchType: 'import-documents' } });
+    await this.db.collection('tasks').updateMany({ taskType: 'document-import' }, { $set: { taskType: 'import-document' } });
   }
 }
--- a/src/components/pages/settings.yml
+++ b/src/components/pages/settings.yml
@@ -16,9 +16,9 @@ footerLinksHeader:
 defaultTagsHeader:
   en: Default tags
   de: Vordefinierte Tags
-createDocumentsBatchHeader:
+createRegenerateDocumentsBatchHeader:
   en: Trigger document recreation
   de: Dokumentwiederherstellung auslösen
-createDocumentsBatch:
+createRegenerateDocumentsBatch:
   en: Regenerate Documents
   de: Dokumente neu erstellen
--- a/src/components/pages/settings.yml
+++ b/src/components/pages/settings.yml
@@ -20,5 +20,5 @@ createRegenerateDocumentsBatchHeader:
   en: Trigger document recreation
   de: Dokumentwiederherstellung auslösen
 createRegenerateDocumentsBatchButton:
-  en: Regenerate Documents
+  en: Regenerate documents
   de: Dokumente neu erstellen
--- a/src/components/pages/settings.yml
+++ b/src/components/pages/settings.yml
@@ -19,6 +19,6 @@ defaultTagsHeader:
 createRegenerateDocumentsBatchHeader:
   en: Trigger document recreation
   de: Dokumentwiederherstellung auslösen
-createRegenerateDocumentsBatch:
+createRegenerateDocumentsBatchButton:
   en: Regenerate Documents
   de: Dokumente neu erstellen
--- a/src/components/pages/settings.yml
+++ b/src/components/pages/settings.yml
@@ -16,3 +16,9 @@ footerLinksHeader:
 defaultTagsHeader:
   en: Default tags
   de: Vordefinierte Tags
+createDocumentsBatchHeader:
+  en: Trigger document recreation
+  de: Dokumentwiederherstellung auslösen
+createDocumentsBatch:
+  en: Regenerate Documents
+  de: Dokumente neu erstellen
--- a/src/components/pages/settings.yml
+++ b/src/components/pages/settings.yml
@@ -16,9 +16,9 @@ footerLinksHeader:
 defaultTagsHeader:
   en: Default tags
   de: Vordefinierte Tags
-createRegenerateDocumentsBatchHeader:
+createDocumentRegenerationBatchHeader:
   en: Trigger document recreation
   de: Dokumentwiederherstellung auslösen
-createRegenerateDocumentsBatchButton:
+createDocumentRegenerationBatchButton:
   en: Regenerate documents
   de: Dokumente neu erstellen
--- a/src/domain/permissions.js
+++ b/src/domain/permissions.js
@@ -18,7 +18,7 @@ const MANAGE_EXPORT = 'manage-export';
 const OWN_ROOMS = 'own-rooms';
 const AUTORIZE_ROOMS_RESOURCES = 'authorize-room-resouces';
 const JOIN_PRIVATE_ROOMS = 'join-private-rooms';
-const REGENERATE_ALL_DOCS = 'regenerate-all-docs';
+const REGENERATE_DOCS = 'regenerate-docs';
 
 const rolesForPermission = {
   [EDIT_DOC]: [ROLE.admin, ROLE.user],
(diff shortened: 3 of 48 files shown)


src/components/pages/settings.yml
```diff
@@ -16,3 +16,9 @@ footerLinksHeader:
 defaultTagsHeader:
   en: Default tags
   de: Vordefinierte Tags
+createRegenerateDocumentsBatchHeader:
+  en: Trigger document recreation
+  de: Dokumentwiederherstellung auslösen
+createRegenerateDocumentsBatchButton:
+  en: Regenerate Documents
```
adrianaluchian, line 23: Some German got mixed in here, small 'd' for 'documents' :)

src/domain/permissions.js
```diff
@@ -36,7 +37,8 @@ const rolesForPermission = {
   [MANAGE_IMPORT]: [ROLE.admin],
   [OWN_ROOMS]: [ROLE.admin, ROLE.user],
   [AUTORIZE_ROOMS_RESOURCES]: [ROLE.admin, ROLE.user],
-  [JOIN_PRIVATE_ROOMS]: [ROLE.admin, ROLE.user]
+  [JOIN_PRIVATE_ROOMS]: [ROLE.admin, ROLE.user],
+  [REGENERATE_ALL_DOCS]: [ROLE.admin]
```
adrianaluchian, line 41: Small thing, I'd shorten this and drop the "all" here, to me it brings no extra info cause it's implied, just like with the other permissions (e.g "VIEW_DOCS", "EDIT_SETTINGS" etc.)

src/server/document-controller.js
```diff
@@ -198,6 +198,17 @@ class DocumentController {
     return res.send(result.length ? result[0].uniqueTags : []);
   }
 
+  async handlePostRegenerateDocumentsBatch(req, res) {
+    const { user } = req;
+    const batch = await this.documentService.createRegenerateDocumentsBatch(user);
+
+    if (!batch) {
+      throw new BadRequest('Another document regeneration is in progress');
```
adrianaluchian, line 206: Why not do this the same way as in import batches and throw in the service?

I think that would be better, because `!batch` can mean also other things (a bug, `createRegenerateDocumentsBatch` having changed, who knows?)

src/server/document-controller.js
```diff
@@ -290,6 +301,12 @@ class DocumentController {
       '/api/v1/docs/tags/*',
       (req, res) => this.handleGetDocTags(req, res)
     );
+
+    router.post(
+      '/api/v1/docs/regenerate-documents-batch',
```
adrianaluchian, line 306: To me this looks a bit misplaced and also not inline with the existing batch creation.

What about extracting this in its separate controller and having the endpoint be
`POST /api/v1/document-regenerations/`?
Equivalent to imports.

src/services/document-service.spec.js
```diff
@@ -1038,4 +1048,114 @@ describe('document-service', () => {
...
+      sandbox.stub(documentLockStore, 'releaseLock');
+      revision = {
+        title: 'Title',
+        slug: 'my-doc',
+        language: 'en',
+        sections: [],
+        tags: ['tag-1']
+      };
+
+      result = await sut.createNewDocumentRevision({ doc: revision, user });
```
adrianaluchian, line 1133: This `result` confused me a bit, I thought it's the result we need to assert, as per our test pattern.
Could we please call this `document`?

src/services/document-service.js
```diff
@@ -461,6 +474,67 @@ class DocumentService {
...
+        await this.documentStore.save(document, { session });
+      });
+    } finally {
+      if (lock) {
+        await this.documentLockStore.releaseLock(lock);
+      }
+    }
+  }
+
+  async createRegenerateDocumentsBatch(user) {
```
adrianaluchian, line 498: In order to have a sort of "entity" name, like "imports", and given also my suggestion [here](https://github.com/educandu/educandu/pull/142/files#r780988677), I think `createDocumentsRegenerationBatch` might be better.

src/services/document-service.js
```diff
@@ -461,6 +474,67 @@ class DocumentService {
...
+    const existingActiveBatch = await this.batchStore.findOne({
+      batchType: BATCH_TYPE.regenerateDocuments,
```
adrianaluchian, line 500: Same here, `documentsRegeneration` ?
And we could align the existing one too, cause I think a better name would have been `documentsImport` as a batch name/type.

src/services/regenerate-document-task-processor.spec.js
```diff
@@ -0,0 +1,56 @@
+import sinon from 'sinon';
```
adrianaluchian, line 1: I'd go for `document-regeneration-task-processor.js`, equivalent to `document-import-task-processor.js`

src/domain/permissions.js
laurentiu-ilici: I thought it emphasize that it's a rather special thing (we might have just "regenerate_docs" for lower role users, but true, we can introduce that if needed later). I'll change

src/services/regenerate-document-task-processor.spec.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  afterEach(() => {
+    sandbox.restore();
+  });
+
+  describe('process', () => {
+    beforeEach(() => {
+      sandbox.stub(documentService, 'regenerateDocument');
+    });
```
adrianaluchian, line 31: This could go in the outer most `describe`, with its matching `afterEach`, I think it's better if all stubbing is done at root level (unless it's an occasional stub, where you'd want the real functionality in some cases, which is rare in unit tests).

I find that gathering all stubbing to the top simplifies the setup of inner tests and reduces confusion or double stubbing.

src/server/document-controller.spec.js
```diff
@@ -0,0 +1,92 @@
...
+    documentService = {
+      createRegenerateDocumentsBatch: sandbox.stub()
+    };
+
+    sut = new DocumentController(documentService, clientDataMapper, pageRenderer);
+  });
+
+  afterEach(async () => {
+    await pruneTestEnvironment(container);
+    sandbox.restore();
```
ahelmberger, line 43: You should reverse those lines, because `pruneTestEnvironment` might otherwise use still mocked functions

ahelmberger: Another thing: we usually don't write tests that go to the db on the controller level (except for `client-data-mapper`), so the test environment isn't really needed here, we usually just unit test the controllers and mock the called services.

src/server/document-controller.js
laurentiu-ilici: The current document-service doesn't use any httpErrors, only  the document-controller does. In my opinion is that this is the right way to do it, throwing http errors in a service makes it tied to the way it is exposed (http) to the outside and less reusable in other contexts. 

I know it's convenient to throw those errors from the service, but diminishes the use of the handlers in our controllers.

If we decide for this change I would move all errors from the handlers to the service.

src/server/document-controller.js
laurentiu-ilici: When I put the endpoint here I looked the imports and the batch belongs to the import. Analogously, the regenerate documents batch belongs to the docs (or so I thought).

src/services/document-service.js
```diff
@@ -461,6 +474,67 @@ class DocumentService {
...
+      _id: uniqueId.create(),
+      createdBy: user._id,
+      createdOn: new Date(),
+      completedOn: null,
+      batchType: BATCH_TYPE.regenerateDocuments,
+      batchParams: {},
+      errors: []
+    };
+
+    const allDocuments = await this.documentStore.find({});
```
ahelmberger, line 518: This might not regenerate docs for all revisions (if there were missing docs for example). A safer way is to get all document keys by a unique query over `key` in the revisions collection. We can project to just return the keys.

src/services/document-service.spec.js
laurentiu-ilici: Renamed, it was confusing, sorry

src/server/document-controller.js
adrianaluchian: For me `!batch` does not directly translate into `BadRequest`, it just translates into batch not created, the reason of which only the service knows.

src/services/document-service.spec.js
```diff
@@ -1038,4 +1048,114 @@ describe('document-service', () => {
...
+      let batch = null;
+      beforeEach(async () => {
+        batch = await sut.createRegenerateDocumentsBatch(user);
+      });
+
+      it('should create the new batch', () => {
+        expect(batch).toEqual({
+          _id: expect.stringMatching(/\w+/),
+          createdBy: user._id,
+          createdOn: new Date(),
```
ahelmberger, line 1089: maybe `createdOn: now,` makes it clearer to the reader that we're using a fake timer

src/services/document-service.spec.js
```diff
@@ -1038,4 +1048,114 @@ describe('document-service', () => {
...
+        },
+        { upsert: true }
+      );
+
+      await sut.regenerateDocument(result.key);
+
+      regeneratedDocument = await db.documents.findOne({ _id: result.key });
+    });
+
+    it('should take the a lock on the document', () => {
```
ahelmberger, line 1149: Typo: "the a lock"

src/services/regenerate-document-task-processor.spec.js
laurentiu-ilici: Got it, fixed, thanks

src/services/document-service.js
laurentiu-ilici: I thought we wanted to regenerate existing documents, okay, I'll do that.

src/services/document-service.spec.js
laurentiu-ilici: Fixed

src/server/document-controller.js
ahelmberger: I see both points. We are using http errors in other services, though, because they transport meaning. Returning `null` is a too magical "code" for "Another document regeneration is in progress". The only alternative would be to throw a custom error (but that would also introduce a new concept).

src/server/document-controller.spec.js
laurentiu-ilici: I removed the test environment, thanks.

src/services/document-service.js
laurentiu-ilici: Fixed, now it selects from documentRevisions with "distinct" key.

src/server/document-controller.js
ahelmberger: We could leave it here for now and have a ticket to unify batch processing:

* Having a batch controller
* Having a batch service
* Make the UI show batches of all kinds (imports and regenerations)

src/services/document-service.spec.js
laurentiu-ilici: Fixed, thanks!

src/services/document-service.js
ahelmberger: Maybe rather `documentRegeneration` and `documentImport` as we have a `document-import-task-processor.js` and soon a `document-regeneration-task-processor.js`. In these kind of compound words we always use the singular for the left part (cf. `documents` vs `documentService`), except when it's a fixed name

ahelmberger: But be aware that we use these constants in the database!

src/server/document-controller.js
laurentiu-ilici: I've mode the BadRequest handling to the service.

src/server/document-controller.js
laurentiu-ilici: I've created the story here: https://educandu.atlassian.net/browse/EDU-250

src/services/document-service.js
laurentiu-ilici: I think we need to clear out the dbs for OMA so this is the best time to do this. What do you think? (we can skip having a migration since it's only test data for now.

src/services/regenerate-document-task-processor.spec.js
laurentiu-ilici: Done

src/services/document-service.js
laurentiu-ilici: I've done the "documentRegeneration" part of this comment, shall I do the second part now or add it to the story about unifying the batches? 
I'm a bit afraid this PR will touch quite a few files if I do "documentImport" now, so I think in the end I'd opt for the latter.

src/services/document-service.js
laurentiu-ilici: Renamed to: createDocumentRegenerationBatch as per Andraes' comment.

src/services/document-service.js
laurentiu-ilici: Done, with migration

migrations/automatic/educandu-2022-01-10-01-update-task-and-batch-type-for-imports.js
```diff
@@ -0,0 +1,18 @@
...
+    this.db = db;
+  }
+
+  async up() {
+    await this.db.collection('batches').updateMany({ batchType: 'import-documents' }, { $set: { batchType: BATCH_TYPE.documentImport } });
+    await this.db.collection('tasks').updateMany({ taskType: 'import-document' }, { $set: { taskType: TASK_TYPE.documentImport } });
+  }
+
+  async down() {
+    await this.db.collection('batches').updateMany({ batchType: BATCH_TYPE.documentImport }, { $set: { batchType: 'import-documents' } });
```
laurentiu-ilici, line 15: This down won't really work if we change the value of the constants but I think we will remove the migration soon anyway.

adrianaluchian: Exactly what I wanted to mention, that it would be better to just use hardcoded strings in the migration scripts and not reference any code from outside.

adrianaluchian: Idempotent, that's a word I haven't used in a while 😋

adrianaluchian (approved):

PR #143, opened 2022-01-10
EDU-248 Change document-metadata-editor into a functional component

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-248

### DIFF
--- a/src/components/document-metadata-editor.js
+++ b/src/components/document-metadata-editor.js
@@ -1,20 +1,19 @@
-import React from 'react';
-import autoBind from 'auto-bind';
 import PropTypes from 'prop-types';
 import urls from '../utils/urls.js';
-import { inject } from './container-context.js';
-import { withTranslation } from 'react-i18next';
-import { withSettings } from './settings-context.js';
-import { withLanguage } from './language-context.js';
+import { useTranslation } from 'react-i18next';
+import React, { useEffect, useState } from 'react';
+import { useService } from './container-context.js';
+import { useSettings } from './settings-context.js';
+import { useLanguage } from './language-context.js';
 import { handleApiError } from '../ui/error-helper.js';
 import inputValidators from '../utils/input-validators.js';
 import { Input, Radio, Tag, Space, Select, Form } from 'antd';
 import { EyeOutlined, EditOutlined } from '@ant-design/icons';
 import LanguageSelect from './localization/language-select.js';
+import { documentRevisionShape } from '../ui/default-prop-types.js';
 import LanguageNameProvider from '../data/language-name-provider.js';
 import DocumentApiClient from '../api-clients/document-api-client.js';
 import CountryFlagAndName from './localization/country-flag-and-name.js';
-import { documentRevisionShape, translationProps, languageProps, settingsProps } from '../ui/default-prop-types.js';
 
 const RadioButton = Radio.Button;
 const RadioGroup = Radio.Group;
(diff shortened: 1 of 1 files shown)


src/components/document-metadata-editor.js
```diff
@@ -1,181 +1,162 @@
...
+  const [mode, setMode] = useState(MODE_PREVIEW);
+  const [tagSuggestions, setTagSuggestions] = useState([]);
+  const [slugValidationStatus, setSlugValidationStatus] = useState('');
+  const [tagsValidationStatus, setTagsValidationStatus] = useState('');
 
-  validateMetadata(metadata) {
+  const validateMetadata = metadata => {
     const isValidSlug = inputValidators.isValidSlug(metadata.slug);
-    this.setState({ slugValidationStatus: isValidSlug ? '' : 'error' });
+    setSlugValidationStatus(isValidSlug ? '' : 'error');
```
ahelmberger, line 38: I think we should make this a warning, because we have a lot of docs where it would be red and you can still save them (unless you type something and make it valide)

src/components/document-metadata-editor.js
```diff
@@ -1,181 +1,162 @@
...
 
     const areValidTags = metadata.tags.length > 0 && metadata.tags.every(tag => inputValidators.isValidTag({ tag }));
-    this.setState({ tagsValidationStatus: areValidTags ? '' : 'error' });
+    setTagsValidationStatus(areValidTags ? '' : 'error');
```
ahelmberger, line 41: Same as above, let's make it just a warning

laurentiu-ilici: But then for a tag that has a length of 2 for example, you would not be able to search. We'd need to lower the amount of chars to 2 in the search which is something that I think was suggested as being needed.

Maybe we should modify the search to do a wildcard search for everything that is > 2 chars long and an exact match for a 1 char or 2 char query (in another ticket of course).

src/components/document-metadata-editor.js
laurentiu-ilici: I'm wondering if we can really make it just a warning in any case, what if I just give you the following slug " "?
I think it could be made to be considered "valid" but also pointless or?

We would need to relax the server validation as well. I've tried locally and it doesn't work at the moment.

laurentiu-ilici (approved):

src/components/document-metadata-editor.js
adrianaluchian: Given the points brought up here, I opted for making a dedicated ticket for this (https://educandu.atlassian.net/browse/EDU-251) and keeping this PR as a change that should not have side-effects and behave as before.

Is that ok with you?

ahelmberger (approved):

src/components/document-metadata-editor.js
laurentiu-ilici: The new ticket currently only says something about the slug. Shall we add the tags validation there as well or in a separate ticket? (I'd vote for a second one if we decide to change the search as well)

adrianaluchian: To me the tags validation change is different than changing to show the slug and tags validation as warning and it has more side-effects 🤔
Also, I think probably this part needs to be specified by the PM side:
"_Maybe we should modify the search to do a wildcard search for everything that is > 2 chars long and an exact match for a 1 char or 2 char query (in another ticket of course)._ "

I'd leave this change for now as to me it is a business change and when it's needed it will come from the PM side.

laurentiu-ilici (approved):

PR #145, opened 2022-01-10
Edu 249 fix login issue

laurentiu-ilici:
Closes: https://educandu.atlassian.net/browse/EDU-249

The problem:
We set the Domain on the HTTP only cookie SESSION_ID on prod which causes it to be sent to all subdomains. We need this since we want to send it to cdn.open-music.academy for example for roomId verification.
The drawback is that staging.open-music.academy is also a subdomain, so after logging into staging as well, we send 2 SESSION_ID cookies with the values for prod and for staging. Since the name is the same the servers finds the first one (the prod one) and says you are not logged in.
The current solution:
Have separate names for the session cookies so there is no confusion on the server side although both sessions will be sent.
We will need to change the cookie name setting on the lambdas if we take this solution as well.

I don't immediately see an alternative, at least not by reading the MDN page on cookies.

Linked issue due to the config changes: https://educandu.atlassian.net/browse/EDU-256


src/server/user-controller.js
```diff
@@ -65,6 +65,7 @@ class UserController {
 
   handleGetLogoutPage(req, res) {
     req.logout();
+    res.clearCookie(this.serverConfig.sessionCookieName);
```
laurentiu-ilici, line 68: We were not clearing the cookie on logout, so even logging out from prod was still making you unable to login from staging.

adrianaluchian (approved):

src/server/user-controller.js
ahelmberger: This is by design, as a session doesn't require to be login, you can also have an anonymous session. But we don't use it, so it's fine to clear the cookie here. We just have to remember in the future, in case we store values in the server session.

laurentiu-ilici: Oh, okay, thanks for the comeback.

PR #146, opened 2022-01-10
EDU-245 Add empty lesson page

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-245

### DIFF
--- a/src/domain/schemas/lesson-schemas.js
+++ b/src/domain/schemas/lesson-schemas.js
@@ -17,4 +17,4 @@ export const lessonDBSchema = joi.object({
 export const getLessonParamsSchema = joi.object({
   lessonId: idOrKeySchema.required(),
   lessonSlug: joi.string()
-});
+}).unknown(true);
--- a/src/domain/schemas/lesson-schemas.js
+++ b/src/domain/schemas/lesson-schemas.js
@@ -15,6 +15,6 @@ export const lessonDBSchema = joi.object({
 });
 
 export const getLessonParamsSchema = joi.object({
-  lessonId: joi.string().required(),
+  lessonId: idOrKeySchema.required(),
   lessonSlug: joi.string()
 });
--- a/src/domain/schemas/lesson-schemas.js
+++ b/src/domain/schemas/lesson-schemas.js
@@ -13,3 +13,8 @@ export const lessonDBSchema = joi.object({
   cdnResources: joi.array().items(joi.string()).required(),
   schedule: joi.object().allow(null)
 });
+
+export const getLessonParamsSchema = joi.object({
+  lessonId: joi.string().required(),
+  lessonSlug: joi.string()
+});
--- a/src/resources/page-names.yml
+++ b/src/resources/page-names.yml
@@ -25,3 +25,6 @@ docs:
 room:
   en: Room - {roomName}
   de: Raum - {roomName}
+lesson:
+  en: Lesson - {lessonTitle}
+  de: Unterricht - {lessonTitle}
--- a/src/resources/page-names.yml
+++ b/src/resources/page-names.yml
@@ -27,4 +27,4 @@ room:
   de: Raum - {roomName}
 lesson:
   en: Lesson - {lessonTitle}
-  de: Unterricht - {lessonTitle}
+  de: Unterrichtseinheit - {lessonTitle}
--- a/src/server/document-controller.js
+++ b/src/server/document-controller.js
@@ -57,14 +57,15 @@ class DocumentController {
 
   async handleGetDocPage(req, res) {
     const { user } = req;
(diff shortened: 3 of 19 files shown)


src/domain/schemas/lesson-schemas.js
```diff
@@ -13,3 +13,8 @@ export const lessonDBSchema = joi.object({
   cdnResources: joi.array().items(joi.string()).required(),
   schedule: joi.object().allow(null)
 });
+
+export const getLessonParamsSchema = joi.object({
+  lessonId: joi.string().required(),
```
ahelmberger, line 18: Don't we wanna use the `idOrKeySchema` also for lessons?

src/resources/page-names.yml
```diff
@@ -25,3 +25,6 @@ docs:
 room:
   en: Room - {roomName}
   de: Raum - {roomName}
+lesson:
+  en: Lesson - {lessonTitle}
+  de: Unterricht - {lessonTitle}
```
ahelmberger, line 30: We have to use "Unterrichtseinheit" here

src/server/document-controller.js
```diff
@@ -217,12 +204,14 @@ class DocumentController {
...
     router.get(
       '/docs/:docKey',
+      validateParams(getDocumentParamsSchema),
       (req, res) => this.handleGetDocPage(req, res)
     );
 
     router.get(
-      '/docs/:docKey/*',
-      (req, res) => this.handleGetDocPageWithSlug(req, res)
+      '/docs/:docKey/:docSlug',
```
ahelmberger, line 212: To match also slugs that contain slashes ("catchall"), the route must be:

~~~
'/docs/:docKey/:docSlug*'
~~~

src/server/lesson-controller.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    router.get(
+      '/lessons/:lessonId',
+      [validateParams(getLessonParamsSchema)],
+      (req, res) => this.handleGetLessonPage(req, res)
+    );
+
+    router.get(
+      '/lessons/:lessonId/:lessonSlug',
```
ahelmberger, line 49: also here: `:lessonSlug*`

src/utils/urls.js
```diff
@@ -144,18 +145,12 @@ function getRoomUrl(roomId) {
   return concatParts(roomsPrefix, encodeURIComponent(roomId));
 }
 
+function getLessonUrl(id, slug) {
+  return concatParts(lessonsPrefix, encodeURIComponent(id), encodeURIComponent(slug));
```
ahelmberger, line 149: `encodeURIComponent(slug)` will escape slashes in the slug, therefore altering the path. I see that we use them "literally" in the somewhat analogous case of docs (which is also not good). I think, to be 100% correct, we would have to do in both cases:

~~~
slug.split('/').map(x => encodeURIComponent(x)).join('/')
~~~
?

src/resources/page-names.yml
adrianaluchian: Updated, danke

src/domain/schemas/lesson-schemas.js
adrianaluchian: Of course, we should. Updated here and in all room schemas as well.

src/server/lesson-controller.js
adrianaluchian: We URI encode the slug in lessons, this should not be an issue. However, since we probably didn't use to do that with articles before, we likely have slashes out there in links.

For consistency I updated both `lessons` and `docs` paths.

src/server/document-controller.js
adrianaluchian: See https://github.com/educandu/educandu/pull/146#discussion_r781997236

src/utils/urls.js
adrianaluchian: Good point, I've used this for both lessons and documents, thanks!

src/server/lesson-controller.js
```diff
@@ -0,0 +1,49 @@
...
+
+  constructor(serverConfig, lessonService, pageRenderer) {
+    this.serverConfig = serverConfig;
+    this.lessonService = lessonService;
+    this.pageRenderer = pageRenderer;
+  }
+
+  async handleGetLessonPage(req, res) {
+    const { lessonId } = req.params;
+    const routeWildcardValue = urls.trimSlashes(req.params[0]);
```
ahelmberger, line 23: I would suggest to use `removeLeadingSlash`, because the original slug could end with a trailing slash and then they wouldn't match anymore.

adrianaluchian: Of course, we don't have nice sanitized slugs in the database. Changed, thanks!

ahelmberger (approved):

PR #148, opened 2022-01-11
EDU-225 remove manual migrations

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-225


src/services/maintenance-service.spec.js
```diff
@@ -0,0 +1,104 @@
...
+        database.hasPendingMigrationScripts.resolves(false);
+        maintenanceLockStore.takeLock.resolves({});
+        await sut.runMaintenance();
+      });
+
+      it('should not have taken the lock', () => {
+        sinon.assert.notCalled(maintenanceLockStore.takeLock);
+      });
+
+      it('should have run the migrations', () => {
```
adrianaluchian, line 50: `should not have run the migrations`

src/services/maintenance-service.spec.js
```diff
@@ -0,0 +1,104 @@
...
+        beforeEach(async () => {
+          sandbox.stub(MaintenanceService, 'MAINTENANCE_LOCK_INTERVAL_IN_SEC').value(0);
+          maintenanceLockStore.takeLock
+            .onFirstCall().rejects({ code: 11000 })
+            .onSecondCall().resolves({});
+
+          await sut.runMaintenance();
+        });
+
+        it('should have checked for pending migrations the lock twice', () => {
```
adrianaluchian, line 91: `should have checked for pending migrations twice`

src/utils/time-utils.js
```diff
@@ -0,0 +1,5 @@
+export function delay(milliseconds) {
+  return new Promise(resolve => {
+    setTimeout(resolve, milliseconds);
+  });
+}
```
adrianaluchian, line 5: 👍  😊

ahelmberger: Your idea :smile:

src/services/maintenance-service.spec.js
ahelmberger: Fixed, thanks, I should check that better

src/services/maintenance-service.spec.js
ahelmberger: Copy/paste error, sorry. Fixed

src/index.js
```diff
@@ -35,37 +39,59 @@ export default async function educandu(options) {
...
+
+    logger.info('Starting server in maintenance mode');
+    const port = await educanduServer.listen({ maintenance: true });
+    logger.info(`Server listening on http://localhost:${port}`);
+
+    if (options.skipMongoMigrations) {
+      logger.info('Skipping database migrations');
+    } else {
+      logger.info('Starting database migrations');
+      await maintenanceService.runMaintenance();
```
adrianaluchian, line 60: Since you opted for this extendable more generic approach, the log messages no longer fit.

How about `maintenanceService.runMaintenance({ skipMongoMigrations: options.skipMongoMigrations })` and having the service do the particular db related logging, and in the future any other maintenance actions?

src/server/educandu-server.js
```diff
@@ -58,34 +75,40 @@ class EducanduServer {
...
-      Promise.resolve().then(() => cb(null, this.serverConfig.port));
+      logger.info('Closing server');
+      await util.promisify(this.server.close.bind(this.server))();
+      this.server = null;
+      logger.info('Server successfully closed');
     } else {
-      logger.info('Starting server');
-      this.server = this.app.listen(this.serverConfig.port, err => err ? cb(err) : cb(null, this.serverConfig.port));
+      logger.info('Cannot close server, it is not listening');
+      await Promise.resolve();
```
adrianaluchian, line 104: I think this is not needed here.

src/index.js
ahelmberger: Sounds like a good idea. I will then also move the db checks into this maintenance service and keep the server config value!

migrations/educandu-2022-01-07-01-create-maintenance-locks-collection.js
```diff
@@ -0,0 +1,26 @@
+// eslint-disable-next-line camelcase
+export default class Educandu_2022_01_07_01_create_maintenance_locks_collection {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    const existingCollections = await this.db.listCollections().toArray();
+
+    if (!existingCollections.find(x => x.name === 'maintenanceLocks')) {
```
adrianaluchian, line 10: Not sure I get when this would be the case, that the collection is there 🤔

ahelmberger: It is always there, because this runs in maintenance mode, after the lock on this collection has already been taken. It's a hen/egg problem. This script basically has a semi-ceremonial effect, it does create the index though.

src/server/educandu-server.js
ahelmberger: Historical code, I removed the whole `else` block, because also the log message is not really helpful.

migrations/educandu-2022-01-07-01-create-maintenance-locks-collection.js
adrianaluchian: Ah true, the lock is needed for the maintenance to start, but the script is run only during maintenance..
Understood.

src/server/educandu-server.js
```diff
@@ -58,34 +75,40 @@ class EducanduServer {
...
+  exitMaintenanceMode() {
+    this.app.locals[SYMBOL_MAINTENACE] = false;
+  }
+
+  listen({ maintenance = false }) {
+    this.app.locals[SYMBOL_MAINTENACE] = !!maintenance;
+    return new Promise((resolve, reject) => {
+      const { port } = this.serverConfig;
+      if (this.server) {
+        logger.info('Cannot start server, it is already listining');
```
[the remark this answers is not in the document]
ahelmberger: Corrrrrrrected!

src/index.js
adrianaluchian: Looks good to me!

src/index.js
```diff
@@ -35,37 +38,55 @@ export default async function educandu(options) {
...
-      if (err) {
-        logger.fatal(err);
-        Graceful.exit(1);
+    logger.info('Starting application');
+
+    const runMaintenance = !serverConfig.skipMaintenance;
+
+    logger.info(`Starting server${runMaintenance ? ' in maintenance mode' : ''}`);
+    const port = await educanduServer.listen({ maintenance: runMaintenance });
+    logger.info(`Server listening on http://localhost:${port}`);
```
adrianaluchian, line 54: Maybe we can run this at the end, after `Application started successfully`. I know it's already listening but this info is lost halfway in the logs 😬

ahelmberger: Done

adrianaluchian (approved):

laurentiu-ilici (approved):

PR #156, opened 2022-01-13
EDU-260 Streamline db collections

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-260

**Note:**
- due to the stray migration script that had not been run
(`educandu-2022-01-10-01-update-task-and-batch-type-for-imports.js`)
I've split the ticket into 2 steps, the follow-up ticket being https://educandu.atlassian.net/browse/EDU-264


src/stores/document-lock-store.js
```diff
@@ -1,11 +1,13 @@
 import Database from '../stores/database.js';
 import LockStoreBase from './lock-store-base.js';
 
+const LOCK_EXPIRATION_TIME_SPAN = { minutes: 30 };
```
ahelmberger, line 4: I was thinking if we need such a long expiration. Maybe 1 minute is enough?

adrianaluchian: That was the next step, consulting you on how long this should be 🤔 I also think 1min should be enough, maybe even less but let's air on the safe side.

adrianaluchian: Reduced to 1 min for document, batch and room locks.

ahelmberger (approved):

PR #158, opened 2022-01-14
EDU-242 Ask for re-login when session has expired

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-242


src/components/account-tab.js
```diff
@@ -1,15 +1,15 @@
 import Logger from '../common/logger.js';
 import EmailInput from './email-input.js';
-import { useUser } from './user-context.js';
 import { Form, Button, message } from 'antd';
 import { useTranslation } from 'react-i18next';
 import errorHelper from '../ui/error-helper.js';
 import UsernameInput from './username-input.js';
-import React, { useRef, useState } from 'react';
-import { useService } from './container-context.js';
+import React, { useEffect, useRef, useState } from 'react';
```
adrianaluchian, line 7: 🙈 🙉 🙊

src/components/account-tab.js
```diff
@@ -56,18 +60,13 @@ function AccountTab({ formItemLayout, tailFormItemLayout }) {
...
-      userApiClient
-    });
+    dialogs.confirmWithPassword(user.username, () => saveAccountData({ username, email }));
   };
 
   const handleResetPasswordClick = async () => {
     try {
-      await userApiClient.requestPasswordReset({ email: state.email });
-      message.success(t('passwordResetEmailSent', { email: state.email }));
+      await userApiClient.requestPasswordReset({ email: user.email });
```
adrianaluchian, line 68: This means this was a bug before, if one would change the email, not save, then request password reset, it would be sent to the wrong email..

src/components/account-tab.js
ahelmberger: :laughing: :rofl: :boom:

src/components/account-tab.js
ahelmberger: Yes

src/ui/api-helper.js
```diff
@@ -0,0 +1,77 @@
...
+  let firstTryError = null;
+  try {
+    result = await executeCall();
+  } catch (err) {
+    firstTryError = err;
+  }
+
+  if (!firstTryError) {
+    return result;
+  }
```
adrianaluchian, line 18: Actually this `if` can go away and just return on line 11, right? Would simplify this long method a bit.
I guess you wanted to make everything more explicit, also with the `secondTryError`, but I would say we could just replace lines `31` -> `42` with `return executeCall()` ?

src/ui/api-helper.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        executeCall.resolves({ result: 'abc' });
+        result = await tryApiCallWithLoginFallback({ executeCall, onLoginRequested });
+      });
+      it('returns the result of the first API call', () => {
+        expect(result).toEqual({ result: 'abc' });
+      });
+      it('does not execute the API call a second time', () => {
+        sinon.assert.calledOnce(executeCall);
+      });
+      it('does not invoce the re-login callback', () => {
```
adrianaluchian, line 28: Typo `invoce` -> `invoke`, also on line `87`.

src/ui/api-helper.js
ahelmberger: Thanks for the suggestion, it looks a lot simpler now.

src/ui/api-helper.js
```diff
@@ -0,0 +1,77 @@
...
+    dialogs.reloginAfterSessionExpired(() => resolve(true), () => resolve(false));
+  });
+}
+
+function createFunctionProxy(func, dialogs) {
+  return new Proxy(func, {
+    apply(target, thisArg, argumentsList) {
+      return tryApiCallWithLoginFallback({
+        executeCall: () => target.apply(thisArg, argumentsList),
+        onLoginRequested: () => handleLoginRequested(dialogs)
```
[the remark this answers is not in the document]
ahelmberger: Yes, I also like your name better, fixed :-)

src/ui/api-helper.spec.js
ahelmberger: Fixed

adrianaluchian (approved):

PR #160, opened 2022-01-14
EDU-265 Add cache headers to HTML pages and remove client side hook

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-265

### DIFF
--- a/src/server/page-renderer.js
+++ b/src/server/page-renderer.js
@@ -76,7 +76,14 @@ class PageRenderer extends PageRendererBase {
       contentProps: props
     });
 
-    return res.type('html').send(html);
+    return res
+      .type('html')
+      .set({
+        'Cache-Control': 'max-age=0, no-cache, no-store, must-revalidate',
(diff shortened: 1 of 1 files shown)


src/server/page-renderer.js
```diff
@@ -76,7 +76,14 @@ class PageRenderer extends PageRendererBase {
       contentProps: props
     });
 
-    return res.type('html').send(html);
+    return res
+      .type('html')
+      .set({
+        'Cache-Control': 'max-age=0, no-cache, no-store, must-revalidate',
+        'Pragma': 'no-cache',
+        'Expires': 'Wed, 11 Jan 1984 05:00:00 GMT'
```
adrianaluchian, line 84: Just to double check that this is not a copy paste thing: we need to set this in the past so that the cache is expired immediately, right?

ahelmberger: Yes :laughing:

adrianaluchian (approved):

PR #161, opened 2022-01-17
EDU-239 Fix prop types warnings

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-239

### DIFF
--- a/src/server/client-data-mapper.js
+++ b/src/server/client-data-mapper.js
@@ -168,10 +168,12 @@ class ClientDataMapper {
 
   _mapRoomInvitation(rawInvitation) {
     const sentOn = rawInvitation.sentOn && rawInvitation.sentOn.toISOString();
+    const expires = rawInvitation.expires && rawInvitation.expires.toISOString();
 
     return {
       ...rawInvitation,
-      sentOn
+      sentOn,
+      expires
     };
   }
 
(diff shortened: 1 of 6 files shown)


src/server/client-data-mapper.js
```diff
@@ -3,12 +3,144 @@ import cloneDeep from '../utils/clone-deep.js';
...
+  const updatedOn = rawLesson.updatedOn && rawLesson.updatedOn.toISOString();
+  const schedule = rawLesson.schedule && mapLessonSchedule(rawLesson.schedule);
+
+  return {
+    ...rawLesson,
+    createdOn,
+    updatedOn,
+    schedule
+  };
+};
```
ahelmberger, line 143: Is there a reason why you implemented all those function outside of the class?

adrianaluchian: Felt like the class was super convoluted 🤔 But I'm not very attached to the idea, I can place them inside the class if you dislike this.

adrianaluchian: Reconsidered, moved everything in the class 👍

ahelmberger: Ok, I felt the class was a little bit empty before :-)

ahelmberger (approved):

PR #163, opened 2022-01-17
EDU-266 Execute search after loading search page

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-266

### DIFF
--- a/src/components/document-metadata-editor.js
+++ b/src/components/document-metadata-editor.js
@@ -2,19 +2,17 @@ import PropTypes from 'prop-types';
 import urls from '../utils/urls.js';
 import { useTranslation } from 'react-i18next';
 import React, { useEffect, useState } from 'react';
-import { useService } from './container-context.js';
 import { useSettings } from './settings-context.js';
 import { useLanguage } from './language-context.js';
 import { handleApiError } from '../ui/error-helper.js';
 import inputValidators from '../utils/input-validators.js';
 import { Input, Radio, Tag, Space, Select, Form } from 'antd';
 import { EyeOutlined, EditOutlined } from '@ant-design/icons';
+import LanguageFlagAndName from './language-flag-and-name.js';
 import LanguageSelect from './localization/language-select.js';
 import { useSessionAwareApiClient } from '../ui/api-helper.js';
 import { documentRevisionShape } from '../ui/default-prop-types.js';
-import LanguageNameProvider from '../data/language-name-provider.js';
 import DocumentApiClient from '../api-clients/document-api-client.js';
-import CountryFlagAndName from './localization/country-flag-and-name.js';
 
 const RadioButton = Radio.Button;
 const RadioGroup = Radio.Group;
@@ -26,7 +24,6 @@ function DocumentMetadataEditor({ documentRevision, onChanged }) {
   const settings = useSettings();
   const { language } = useLanguage();
   const { t } = useTranslation('documentMetadataEditor');
-  const languageNameProvider = useService(LanguageNameProvider);
   const documentApiClient = useSessionAwareApiClient(DocumentApiClient);
 
   const [mode, setMode] = useState(MODE_PREVIEW);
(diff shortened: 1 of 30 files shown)


src/components/document-metadata-editor.js
```diff
@@ -86,15 +83,13 @@ function DocumentMetadataEditor({ documentRevision, onChanged }) {
...
-    const docLanguage = languageNameProvider.getData(language)[documentRevision.language];
-
     switch (mode) {
       case MODE_PREVIEW:
         return (
           <div>
             <span>{t('common:title')}:</span> <span>{documentRevision.title}</span>
             <br />
-            <span>{t('common:language')}:</span> <span><CountryFlagAndName code={docLanguage.flag} name={docLanguage.name} /></span>
+            <span>{t('common:language')}:</span> <span><LanguageFlagAndName language={language} /></span>
```
adrianaluchian, line 92: This should be document language instead of UI language.

src/components/pages/docs.js
```diff
@@ -114,8 +110,7 @@ function Docs({ initialState, PageTemplate }) {
   };
 
   const renderLanguage = documentLanguage => {
-    const lang = languageNameProvider.getData(language)[documentLanguage];
-    return <CountryFlagAndName code={lang.flag} name={`${documentLanguage} (${lang.name})`} flagOnly />;
+    return <LanguageFlagAndName language={documentLanguage} flagOnly />;
```
adrianaluchian, line 113: Nice simplification 😎

src/components/pages/search.js
```diff
@@ -1,54 +1,67 @@
...
-import { useDateFormat, useLanguage } from '../language-context.js';
-import LanguageNameProvider from '../../data/language-name-provider.js';
-import CountryFlagAndName from '../localization/country-flag-and-name.js';
+import SearchApiClient from '../../api-clients/search-api-client.js';
+
+function sortDocs(docs) {
+  return docs
+    .map(doc => ({
+      ...doc,
+      tags: [...new Set(doc.tags)], // Some docs have duplicate tags we don't want to render
```
adrianaluchian, line 23: I know this code was there before, but I think now we can remove the comment since we ran the script to reduce duplicates.

src/server/search-controller.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  handleGetSearchPage(req, res) {
+    return this.pageRenderer.sendPage(req, res, PAGE_NAME.search, {});
+  }
+
+  async handleGetSearchResult(req, res) {
+    const { query } = req.query;
+    const docs = await this.documentService.getDocumentsMetadataByTags(query);
+    const result = await this.clientDataMapper.mapDocsOrRevisions(docs, req.user);
```
adrianaluchian, line 24: Nice that you introduced mapping, was just checking that after seeing a PropTypes warning for `updatedOn` 😬

src/components/document-metadata-editor.js
ahelmberger: Oh, yes. Thanks, fixed.

src/components/pages/docs.js
ahelmberger: We had it in so many places that I felt it was time to extract it.

src/components/pages/search.js
ahelmberger: Yes, removed, and that started a chain of more simplifications on the page.

src/server/search-controller.js
ahelmberger: I actually could delete the searchResultPropType as we do not have initial state for this page anymore. Thanks for hinting me to it :-)

src/components/pages/search.js
adrianaluchian: Oh nice, especially with gathering the filtering + sorting of docs in one place and simplifying it. Was a bit scattered and complicated before..

adrianaluchian (approved):

PR #165, opened 2022-01-20
EDU-268 Extract email translations

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-268


src/components/profile-tab.yml
```diff
@@ -30,15 +30,15 @@ updateSuccessMessage:
...
-    ELMU does not manage profile pictures itself, but uses the <0>Gravatar</0> internet service.
+    Profile pictures are managed through the <0>Gravatar</0> internet service.
     A Gravatar is a globally recognized avatar, which is linked to your email address.
     This service can then be used by other websites to display a profile picture for a given email address.
-    Please note that ELMU has no influence on the use of your data by this service.
+    Please note that our website has no influence on the use of your data by this service.
     If you agree, you can create a profile <0>here</0>.
   de: |
-    ELMU verwaltet Profilbilder nicht selbst, sondern verwendet den Internetdienst <0>Gravatar</0>.
+    Profilbilder sind durch den Internetdienst <0>Gravatar</0> geführt.
```
ahelmberger, line 39: "sind geführt" is OK, but "werden verwaltet" would sound better

adrianaluchian: 🙏 Changed

src/services/mail-service.js
```diff
@@ -1,149 +1,106 @@
...
+    const subject = this.translators
+      .map(t => t('mailService:registrationVerificationEmail.subject', { username, verificationLink }))
+      .join(' / ');
+
+    const text = this.translators
+      .map(t => t('mailService:registrationVerificationEmail.text', { username, verificationLink }))
+      .join('\n\n');
+
+    const html = this.translators
+      .map(t => t('mailService:registrationVerificationEmail.html', { username, verificationLink }))
```
ahelmberger, line 32: Actually, we should HTML encode these values. What if someone chooses this username: "<script>doSmthBad()</script>"?

src/services/mail-service.js
```diff
@@ -1,149 +1,106 @@
...
+    const html = this.translators
+      .map(t => t('mailService:roomDeletionNotificationEmail.html', { ownerName, roomName }))
+      .join('\n');
 
+    const message = { from: this.emailSenderAddress, to: email, subject, text, html };
     return this._sendMail(message);
   }
 
+  _sanitizeText(translation) {
+    return translation.replaceAll('\\n', '\n');
```
ahelmberger, line 106: This seems to mitigate an error that happened earlier in the chain (converting yml to JSON seems to sneak it in). We should fix the root cause here, I think

src/services/mail-service.js
adrianaluchian: Very good point, done, thanks!

src/services/mail-service.js
adrianaluchian: I'm not sure if this won't break other translation cases.

ahelmberger (approved):

PR #177, opened 2022-01-25
EDU-286 Make license configurable through settings

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-286

### DIFF
--- a/src/ui/validation.js
+++ b/src/ui/validation.js
@@ -6,7 +6,7 @@ const URL_SECURE_PROTO_REGEX = /^https:\/\/.+$/;
 
 export const MARKDOWN_REGEX_BOLD_OR_ITALIC_WITHIN_HEADERS = /^[#]+\s*[_*]+.+[_*]+\s*[#]*\s*$/m;
 
-export function validateUrl(url, t, { allowEmpty = false, allowInsecure = false } = {}) {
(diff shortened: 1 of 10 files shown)


src/ui/validation.js
```diff
@@ -6,13 +6,13 @@ const URL_SECURE_PROTO_REGEX = /^https:\/\/.+$/;
 
 export const MARKDOWN_REGEX_BOLD_OR_ITALIC_WITHIN_HEADERS = /^[#]+\s*[_*]+.+[_*]+\s*[#]*\s*$/m;
 
-export function validateUrl(url, t, { allowInsecure = false } = {}) {
+export function validateUrl(url, t, { allowEmpty = false, allowInsecure = false } = {}) {
   let validateStatus;
```
adrianaluchian, line 10: I'd like to suggest that we set the fallback object too:
`validateUrl(url, t, { allowEmpty = false, allowInsecure = false } = { allowEmpty: false, allowInsecure: false })`

In this case both values would be falsy anyway, of course there would be no difference in the result when one would call `validateUrl('url', t, { allowEmpty: false })` vs `validateUrl('url', t)`, but I think it's a good pattern for us to follow.

ahelmberger: Agree, changed, thanks

adrianaluchian (approved):

PR #179, opened 2022-01-26
EDU-247 Edit control panel

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-247

Added restriction for room owner.


ahelmberger:
The `DocView` used here on the page has not been updated yet to also handle lessons, but we can do this later when we can test this without manually updating the db.

ahelmberger (approved):

adrianaluchian:
> The `DocView` used here on the page has not been updated yet to also handle lessons, but we can do this later when we can test this without manually updating the db.

👍 Gladly removed! Was just thinking you might want to keep it there for the next step.

PR #181, opened 2022-01-26
EDU-296 Edit lesson metadata

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-296


src/components/lesson-metadata-modal.js
```diff
@@ -10,24 +10,30 @@ import React, { useState, useRef, useEffect } from 'react';
...
 
 const FormItem = Form.Item;
 
 const logger = new Logger(import.meta.url);
 
+export const LESSON_MODAL_MODE = {
+  create: 'create',
+  update: 'update'
+};
+
```
ahelmberger, line 24: I put the comment here, although I mean line 131 (but that one is not changed, so Github doesn't let me put a comment there). The title of the modal should not always be `t('newLesson')` when `mode === LESSON_MODAL_MODE.update`

adrianaluchian: Nice catch, than you, corrected!

ahelmberger (approved):

PR #183, opened 2022-01-26
EDU-290 Create plugin selector dialog

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-290

### DIFF
--- a/src/components/grid-selector.js
+++ b/src/components/grid-selector.js
@@ -1,0 +1,57 @@
+import React, { useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+function GridSelector({ items, selectedItemKey, onSelectionChange }) {
+  const listRef = useRef();
+
+  useEffect(() => {
+    const selectedElement = listRef.current?.querySelector(`[data-item-key="${selectedItemKey.replaceAll('"', '\\"')}"]`);
+    if (selectedElement && selectedElement !== document.activeElement) {
+      selectedElement.focus();
+    }
+  }, [selectedItemKey]);
+
+  const handleFocusCapture = event => {
+    const itemKey = event.target.getAttribute('data-item-key');
+    if (itemKey && itemKey !== selectedItemKey) {
+      onSelectionChange?.(itemKey, false);
+    }
+  };
+
+  return (
+    <ul ref={listRef} className="GridSelector" onFocusCapture={handleFocusCapture}>
+      {items.map((item, index) => (
+        <li
+          key={item.key}
+          tabIndex={index + 1}
+          data-item-key={item.key}
+          className={classNames('GridSelector-item', { 'GridSelector-item--selected': item.key === selectedItemKey })}
+          onKeyUp={event => event.key === 'Enter' && onSelectionChange?.(item.key, true)}
+          onDoubleClick={() => onSelectionChange?.(item.key, true)}
+          >
+          <div className="GridSelector-itemIcon">{item.icon}</div>
+          <div className="GridSelector-itemLabel">{item.label}</div>
+        </li>
+      ))}
+    </ul>
+  );
+}
+
+GridSelector.propTypes = {
+  items: PropTypes.arrayOf(PropTypes.shape({
+    key: PropTypes.string.isRequired,
+    icon: PropTypes.node,
+    label: PropTypes.node
+  })),
+  onSelectionChange: PropTypes.func,
+  selectedItemKey: PropTypes.string
+};
+
+GridSelector.defaultProps = {
+  items: [],
+  onSelectionChange: () => {},
+  selectedItemKey: null
+};
+
+export default GridSelector;
--- a/src/components/grid-selector.js
+++ b/src/components/grid-selector.js
@@ -12,23 +12,36 @@ function GridSelector({ items, selectedItemKey, onSelectionChange }) {
     }
   }, [selectedItemKey]);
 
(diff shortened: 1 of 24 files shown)


src/components/grid-selector.less
```diff
@@ -0,0 +1,41 @@
...
+  gap: 2px;
+  height: auto;
+  width: 100%;
+  margin: 0;
+  padding: @edu-button-spacing;
+  border: 1px solid transparent;
+  border-radius: @edu-border-radius-base;
+  cursor: pointer;
+  user-select: none;
+  outline: none !important; // Removes user agent black border when focused by keyboard
```
adrianaluchian, line 26: Should this maybe go into `&.GridSelector-item--selected &:focus` ? Or maybe in it's own mixin?

src/components/grid-selector.js
```diff
@@ -0,0 +1,57 @@
...
+    const selectedElement = listRef.current?.querySelector(`[data-item-key="${selectedItemKey.replaceAll('"', '\\"')}"]`);
+    if (selectedElement && selectedElement !== document.activeElement) {
+      selectedElement.focus();
+    }
+  }, [selectedItemKey]);
+
+  const handleFocusCapture = event => {
+    const itemKey = event.target.getAttribute('data-item-key');
+    if (itemKey && itemKey !== selectedItemKey) {
+      onSelectionChange?.(itemKey, false);
```
adrianaluchian, line 18: Elvis can leave the building, since `onSelectionChange` defaults to empty function.

src/components/grid-selector.js
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  return (
+    <ul ref={listRef} className="GridSelector" onFocusCapture={handleFocusCapture}>
+      {items.map((item, index) => (
+        <li
+          key={item.key}
+          tabIndex={index + 1}
+          data-item-key={item.key}
+          className={classNames('GridSelector-item', { 'GridSelector-item--selected': item.key === selectedItemKey })}
```
adrianaluchian, line 29: This class matches better as a component state, `is-selected`, I feel.

src/components/plugin-selector-dialog.js
```diff
@@ -0,0 +1,57 @@
...
+function PluginSelectorDialog({ visible, onSelected, onCancel }) {
+  const { t } = useTranslation('pluginSelectorDialog');
+  const infoFactory = useService(InfoFactory);
+
+  const pluginItems = useMemo(() => {
+    return infoFactory.getRegisteredTypes()
+      .map(typeName => infoFactory.createInfo(typeName))
+      .map(info => ({
+        key: info.type,
+        icon: info.getIcon?.(t) || null,
```
adrianaluchian, line 18: We could use a default icon here, maybe `<QuestionOutlined />`

src/components/plugin-selector-dialog.js
```diff
@@ -0,0 +1,57 @@
...
+        label: info.getName(t)
+      }));
+  }, [infoFactory, t]);
+
+  const [selectedPluginType, setSelectedPluginType] = useState(pluginItems[0]?.key || null);
+
+  const handleOk = () => {
+    onSelected?.(selectedPluginType);
```
adrianaluchian, line 26: Same about this Elvis 😋

src/components/plugin-selector-dialog.js
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  const handleCancel = () => onCancel?.();
```
adrianaluchian, line 29: Should we reset `selectedPluginType` on cancel (or might need to be on ok), so that the first item is always selected when re-opening the modal? 🤔

adrianaluchian:
Looks really really nice, this plugin selection modal! 🥳

src/components/grid-selector.less
ahelmberger: It should and it did (both of your suggestions) :smile:

src/components/grid-selector.js
ahelmberger: Elvis has stopped rocking the house, thanks!

src/components/grid-selector.js
ahelmberger: Correct, changed it.

src/components/plugin-selector-dialog.js
ahelmberger: Good idea, changed, thanks!

src/components/plugin-selector-dialog.js
ahelmberger: He's gone! (or is he ... :thinking:)

src/components/plugin-selector-dialog.js
ahelmberger: Actually, I intended to persist the last choice in session storage until I realized that it's working out of the box in the React state. I think it might be logical to the user to pre-select the last selected plugin while being on the same page.

ahelmberger:
> Looks really really nice, this plugin selection modal! partying_face

Thankya :relaxed:

src/components/grid-selector.js
adrianaluchian: 🕺

src/components/plugin-selector-dialog.js
adrianaluchian: So it was intentional 👍  Ok, then let's have it preselected.

adrianaluchian (approved):

PR #184, opened 2022-01-26
EDU-291 Render lesson sections in view mode

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-291


src/ui/default-prop-types.js
```diff
@@ -8,13 +8,12 @@ export const translationProps = {
 };
 
 export const sectionDisplayProps = {
-  docKey: PropTypes.string.isRequired,
-  sectionKey: PropTypes.string.isRequired,
```
adrianaluchian, line 12: Small clarification here:
- `sectionKey` was never used in either display or edit section cases
- `docKey` was only used in some of the edit section cases, I changed it into the generic `containerParentId` (aka document key or lesson._id)

src/components/sections-view.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import SectionDisplay from './section-display.js';
+import { sectionShape } from '../ui/default-prop-types.js';
+
+function SectionsView({ sections, sectionsContainerId, onAction }) {
+  return (
+    <article data-document-key={sectionsContainerId}>
```
ahelmberger, line 8: We can remove this data attribute, I cannot find any usages.

src/ui/default-prop-types.js
ahelmberger: When we overhaul the file picker component, this can disappear completely (it's currently used to set the initial folder to show)

adrianaluchian: Yes, I noticed that it's only used by the components that need to upload to S3 🤔

src/components/sections-view.js
adrianaluchian: Nice, done! The `sectionsContainerId` prop went away along with it.

ahelmberger (approved):

PR #186, opened 2022-01-26
EDU-300 Fix terms and conditions link

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-300

### DIFF
--- a/src/components/pages/register.js
+++ b/src/components/pages/register.js
@@ -23,7 +23,7 @@ const FormItem = Form.Item;
 function Register({ PageTemplate, SiteLogo }) {
   const formRef = useRef(null);
   const settings = useSettings();
-  const language = useLanguage();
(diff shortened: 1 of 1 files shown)


src/components/pages/register.js
```diff
@@ -23,7 +23,7 @@ const FormItem = Form.Item;
 function Register({ PageTemplate, SiteLogo }) {
   const formRef = useRef(null);
   const settings = useSettings();
-  const language = useLanguage();
+  const { language } = useLanguage();
```
ahelmberger, line 26: Huh :monocle_face:

ahelmberger (approved):

src/components/pages/register.js
adrianaluchian: Was the only such case.

PR #189, opened 2022-01-26
EDU-301 Redirect room and lesson pages when session expired

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-301

### DIFF
--- a/src/server/lesson-controller.js
+++ b/src/server/lesson-controller.js
@@ -38,6 +38,10 @@ export default class LessonController {
     const { lessonId } = req.params;
     const routeWildcardValue = urls.removeLeadingSlash(req.params['0']);
 
+    if (!user) {
+      return res.redirect(urls.getLoginUrl(req.path));
+    }
+
     const lesson = await this.lessonService.getLessonById(lessonId);
 
     if (!lesson) {
--- a/src/server/lesson-controller.js
+++ b/src/server/lesson-controller.js
@@ -20,7 +20,7 @@ import {
 } from '../domain/schemas/lesson-schemas.js';
 
 const jsonParser = express.json();
-const { NotFound, BadRequest, Forbidden } = httpErrors;
+const { NotFound, BadRequest, Forbidden, Unauthorized } = httpErrors;
 
 export default class LessonController {
   static get inject() { return [ServerConfig, LessonService, RoomService, ClientDataMapper, PageRenderer]; }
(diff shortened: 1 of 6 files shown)


src/server/lesson-controller.js
```diff
@@ -37,6 +37,10 @@ export default class LessonController {
     const { lessonId } = req.params;
     const routeWildcardValue = urls.removeLeadingSlash(req.params['0']);
 
+    if (!user) {
+      return res.redirect(urls.getLoginUrl(req.path));
```
ahelmberger, line 41: We usually do not manually redirect to the login page because the whole logic for redirecting and navigating back to the original page is already implemented in the global error handler. If authentication is required we should throw a `new Unauthorized()` (with "Unauthorized" actually meaning status 401).

Having that said, this check has to come further down (to around line 58) as it is only required to be authenticated when navigating to a private room, but this clause here prevents anonymous users from accessing any room, including public ones (as far as I can see).

src/server/room-controller.js
```diff
@@ -46,6 +46,10 @@ export default class RoomController {
     const { user } = req;
     const { token } = req.params;
 
+    if (!user) {
```
ahelmberger, line 49: This can never be the case (unless passport has a bug) as this handler is protected by our `needs-permission-middleware.js` which in turn calls `req.isAuthenticated()` which in turn already does the same check: https://github.com/jaredhanson/passport/blob/master/lib/http/request.js#L69

src/server/room-controller.js
```diff
@@ -118,9 +122,13 @@ export default class RoomController {
 
   async handleGetRoomPage(req, res) {
     const { roomId } = req.params;
-    const { _id: userId } = req.user;
+    const userId = req.user?._id;
     const routeWildcardValue = urls.removeLeadingSlash(req.params['0']);
 
+    if (!userId) {
```
ahelmberger, line 128: This should also never happen as this page is only server after the `needsPermission(permissions.OWN_ROOMS)` middleware.

Here, I think it is also the case that we cannot use this middleware as we have to allow unauthenticated users to access public rooms.

src/server/lesson-controller.js
adrianaluchian: Then let's revisit our access requirements for rooms and lessons because:
- both now use the permissions middleware, which means none can be accessed by anonymous users
- controllers for both pages reach the `user._id` part of the code if the session is deleted (which indeed should not happen having the permissions middleware)
- I was under the possibly false impression that rooms (and thus lessons as well), even if public, they would require authentication (reason for which we had permissions requirements on rooms already)

src/server/room-controller.js
adrianaluchian: Please see comment here 👉 https://github.com/educandu/educandu/pull/189#discussion_r793389400

src/server/room-controller.js
adrianaluchian: Please see comment here 👉 https://github.com/educandu/educandu/pull/189#discussion_r793389400

ahelmberger (approved):

PR #195, opened 2022-02-01
EDU-306 Open new doc page in edit mode

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-306


src/components/default-page-template.js
```diff
@@ -44,24 +52,6 @@ function DefaultPageTemplate({ children, fullScreen, headerActions, alerts }) {
...
-          disabled={!!action.disabled}
-          icon={<Icon component={action.icon} />}
-          onClick={action.handleClick}
-          >
-          {action.text}
-        </Button>
-      </Restricted>
-    ));
-  }
-
```
ahelmberger, line 64: So nice :smile:

src/components/default-page-template.less
```diff
@@ -47,13 +47,6 @@
     justify-content: flex-start;
   }
 
-  &.DefaultPageTemplate-headerContent--center {
-    flex: none;
-    display: flex;
-    align-items: center;
-    justify-content: center;
-  }
-
```
ahelmberger, line 56: Now that the center is gone, we can give the left side more room (now it has max the same as the right side, but the right side only has two buttons)

adrianaluchian: I've checked this and I feel it's ok, now they each take half, growing to however much each needs when the room shrinks:

Screen Shot 2022-02-01 at 10 42 47

Or did I get this wrong? 🤔

ahelmberger: The left can never take more than 50% if both have a flex grow of 1, in case of OMA that gave me a headache on mobile as the name is very long and the left side would need much more space than the right.

adrianaluchian: Simplified as discussed 👍

ahelmberger (approved):

PR #207, opened 2022-02-04
EDU-320 Fix error when restoring sections after they were hard-deleted

ahelmberger:
Fixes https://educandu.atlassian.net/browse/EDU-320

### DIFF
--- a/src/services/section-helper.js
+++ b/src/services/section-helper.js
@@ -1,0 +1,112 @@
+import deepEqual from 'fast-deep-equal';
+import Logger from '../common/logger.js';
+import uniqueId from '../utils/unique-id.js';
+import cloneDeep from '../utils/clone-deep.js';
+
+const logger = new Logger(import.meta.url);
+
+function checkPlausibility({ section, ancestorSection = null }) {
+  if (!section.key || !section.type) {
+    throw new Error('Section key and type must be specified');
+  }
+
+  if (ancestorSection && (!ancestorSection.key || !ancestorSection.type)) {
+    throw new Error('Ancestor section key and type must be specified');
+  }
+
+  if (ancestorSection && ancestorSection.type !== section.type) {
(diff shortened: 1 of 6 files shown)


src/services/section-helper.js
```diff
@@ -0,0 +1,112 @@
...
+        return cloneDeep(ancestorSection);
+      }
+
+      // Otherwise we start a new revision:
+      logger.info(`Creating new revision for section key ${section.key}`);
+      return {
+        revision: uniqueId.create(),
+        key: section.key,
+        deletedOn: cloneDeep(section.deletedOn),
+        deletedBy: cloneDeep(section.deletedBy),
```
adrianaluchian, line 52: These values don't need cloning, right?

src/services/section-helper.js
```diff
@@ -0,0 +1,112 @@
...
+  logger.info(`Creating new revision for section key ${section.key}`);
+  return {
+    revision: uniqueId.create(),
+    key: section.key,
+    deletedOn: null,
+    deletedBy: null,
+    deletedBecause: null,
+    type: section.type,
+    content: cloneDeep(section.content)
+  };
```
adrianaluchian, line 111: As this new section creation is done twice, we could extract it in a function. With this we could reduce a bit the comments (lines 76 and 101), as it will be self explanatory through the function name.

src/services/section-helper.js
ahelmberger: Not the `deletedBy`, and for `deletedOn` we can only take `new Date(section.deletedOn.getTime())`, will do.

ahelmberger: Done

src/services/section-helper.js
ahelmberger: True, I extracted all three similar cases into a function

adrianaluchian (approved):

PR #214, opened 2022-02-08
EDU-278 Fix textarea cursor jumping in confirmation dialog

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-278

### DIFF
--- a/src/components/confirmation-dialogs.js
+++ b/src/components/confirmation-dialogs.js
@@ -80,12 +80,12 @@ export function confirmCdnFileDelete(t, fileName, onOk, onCancel = () => {}) {
 
 export function confirmSectionHardDelete(
   t,
-  formRef,
   onOk,
   onCancel = () => {}
 ) {
   let dialog = null;
   let createDialogProps = null;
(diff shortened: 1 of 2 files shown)


src/components/confirmation-dialogs.js
```diff
@@ -79,21 +80,20 @@ export function confirmCdnFileDelete(t, fileName, onOk, onCancel = () => {}) {
 
 export function confirmSectionHardDelete(
   t,
+  formRef,
```
ahelmberger, line 83: I don't think you have to hand in the ref into this function. You could just call `const formRef = React.createRef()` inside the function. But technically speaking, the ref to hand into a Antd form has to be created by `Form.useForm()` not `React.useRef()` (because I guess it puts some special properties onto the ref). But obviously it also works without :no_good_man:

ahelmberger: True, you handed it directly to the `ref` prop, so this is kind of the "raw" approach to get the form instance.

ahelmberger (approved):

PR #215, opened 2022-02-08
EDU-328 Convert user inputs to functional components

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-328


src/components/email-input.js
```diff
@@ -1,66 +1,64 @@
...
+      required: true,
+      message: t('enterEmail')
+    },
+    {
+      type: 'email',
+      message: t('emailIsInvalid')
+    },
+    {
+      validator: (rule, value) => {
+        return value && forbiddenEmails.includes(value.toLowerCase())
```
ahelmberger, line 38: Not related to the ticket, but it would be nice to have here `normalize(value)`, so it is consistent with the form item's normalization function (although it's pretty straight-forward)

src/components/username-input.js
```diff
@@ -1,58 +1,55 @@
...
+      validator: (rule, value) => {
+        const minLength = 6;
+        return value && value.trim().length < minLength
+          ? Promise.reject(new Error(t('usernameIsTooShort', { length: minLength })))
+          : Promise.resolve();
+      }
+    },
+    {
+      validator: (rule, value) => {
+        return value && forbiddenUsernames.includes(value.toLowerCase())
```
ahelmberger, line 30: Also not related to the ticket, but usernames are case-sensitive, we are too strict here. But maybe that's a good thing, what do you think? I don't think we should change it in this PR, just to keep it in mind.

src/components/email-input.js
adrianaluchian: Sure, makes sense, change to use the function 👍

src/components/username-input.js
adrianaluchian: Interesting, that's true, usernames are case sensitive. It might be a good thing to keep this restriction here though, so that only the person that has selected a certain username can use the variations of it (given by changing character casings) when they would want to change their username.

It also avoids the confusion that would come with more users having the same username with different casing 🤔

ahelmberger: Yep, agree.

ahelmberger (approved):

PR #216, opened 2022-02-08
EDU-323 Fix bugs in quick tester and ear training editors

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-323

### DIFF
--- a/src/plugins/ear-training/ear-training.yml
+++ b/src/plugins/ear-training/ear-training.yml
@@ -10,9 +10,6 @@ solve:
 reset:
   en: Reset
   de: Zurücksetzen
-titleLabel:
-  en: Title
-  de: Titel
 startAbcCode:
   en: Start ABC code
   de: Vorgabe-ABC-Code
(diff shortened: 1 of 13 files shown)


src/plugins/ear-training/ear-training.yml
```diff
@@ -10,6 +10,9 @@ solve:
 reset:
   en: Reset
   de: Zurücksetzen
+titleLabel:
+  en: Title
+  de: Titel
```
adrianaluchian, line 15: Hint hint, there's a title in `common.yml`

adrianaluchian:
In the quick tester, I can't select "random" display order (the hover effect is shown, but the radio value does not change). However, the answers are always shown in random order 🤔

src/plugins/ear-training/ear-training.yml
ahelmberger: Took that one (also in quick-tester), thanks

adrianaluchian (approved):

PR #226, opened 2022-02-10
EDU-336 Set loading state on edit control panel save button

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-336


ahelmberger:
Feels very good now, even with slow network!

PR #233, opened 2022-02-15
EDU-346 Fix height for CDN file picker

ahelmberger:
Fixes https://educandu.atlassian.net/browse/EDU-346

### DIFF
--- a/src/styles/global-variables.less
+++ b/src/styles/global-variables.less
@@ -16,8 +16,8 @@
 // Layering
 @edu-z-index-base-layer: 0;
 @edu-z-index-cover-layer: 1;
-@edu-z-index-overlay-layer: 1000;
-@edu-z-index-sticky-content-layer: 1001;
+@edu-z-index-overlay-layer: 900;
+@edu-z-index-sticky-content-layer: 901;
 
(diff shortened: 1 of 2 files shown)


src/styles/global-variables.less
```diff
@@ -16,8 +16,8 @@
 // Layering
 @edu-z-index-base-layer: 0;
 @edu-z-index-cover-layer: 1;
-@edu-z-index-overlay-layer: 1000;
-@edu-z-index-sticky-content-layer: 1001;
+@edu-z-index-overlay-layer: 900;
+@edu-z-index-sticky-content-layer: 901;
```
adrianaluchian, line 20: Should we maybe extract our own `@edu-z-index-modal-layer` and then use this to set the ant design modal index? 
Maybe in the future we introduce the bug if we create another layer here and forget this situation with the ant modal 🤔

adrianaluchian: I see. Then maybe we should just mention in our z-index section that ant-design layers start at 1000 and our maximum should be 999, or something like that. What do you think?

adrianaluchian (approved):

PR #238, opened 2022-02-16
EDU-352 Wrap CDN props in public storage object

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-352

### DIFF
--- a/src/components/sections-display.js
+++ b/src/components/sections-display.js
@@ -3,14 +3,14 @@ import { Button, Divider } from 'antd';
 import { PlusOutlined } from '@ant-design/icons';
 import React, { Fragment, useState } from 'react';
 import SectionDisplay from './section-display.js';
-import { sectionShape } from '../ui/default-prop-types.js';
 import PluginSelectorDialog from './plugin-selector-dialog.js';
+import { sectionShape, storageShape } from '../ui/default-prop-types.js';
 import { DragDropContext, Droppable, Draggable } from 'react-beautiful-dnd';
 
 function SectionsDisplay({
   sections,
-  sectionsContainerId,
   pendingSectionKeys,
+  publicStorage,
   canEdit,
   canHardDelete,
   onPendingSectionApply,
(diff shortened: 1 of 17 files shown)


src/ui/default-prop-types.js
```diff
@@ -283,3 +282,9 @@ export const lessonMetadataShape = PropTypes.shape({
 export const lessonShape = PropTypes.shape({
   ...lessonMetadataProps
 });
+
+export const storageShape = PropTypes.shape({
```
ahelmberger, line 286: I think we have to be more specific with the naming to indicate that this is the config for the file picker. We will also have a storage shape on the user (the `storage` property) and a storage plan shape.

adrianaluchian: Indeed, renamed to `filePickerStorageShape`, thanks!

src/components/sections-display.js
```diff
@@ -3,14 +3,14 @@ import { Button, Divider } from 'antd';
 import { PlusOutlined } from '@ant-design/icons';
 import React, { Fragment, useState } from 'react';
 import SectionDisplay from './section-display.js';
-import { sectionShape } from '../ui/default-prop-types.js';
 import PluginSelectorDialog from './plugin-selector-dialog.js';
+import { sectionShape, filePickerStorageShape } from '../ui/default-prop-types.js';
```
ahelmberger, line 7: This is kinda "unorderly" :wink:

src/plugins/ear-training/editing/ear-training-sound-editor.js
```diff
@@ -4,14 +4,15 @@ import { Input, Radio } from 'antd';
 import { SOUND_TYPE } from '../constants.js';
 import { useTranslation } from 'react-i18next';
 import ClientConfig from '../../../bootstrap/client-config.js';
+import { filePickerStorageShape } from '../../../ui/default-prop-types.js';
```
ahelmberger, line 7: This too

ahelmberger (approved):

src/components/sections-display.js
adrianaluchian: 😱 Na das kann nicht sein! Korregiert!

ahelmberger: Thanks :sweat_smile:

PR #239, opened 2022-02-16
EDU-356 Show storage tab on MySpace page

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-356

Decided to show this in a separate tab after all, due to the current form layout we have on the account tab.

_Example with < 95% occupied storage:_

Screen Shot 2022-02-16 at 17 07 37

_Example with >= 95% occupied storage:_

Screen Shot 2022-02-16 at 17 07 59

### DIFF
--- a/src/components/pages/my-space.js
+++ b/src/components/pages/my-space.js
@@ -4,20 +4,23 @@ import PropTypes from 'prop-types';
 import RoomsTab from '../rooms-tab.js';
 import ProfileTab from '../profile-tab.js';
 import AccountTab from '../account-tab.js';
+import { useUser } from '../user-context.js';
+import UsedStorage from '../used-storage.js';
 import { useTranslation } from 'react-i18next';
 import { useService } from '../container-context.js';
-import { roomShape } from '../../ui/default-prop-types.js';
 import ClientConfig from '../../bootstrap/client-config.js';
 import { useGlobalAlerts } from '../../ui/global-alerts.js';
+import { roomShape, storagePlanShape } from '../../ui/default-prop-types.js';
 
 const { TabPane } = Tabs;
 
 function MySpace({ initialState, PageTemplate }) {
+  const user = useUser();
   const alerts = useGlobalAlerts();
   const { t } = useTranslation('mySpace');
   const clientConfig = useService(ClientConfig);
 
-  const { rooms } = initialState;
+  const { rooms, storagePlan } = initialState;
 
   const formItemLayout = {
     labelCol: {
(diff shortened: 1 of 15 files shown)


src/components/pages/my-space.js
```diff
@@ -59,6 +62,13 @@ function MySpace({ initialState, PageTemplate }) {
           <TabPane className="Tabs-tabPane" tab={t('accountTabTitle')} key="3">
             <AccountTab formItemLayout={formItemLayout} tailFormItemLayout={tailFormItemLayout} />
           </TabPane>
+          {user.storage && (
+            <TabPane className="Tabs-tabPane" tab={t('common:storage')} key="4">
+              <h5>{`"${storagePlan.name}" ${t('storagePlanLabel')}`}</h5>
```
ahelmberger, line 67: I think at a point in the near future (when we can remove storage plans from users), there might be the scenario where:
* `storagePlan` is `null`
* `user.storage` is smth like `{ plan: null, usedStorageInBytes: 184637, reminders: [] }`

I'd suggest to show something like "No storage plan" here in case `storagePlan` is not defined.

src/components/used-storage.js
```diff
@@ -0,0 +1,46 @@
...
+  if (kilobytes < 1000) {
+    return `${kilobytes} KB`;
+  }
+  const megabytes = kilobytes / 1000;
+  if (megabytes < 1000) {
+    return `${megabytes} MB`;
+  }
+  const gigabytes = megabytes / 1000;
+  return `${gigabytes} GB`;
+}
```
ahelmberger, line 16: We have the module `pretty-bytes` which does "pretty" much this (it's used for example in the CDN file browser).

src/components/used-storage.js
```diff
@@ -0,0 +1,46 @@
...
+
+function UsedStorage({ usedBytes, maxBytes }) {
+  const format = percent => `${percent} %`;
+  const percent = usedBytes * 100 / maxBytes;
```
ahelmberger, line 20: We might want to take two edge cases into account here:
1. `maxBytes` is `0` then we would divide by zero
2. The scenario I described above (a user had private space and uploaded some files, then an admin takes unassigns the plan from the user), then the percentage will be higher than 100%, so we probably should use `Math.min(100, percent)` here

ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-356
> 
> Decided to show this in a separate tab after all, due to the current form layout we have on the account tab.
> 
> _Example with < 95% occupied storage:_
> 
> Screen Shot 2022-02-16 at 17 07 37
> 
> _Example with >= 95% occupied storage:_
> 
> Screen Shot 2022-02-16 at 17 07 59

Looks really nice :heart_eyes:

src/components/used-storage.js
adrianaluchian: I was under the impression this is the only case where we show byte size to the user and didn't want to introduce a library for that.
Cool, used `pretty-bytes` 👍

src/components/used-storage.js
adrianaluchian: Didn't consider covering this case since we don't have it yet, but it's not difficult and better be safe indeed.
Done, thanks!

src/components/pages/my-space.js
adrianaluchian: Treated `storagePlan === null`, thanks

adrianaluchian:
> Looks really nice 😍
Thanks, looks a tad better now with the progress bar label on the same line and the right-side overflow issue fixed 😬

ahelmberger (approved):

PR #240, opened 2022-02-16
EDU-353 Pass privateStorage down to repository browser

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-353

### DIFF
--- a/src/components/audio-player.js
+++ b/src/components/audio-player.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { inject } from './container-context.js';
+import { useService } from './container-context.js';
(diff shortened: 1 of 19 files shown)


src/components/audio-player.js
```diff
@@ -28,6 +28,4 @@ AudioPlayer.defaultProps = {
   soundUrl: ''
 };
 
-export default inject({
```
adrianaluchian, line 31: Uh, leftover ..

adrianaluchian (approved):

src/components/audio-player.js
ahelmberger: Thanks for leaving me one :smile:

PR #241, opened 2022-02-17
EDU-358 Add storage information to users page

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-358

### DIFF
--- a/src/api-clients/user-api-client.js
+++ b/src/api-clients/user-api-client.js
@@ -116,7 +116,7 @@ class UserApiClient {
       .then(res => res.data);
   }
 
-  removeUserStorageReminders({ userId }) {
+  deleteAllUserStorageReminders({ userId }) {
     return this.httpClient
       .delete(
         `/api/v1/users/${encodeURIComponent(userId)}/storageReminders`,
--- a/src/api-clients/user-api-client.js
+++ b/src/api-clients/user-api-client.js
@@ -105,6 +105,25 @@ class UserApiClient {
       )
       .then(res => res.data);
   }
+
+  addUserStorageReminder({ userId }) {
+    return this.httpClient
+      .post(
+        `/api/v1/users/${encodeURIComponent(userId)}/storageReminders`,
+        null,
+        { responseType: 'json' }
+      )
+      .then(res => res.data);
+  }
+
+  removeUserStorageReminders({ userId }) {
+    return this.httpClient
+      .delete(
+        `/api/v1/users/${encodeURIComponent(userId)}/storageReminders`,
+        { responseType: 'json' }
+      )
+      .then(res => res.data);
+  }
 }
 
 export default UserApiClient;
--- a/src/components/pages/my-space.less
+++ b/src/components/pages/my-space.less
@@ -11,3 +11,8 @@
 .MySpacePage-usedStorageLabel {
   margin-right: 15px;
 }
+
+.MySpacePage-usedStorageBar {
+  margin: 10px 0;
+  flex: 0 1 100%;
+}
--- a/src/components/pages/users.js
+++ b/src/components/pages/users.js
@@ -1,20 +1,25 @@
-import firstBy from 'thenby';
+import by from 'thenby';
 import PropTypes from 'prop-types';
 import prettyBytes from 'pretty-bytes';
-import React, { useState } from 'react';
 import Logger from '../../common/logger.js';
+import UsedStorage from '../used-storage.js';
+import { useUser } from '../user-context.js';
 import { useTranslation } from 'react-i18next';
 import { ROLE } from '../../domain/constants.js';
-import { useLocale } from '../locale-context.js';
 import errorHelper from '../../ui/error-helper.js';
-import { Table, Popover, Tabs, Select } from 'antd';
+import React, { useEffect, useState } from 'react';
+import { replaceItem } from '../../utils/array-utils.js';
 import UserRoleTagEditor from '../user-role-tag-editor.js';
 import { useGlobalAlerts } from '../../ui/global-alerts.js';
 import { getDefaultStorage } from '../../domain/storage.js';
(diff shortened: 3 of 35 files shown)


src/components/pages/my-space.less
```diff
@@ -11,3 +11,8 @@
 .MySpacePage-usedStorageLabel {
   margin-right: 15px;
 }
+
+.MySpacePage-usedStorageBar {
+  margin: 10px 0;
```
adrianaluchian, line 16: It would be better to put this margin in the usage side of the component, cause I'm adding this component to the repository browser and it would create too much space 😬

src/components/pages/users.js
```diff
@@ -26,38 +31,201 @@ const availableRoles = Object.values(ROLE);
...
     } else {
-      internalUsers.push(user);
+      internalUsers.push(enrichedUserObject);
     }
-  }
 
-  return { internalUsers, externalUsers };
-}
+    if (user.storage || user.storagePlan) {
+      storageUsers.push(enrichedUserObject);
```
adrianaluchian, line 59: Should we maybe decide here based on the occupied storage value instead of the `storagePlan`?
Basically any case in which a user occupies storage space, so nothing falls through the cracks.

src/components/pages/users.js
```diff
@@ -26,38 +31,201 @@ const availableRoles = Object.values(ROLE);
...
+        }
+        : null;
+
+      setIsSaving(true);
+      setUsers(oldUsers => replaceItem(oldUsers, { ...user, storage: newStorage }));
+
+      let finalStorage;
+      try {
+        await roomApiClient.deleteAllPrivateRoomsForUser({ ownerId: user._id });
+        finalStorage = await userApiClient.removeUserStorageReminders({ userId: user._id });
```
adrianaluchian, line 219: In the next step regarding storage, I'm thinking we might want to do the room deletion in a transaction 🤔 Maybe we can introduce that when we introduce the user lock.

src/server/room-controller.js
```diff
@@ -195,6 +214,12 @@ export default class RoomController {
       (req, res) => this.handlePatchRoom(req, res)
     );
 
+    router.delete(
+      '/api/v1/rooms',
+      [needsPermission(permissions.OWN_ROOMS), validateQuery(deleteRoomsQuerySchema)],
```
adrianaluchian, line 219: I think we need to use another permission here, one that is only allocated to admins (`OWN_ROOMS` is a permission normal users have as well)

src/server/room-controller.spec.js
```diff
@@ -649,6 +650,67 @@ describe('room-controller', () => {
...
+        protocol: 'https',
+        headers: { host: 'educandu.dev' },
+        query: { ownerId: user._id, access: ROOM_ACCESS_LEVEL.private }
+      });
+      req.user = user;
+
+      res = httpMocks.createResponse({ eventEmitter: EventEmitter });
+      res.on('end', done);
+
+      roomService.getRoomsOwnedByUser.resolves([roomA, roomB]);
```
adrianaluchian, line 672: It's a little confusing that here this resolves with the users of `userJacky` but the `req.query.onwerId` is not set up to the same user.
I think it would be good to include `withArgs` here as well, to resolve per user. That would basically point to the correction needed in the `req` setup.

src/components/pages/users.js
ahelmberger: Yes, I also think this should happen within one call/transaction, but at the moment we manage the rooms with the `roomClient/Service` and the storage reminders with `userClient/Service`. I think we already agreed to factor out a `storageClient/Service` in order to disentangle these dependencies.

adrianaluchian: Yes exactly, I only mentioned thi to we keep it in mind as well.

src/components/pages/users.js
ahelmberger: You are right, but even worse, the second condition makes no sense at all, it can never be true while the first one is false. I changed it to `user.storage?.plan || user.storage?.usedStorageInBytes`, so nothing can fall through the cracks now.

adrianaluchian: Oh right, there is no `user.storagePlan`, there's `user.storage.plan`

src/api-clients/user-api-client.js
```diff
@@ -105,6 +105,25 @@ class UserApiClient {
...
+    return this.httpClient
+      .post(
+        `/api/v1/users/${encodeURIComponent(userId)}/storageReminders`,
+        null,
+        { responseType: 'json' }
+      )
+      .then(res => res.data);
+  }
+
+  removeUserStorageReminders({ userId }) {
```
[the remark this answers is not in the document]
ahelmberger: Yes, good idea. I renamed all corresponding functions (client, controller and service) to follow the pattern.

src/server/room-controller.js
ahelmberger: But I was under the impression that every user should be able to delete their own rooms?

adrianaluchian: But they do that room by room, right? That's basically the `DELETE /api/v1/rooms/:roomId` endpoint.
I thought this endpoint is for an admin to delete all private rooms of a user (since we pass "ownerId"). For cases where users should delete their own rooms, we should check that the `req.user` is the owner.

ahelmberger: Done, I also renamed the handler to `handleDeleteAllRoomsForUser` in order to clearly distinguish it from the "normal" delete handler.

src/server/room-controller.spec.js
ahelmberger: I added `withArgs`, and also detected that it should resolve to all three rooms (I added `roomC` later), as they are all owned by `dagobert-the-third`

adrianaluchian: This guy, he owns too much 🤔

adrianaluchian (approved):

PR #247, opened 2022-02-22
EDU-360 Change user default storage from null to object with default data

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-360


src/components/pages/users.js
```diff
@@ -55,7 +54,7 @@ function createUserSubsets(users, storagePlans) {
       internalUsers.push(enrichedUserObject);
     }
 
-    if (user.storage?.plan || user.storage?.usedBytes) {
+    if (user.storage.plan || user.storage.usedBytes) {
```
ahelmberger, line 57: Actually, it also should be shown if there are (for whatever reason) reminders:

`if (user.storage.plan || user.storage.usedBytes || user.storage.reminders.length) {`

adrianaluchian: Good idea, done!

ahelmberger (approved):

PR #248, opened 2022-02-22
EDU-367 Consolidate locks collections

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-367


src/stores/lock-store-base.js
```diff
@@ -3,20 +3,23 @@ import StoreBase from './store-base.js';
...
-    this.expirationTimeInMinutes = expirationTimeInMinutes;
-  }
+  async takeLock({ type, key, expirationTimeInMinutes }) {
+    const expires = expirationTimeInMinutes ? moment().add(expirationTimeInMinutes, 'minutes').toDate() : null;
+
+    const lock = {
+      _id: uniqueId.create(),
+      type,
+      key,
+      sessionKey: uniqueId.create(),
```
adrianaluchian, line 13: I would like to delete this `sessionKey`, as I can't find any usage of it. Do you remember how this was used?

ahelmberger: I think it was once planned as a means to correlate eventual lingering locks to a certain action, but I think we can delete it, the system has proven reliable so it will never be needed, I guess.

adrianaluchian: Ok I see, thanks for the insight 👍 Removed!

src/stores/lock-store-base.js
```diff
@@ -3,20 +3,23 @@ import StoreBase from './store-base.js';
 import uniqueId from '../utils/unique-id.js';
 
 class LockStoreBase extends StoreBase {
```
ahelmberger, line 5: As this class is now base class to only one remaining lock store, I think we do not need it anymore, we could just pull up all members into the actual `LockStore`.

adrianaluchian: I was thinking if I should do that or keep the separation 🤔 Ok, will merge it into the `LockStore`.

adrianaluchian: Merged, thanks!

ahelmberger (approved):

PR #250, opened 2022-02-22
EDU-369 Push data access logic in room service down to room (invitation) store

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-369

### DIFF
--- a/src/services/room-service.js
+++ b/src/services/room-service.js
@@ -17,13 +17,6 @@ const logger = new Logger(import.meta.url);
 
 const PENDING_ROOM_INVITATION_EXPIRATION_IN_DAYS = 7;
 
-const roomInvitationProjection = {
-  _id: 1,
-  email: 1,
-  sentOn: 1,
-  expires: 1
-};
-
 export default class RoomService {
   static get inject() {
     return [RoomStore, RoomInvitationStore, LessonStore, LockStore, UserService, StorageService, TransactionRunner];
@@ -40,33 +33,20 @@ export default class RoomService {
   }
 
   getRoomById(roomId) {
-    return this.roomStore.findOne({ _id: roomId });
+    return this.roomStore.getRoomById({ roomId });
   }
 
-  async _getRooms({ ownerId, memberId }) {
-    const orFilters = [];
-
-    if (ownerId) {
-      orFilters.push({ owner: ownerId });
-    }
-    if (memberId) {
-      orFilters.push({ members: { $elemMatch: { userId: memberId } } });
-    }
-
-    const filter = orFilters.length === 1 ? orFilters[0] : { $or: orFilters };
-
-    const rooms = await this.roomStore.find(filter);
-    return rooms;
+  getRoomsOwnedByUser(userId) {
+    return this.roomStore.getRoomsByOwnerId({ ownerId: userId });
   }
 
-  async getRoomsOwnedByUser(userId) {
-    const rooms = await this._getRooms({ ownerId: userId });
-    return rooms;
+  getRoomsOwnedOrJoinedByUser(userId) {
+    return this.roomStore.getRoomsOwnedOrJoinedByUser({ userId });
   }
 
-  async getRoomsOwnedOrJoinedByUser(userId) {
-    const rooms = await this._getRooms({ ownerId: userId, memberId: userId });
-    return rooms;
+  async isRoomOwnerOrMember(roomId, userId) {
+    const room = await this.roomStore.getRoomsByIdOwnedOrJoinedByUser({ roomId, userId });
+    return !!room;
   }
 
   async createRoom({ name, slug, access, user }) {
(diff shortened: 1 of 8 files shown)


src/services/room-service.js
```diff
@@ -40,33 +33,20 @@ export default class RoomService {
...
-    if (memberId) {
-      orFilters.push({ members: { $elemMatch: { userId: memberId } } });
-    }
-
-    const filter = orFilters.length === 1 ? orFilters[0] : { $or: orFilters };
-
-    const rooms = await this.roomStore.find(filter);
-    return rooms;
+  getRoomsOwnedByUser(userId) {
+    return this.roomStore.getRoomsByOwnerId({ ownerId: userId });
```
adrianaluchian, line 40: Same here, we can simplify.

src/services/room-service.js
```diff
@@ -96,20 +76,45 @@ export default class RoomService {
     return updatedRoom;
   }
 
-  async findOwnedRoomById({ roomId, ownerId }) {
+  async deleteRoom(roomId, user) {
     const room = await this.roomStore.findOne({ _id: roomId });
```
adrianaluchian, line 80: I think we should use `this.roomStore.getRoomById` here.

src/services/room-service.js
```diff
@@ -40,33 +33,20 @@ export default class RoomService {
   }
 
   getRoomById(roomId) {
-    return this.roomStore.findOne({ _id: roomId });
+    return this.roomStore.getRoomById({ roomId });
```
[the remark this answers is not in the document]
ahelmberger: I always wrapped all arguments of the store functions in an object, as the second param is an object that contains the optional `session` for using it in a transaction. Don't know, I thought this might be a good pattern (similar to how mongodb does the find, update, delete calls. What do you think?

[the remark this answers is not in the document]
adrianaluchian: Or maybe better, `getRoomById({ roomId, session })` instead of  `getRoomById({ roomId }, { session })`?

src/services/room-service.js
```diff
@@ -96,20 +76,45 @@ export default class RoomService {
...
+    }
+
+    if (room.owner !== user._id) {
+      throw new Forbidden();
+    }
+
+    await this.transactionRunner.run(async session => {
+      await this.lessonStore.deleteMany({ roomId }, { session });
+      await this.roomInvitationStore.deleteMany({ roomId }, { session });
+      await this.roomStore.deleteOne({ _id: roomId }, { session });
```
adrianaluchian, line 93: For me it would be nicer if this filter mapping (`{ _id: roomId }`) would be done by the store, and the store would expose a `deleteRoomById({ roomId, session })` method.
Then one by one we could do the same for the other stores used in here too.

ahelmberger: Sounds good to me, I also thought about it, but was not sure how far we wnat to go with creating specialized functions, but I do like to have these `...ById` methods. I will add it.

src/services/room-service.js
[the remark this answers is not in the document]
adrianaluchian: I honestly would keep the 'mongo' logic inside the store and expose something that suits us best for the store users. To me the store doesn't need to copy the mongo API, I see it as a wrapper.

src/stores/room-store.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+describe('room-store', () => {
+  let sut;
+  let myUser;
+  let container;
+  let otherUser;
+  let onlyJoiningUser;
+
+  const now = new Date();
+  const sandbox = sinon.createSandbox();
```
adrianaluchian, line 14: This guy is not needed here 🤔

src/services/room-service.js
ahelmberger: Done, I also created special store functions for most use cases (except `save`)

src/stores/room-store.spec.js
ahelmberger: True, the guy is removed.

src/services/room-service.js
ahelmberger: Update: All generic functions are not used anymore and `RoomStore` and `RoomInvitationStore` do not inherit from `StoreBase` anymore.

src/stores/room-invitation-store.js
```diff
@@ -8,31 +7,35 @@ const roomInvitationMetadataProjection = {
   expires: 1
 };
 
-class RoomInvitationStore extends StoreBase {
+class RoomInvitationStore {
```
adrianaluchian, line 10: Good idea to not extend `StoreBase` anymore 👍

adrianaluchian (approved):

src/stores/room-invitation-store.js
ahelmberger: Yeah, we can throw it away after we've updated all stores.

PR #252, opened 2022-02-22
EDU-368 UserService: push data access down to store

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-368

### DIFF
--- a/src/services/client-data-mapping-service.js
+++ b/src/services/client-data-mapping-service.js
@@ -2,14 +2,13 @@ import uniqueId from '../utils/unique-id.js';
 import cloneDeep from '../utils/clone-deep.js';
 import UserStore from '../stores/user-store.js';
 import privateData from '../domain/private-data.js';
-import UserService from '../services/user-service.js';
+import { extractUserIdsFromDocsOrRevisions } from '../domain/data-extractors.js';
 
-class ClientDataMapper {
-  static get inject() { return [UserStore, UserService]; }
+class ClientDataMappingService {
+  static get inject() { return [UserStore]; }
 
-  constructor(userStore, userService) {
+  constructor(userStore) {
     this.userStore = userStore;
-    this.userService = userService;
   }
 
   mapWebsiteUser(user) {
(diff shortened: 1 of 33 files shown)


src/stores/user-store.js
```diff
@@ -1,11 +1,58 @@
 import Database from './database.js';
-import StoreBase from './store-base.js';
-
-class UserStore extends StoreBase {
+class UserStore {
```
ahelmberger, line 2: A blank line here would be nice :-)

ahelmberger:
When I read the PR I realized that we had a (slightly) different vision: In https://github.com/educandu/educandu/pull/250 the controller still only use services and services only use stores (and other services, which we don't like), and the other way around, stores are only used by services, and services are only used by controllers or similar high-level logic modules.
In this PR, stores are now used from all higher layers, even `index.js`. That leaves me somewhat undecided if it's ok for me or not, I even started reading other people's opinions on it (e.g. https://stackoverflow.com/questions/16211858/in-multi-layered-architecture-can-i-skip-the-business-layer-for-crud-operations#16212580). Maybe we could have a short discussion to get on the same page and then either you or me will adjust, because now we already have two diverging solutions?

src/stores/user-store.spec.js
```diff
@@ -0,0 +1,186 @@
...
+    });
+  });
+
+  describe('findUserByUsernameOrEmail', () => {
+    // BeforeEach(async () => {
+    //   await Promise.all([
+    //     setupTestUser(container, { username: user.username, email: 'james@jameson.com' }),
+    //     setupTestUser(container, { name: 'minnie', email: 'minnie@mouse.com' })
+    //   ]);
+    // });
```
ahelmberger, line 91: I guess this can be deleted?

src/stores/user-store.js
adrianaluchian: I agree 😄 Added

src/stores/user-store.spec.js
adrianaluchian: Of course, thanks! Removed

adrianaluchian:
> [replied comment] (https://github.com/educandu/educandu/pull/252#issuecomment-1048074541)

In my understanding, we could use stores in any upper layer and never horizontally, and that would solve our circular dependency risk. But I'm also not very happy with using both store and service in a controller.
Regarding the discussion link you added, I think we're in the middle between just needing to proxy the store and actually having business logic in services. 
We could go towards what you described, only using the store layer in the service layer, but try to not just proxy the "dumb"/simple methods and maybe offload a bit more of the business logic into them, from the upper layers.

adrianaluchian:
> > [#252 (comment)](https://github.com/educandu/educandu/pull/252#issuecomment-1048074541)
> 
> https://github.com/educandu/educandu/pull/252#issuecomment-1048565597

Done as discussed!

src/services/client-data-mapping-service.js
```diff
@@ -1,13 +1,14 @@
 import uniqueId from '../utils/unique-id.js';
 import cloneDeep from '../utils/clone-deep.js';
+import UserStore from '../stores/user-store.js';
 import privateData from '../domain/private-data.js';
-import UserService from '../services/user-service.js';
+import { extractUserIdsFromDocsOrRevisions } from '../domain/data-extractors.js';
```
adrianaluchian, line 5: Extracted this out of `user-service` into a simple domain helper, and with this:
- the `client-data-mapping-service` is decoupled from the `user-service`
- the `export-service` is decoupled from the `user-service`

adrianaluchian: The remaining service to service dependencies now are:

Screen Shot 2022-02-23 at 11 35 36

ahelmberger: Nice, this is a good pattern for sharing code without getting into interdependencies. :heart:

ahelmberger (approved):

PR #253, opened 2022-02-22
EDU-365 Add description field to documents

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-371


src/domain/schemas/document-schemas.js
```diff
@@ -83,6 +85,7 @@ export const documentDBSchema = joi.object({
   updatedOn: joi.date().required(),
   updatedBy: idOrKeySchema.required(),
   title: joi.string().required(),
+  description: joi.string().allow('').max(1000).required(),
```
adrianaluchian, line 88: What do you think about having this value in a const somewhere, using it here and also passing it to the translation?
We have other cases like this too (tag validation I think, probably password too), maybe we could change those as well in the future?

src/components/document-metadata-editor.js
```diff
@@ -1,153 +0,0 @@
-import PropTypes from 'prop-types';
```
adrianaluchian, line 1: Huh! Interesting leftover, this component..

src/domain/schemas/document-schemas.js
ahelmberger: True, good idea. I'll extract it.

ahelmberger: Extracted it (and a few others) into `validation-constants.js`

src/components/document-metadata-editor.js
ahelmberger: Yeah, nice surprise, always happy to delete stuff :-)

src/domain/schemas/shared-schemas.js
```diff
@@ -1,9 +1,11 @@
 import joi from 'joi';
+import { slugValidationPattern } from '../validation-constants.js';
 
 export const idOrKeySchema = joi.string().alphanum().min(15).max(30);
+
 export const boolStringSchema = joi.any().valid('true', 'false', true, false);
 
-export const slugSchema = joi.string().allow('').required();
+export const slugSchema = joi.string().pattern(slugValidationPattern).allow('').required();
```
adrianaluchian, line 8: I realize not that this might be problematic on importing old slugs 🤔

src/domain/schemas/document-schemas.js
adrianaluchian: Nice!

src/domain/schemas/shared-schemas.js
ahelmberger: True, I removed it for now and left a comment for re-introducing it together with https://educandu.atlassian.net/browse/EDU-308

adrianaluchian (approved):

PR #254, opened 2022-02-22
EDU-371 Do not propagate click on storage file picker

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-371

### DIFF
--- a/src/components/section-display.js
+++ b/src/components/section-display.js
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { Button, Tooltip } from 'antd';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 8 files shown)


ahelmberger:
Looks really nice! I was still able to toggle between edit/display mode by Ctrl-Click on the black-transparent modal overlay, but I think that's a neglectable problem.

ahelmberger (approved):

adrianaluchian:
> Looks really nice! I was still able to toggle between edit/display mode by Ctrl-Click on the black-transparent modal overlay, but I think that's a neglectable problem.
I noticed that too, but actually to me that's desired behavior, since the grey area click is implemented to dismiss the modal.

PR #263, opened 2022-02-24
EDU-376 Unify email validation

ahelmberger:
Fixes https://educandu.atlassian.net/browse/EDU-376


src/components/username-form-item.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Form, Input } from 'antd';
+import { useTranslation } from 'react-i18next';
+import { minUsernameLength } from '../domain/validation-constants.js';
+
+const FormItem = Form.Item;
+
+function UsernameFormItem({ name, usernamesInUse, ...formItemProps }) {
```
adrianaluchian, line 9: We would need to pass in the initial value here for the case of the account tab, where the username needs to be prefilled

ahelmberger: Thanks for spotting that, it's passed in there now.

adrianaluchian (approved):

PR #265, opened 2022-02-24
EDU-381 Add newly created invitations without page reload

ahelmberger:
Fixes https://educandu.atlassian.net/browse/EDU-381

### DIFF
--- a/src/components/pages/room.js
+++ b/src/components/pages/room.js
@@ -1,3 +1,4 @@
+import by from 'thenby';
 import PropTypes from 'prop-types';
 import Markdown from '../markdown.js';
 import urls from '../../utils/urls.js';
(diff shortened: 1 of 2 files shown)


src/components/pages/room.js
```diff
@@ -67,13 +68,15 @@ export default function Room({ PageTemplate, initialState }) {
...
+  const handleInvitationModalClose = newInvitation => {
     setIsRoomInvitationModalVisible(false);
-
-    if (!wasNewInvitationCreated) {
-      return;
+    if (newInvitation) {
+      setInvitations(currentInvitations => {
+        const invitationsByEmail = new Map(currentInvitations.map(x => [x.email, x]));
+        invitationsByEmail.set(newInvitation.email, newInvitation);
+        return [...invitationsByEmail.values()].sort(by(x => x.sentOn));
```
adrianaluchian, line 77: Nice touch, showing just the lates invitation

ahelmberger: Actually, I show all, but there can never be two with the same email (the server deletes the older one), so that's what I'm doing here also in the frontend.

adrianaluchian (approved):

PR #273, opened 2022-02-25
EDU-388 Support home page template injection

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-388

### DIFF
--- a/src/components/pages/index.js
+++ b/src/components/pages/index.js
@@ -1,14 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Markdown from '../markdown.js';
 import urls from '../../utils/urls.js';
 import SearchBar from '../search-bar.js';
-import { useSettings } from '../settings-context.js';
 import { useGlobalAlerts } from '../../ui/global-alerts.js';
 
-function Index({ PageTemplate, HomePageLogo, SiteLogo }) {
-  const settings = useSettings();
-
+function Index({ HomePageTemplate }) {
   const handleSearch = searchText => {
     window.location = urls.getSearchUrl(searchText);
   };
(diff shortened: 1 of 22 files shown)


src/components/pages/index.js
```diff
@@ -1,47 +1,27 @@
...
+function Index({ HomePageTemplate }) {
   const handleSearch = searchText => {
     window.location = urls.getSearchUrl(searchText);
   };
 
   const alerts = useGlobalAlerts();
 
   return (
-    <PageTemplate alerts={alerts} fullScreen>
+    <HomePageTemplate alerts={alerts} fullScreen>
```
ahelmberger, line 15: I think as there is only one homepage, we do not need the distinction fullscreen=true/false ?

adrianaluchian: Was thinking the same, but if no `HomePageTemplate` is set, we default to `DefaultPageTemplate` which requires `fullScreen`.
However we will pass in a HomePageTemplate anyway from both projects, so then I guess we could just simplify and drop `fullScreen`

adrianaluchian: Scratch that, I got it wrong 🤦‍♀️ Will remove `fullScreen`

ahelmberger: Ah, I see. Then I'll leave that decision to your taste :-)

ahelmberger (approved):

src/components/pages/index.js
ahelmberger: Ok, too late :laughing:

adrianaluchian: 😅 thanks! Updated though!

PR #275, opened 2022-02-26
EDU-392 Remove store base inheritance from non document stores

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-392

### DIFF
--- a/src/stores/batch-store.js
+++ b/src/stores/batch-store.js
@@ -1,11 +1,38 @@
 import Database from './database.js';
-import StoreBase from './store-base.js';
 
-class BatchStore extends StoreBase {
+class BatchStore {
   static get inject() { return [Database]; }
 
   constructor(db) {
-    super(db.batches);
+    this.collection = db.batches;
+  }
+
+  getBatchById(batchId, { session } = {}) {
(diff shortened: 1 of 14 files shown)


src/stores/batch-store.js
```diff
@@ -1,11 +1,38 @@
...
+
+  getUncompletedBatchByTypeAndHost({ batchType, hostName }, { session } = {}) {
+    return this.collection.findOne({ batchType, 'batchParams.hostName': hostName, 'completedOn': null }, { session });
+  }
+
+  getUncompletedBatch({ session } = {}) {
+    return this.collection.findOne({ completedOn: null }, { session });
+  }
+
+  addBatch(batch, { session } = {}) {
```
ahelmberger, line 30: Or should we go with `insertBatch` in order to stay closer to the MongoDB terminology? I know that we translate `get` into `find` and `save` into `upsert`, but `add` sounds very weak to me. Alternatively. I could imaging `createBatch` (like the `C` in `CRUD`). Just a thought, in the end we should just stick to one terminology whichever it is.

adrianaluchian: `createBatch` actually would be better, will change.

PR #278, opened 2022-02-28
EDU-394 Remove store base inheritance from document store

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-394


src/stores/store-helper.js
```diff
@@ -0,0 +1,30 @@
+export const getTagsQuery = searchString => [
+  { $unwind: '$tags' },
+  {
+    $match:
+    {
+      $and: [
+        { tags: { $regex: `.*${searchString}.*`, $options: 'i' } },
+        { slug: { $ne: null } } // Is this still needed?
```
adrianaluchian, line 8: I am wondering if this condition is (still) needed actually

ahelmberger: Let's remove it, it doesn't do anything anyways

ahelmberger (approved):

src/stores/store-helper.js
adrianaluchian: 👍 Removed and reduced tag searching to documents, like discussed

PR #279, opened 2022-03-02
EDU-399 Change heading and general spacing

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-399

With this PR:
- the educandu default styling changed towards reflecting the OMA styling direction
- the `@edu-content-padding` is no longer the variable for the page padding,`@edu-page-content-padding` is used instead
- the `.m-page-content` mixin is to be used on all main page areas (header, content, footer)

These changes give a little bit more freedom to the application using educandu, in what concerns page template styling.

### DIFF
--- a/src/components/default-page-header.less
+++ b/src/components/default-page-header.less
@@ -1,6 +1,5 @@
 .DefaultPageHeader {
   .m-page-content;
-  top: 0;
   flex: none;
   width: 100%;
 }


src/components/default-page-header.less
```diff
@@ -1,27 +1,18 @@
-@default-page-header-content-min-height: 32px;
-@default-page-header-min-height: @default-page-header-content-min-height + (2* @edu-content-padding);
-
 .DefaultPageHeader {
+  .m-page-content;
   top: 0;
```
ahelmberger, line 3: Now that the position is static, we do not need the `top` rule anymore.

ahelmberger (approved):

PR #280, opened 2022-03-02
EDU-405 Include full Lato font family v22

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-405


adrianaluchian:
@ahelmberger, could you please run this locally also?
I am getting a curious error and am wondering if it is a local issue.

PR #281, opened 2022-03-02
EDU-407 Make profile picture round

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-407


src/styles/avatar.less
```diff
@@ -0,0 +1,3 @@
+.Avatar {
+  border: 1px solid @grey-4;
```
ahelmberger, line 2: :+1:

PR #288, opened 2022-03-04
EDU-401 Redesign search page inputs

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-401

and looks like this
Screen Shot 2022-03-04 at 12 01 16

Screen Shot 2022-03-04 at 12 02 06


src/components/pages/search.js
```diff
@@ -96,33 +151,49 @@ function Search({ PageTemplate }) {
...
+            value={null}
+            options={tagOptions}
+            onChange={handleSelectTag}
+            placeholder={(
+              <span className="SearchPage-filterPlaceholder">
+                <span>{`${t('filterPlaceholder')} (${selectedTags.length})`}</span>
+                <SearchOutlined className="SearchPage-filterPlaceholderIcon" />
+              </span>
+            )}
+            suffixIcon={<MenuOutlined />}
```
adrianaluchian, line 172: Once Mani provides the proper icon, we can replace this one.
Sadly we can't get rid of the search icon before it, because it goes out of the box with the option to search within the options list. But I think we can live with it, it makes it more intuitive actually.

src/styles/tag.less
```diff
@@ -0,0 +1,29 @@
...
+  }
+  &:last-of-type {
+    margin-right: 0;
+  }
+
+  &.Tag--selected {
+    color: @edu-white;
+    background-color: @edu-primary-color;
+
+    & > .ant-tag-close-icon {
```
adrianaluchian, line 25: Sadly there is no other way to overwrite ant design's style here, even when we use our own custom icon.

src/components/pages/search.js
adrianaluchian: Changed!

ahelmberger (approved):

PR #289, opened 2022-03-04
EDU-403 Change links styling

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-403

### DIFF
--- a/src/components/account-tab.js
+++ b/src/components/account-tab.js
@@ -77,7 +77,7 @@ function AccountTab({ formItemLayout, tailFormItemLayout }) {
       <UsernameFormItem name="username" usernamesInUse={state.forbiddenUsernames} initialValue={user.username} {...formItemLayout} />
       <EmailFormItem name="email" emailsInUse={state.forbiddenEmails} initialValue={user.email} {...formItemLayout} />
       <FormItem {...tailFormItemLayout}>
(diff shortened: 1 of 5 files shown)


src/components/account-tab.js
```diff
@@ -77,7 +77,7 @@ function AccountTab({ formItemLayout, tailFormItemLayout }) {
       <UsernameFormItem name="username" usernamesInUse={state.forbiddenUsernames} initialValue={user.username} {...formItemLayout} />
       <EmailFormItem name="email" emailsInUse={state.forbiddenEmails} initialValue={user.email} {...formItemLayout} />
       <FormItem {...tailFormItemLayout}>
-        <Button type="link" size="small" onClick={handleResetPasswordClick}>{t('resetPassword')}</Button>
+        <a onClick={handleResetPasswordClick}>{t('resetPassword')}</a>
```
ahelmberger, line 80: Just out of curiosity, why did you change the Button into an a?

ahelmberger (approved):

src/components/account-tab.js
adrianaluchian: I was clicking through the pages to make sure nothing is broken or looks very bad after this styling change and I noticed that this guy was the only one that was not restyled as a link (had no underline). So I streamlined it.

PR #292, opened 2022-03-04
EDU-412 Replace antd icons with equivalents

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-412

### DIFF
--- a/src/components/storage-browser.less
+++ b/src/components/storage-browser.less
@@ -66,13 +66,6 @@
   }
 }
 
-.StorageBrowser-tableDeleteCell {
-  color: @edu-error-color;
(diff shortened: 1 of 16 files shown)


src/components/storage-browser.less
```diff
@@ -68,6 +68,9 @@
 
 .StorageBrowser-tableDeleteCell {
```
ahelmberger, line 69: This class is not used anymore

adrianaluchian: The class that started the extraction of the delete button, of course 🤦‍♀️ 
Removed, thank you!

ahelmberger (approved):

PR #293, opened 2022-03-07
EDU-413 Redesign search results table

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-413

### DIFF
--- a/src/components/items-expander.js
+++ b/src/components/items-expander.js
@@ -10,12 +10,12 @@ function ItemsExpander({ className, expandLinkClassName, items, renderItem, init
 
   return (
     <div className={className}>
-      {items.filter((_tag, index) => isExpanded || (showExpander && index < initialItemCount)).map(renderItem)}
+      {items.filter((_tag, index) => isExpanded || !showExpander || index < initialItemCount).map(renderItem)}
       {showExpander && isExpanded && (
         <a className={expandLinkClassName} onClick={() => setIsExpanded(false)}>{t('collapse')}</a>
       )}
       {showExpander && !isExpanded && (
-        <a className={expandLinkClassName} onClick={() => setIsExpanded(true)}>{t('expand', { count: items.length - 10 })}</a>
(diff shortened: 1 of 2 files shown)


src/components/pages/search.less
```diff
@@ -48,3 +48,34 @@
...
 }
+
+.SearchPage-cellTagsExpandLink {
+  font-size: 0.9em;
+  display: inline-block;
+  margin-left: 0.2em;
+  height: 20px;
+  padding: 3px;
+  margin: 3px;
+  margin-left: 1em;
```
adrianaluchian, line 59: The margins here are a bit confusing, `margin-left` is defined twice and there's a mix of margins in `px` and `em` 🤔

src/components/items-expander.js
```diff
@@ -0,0 +1,38 @@
...
+  const showExpander = items.length > initialItemCount + 1;
+
+  return (
+    <div className={className}>
+      {items.filter((_tag, index) => isExpanded || (showExpander && index < initialItemCount)).map(renderItem)}
+      {showExpander && isExpanded && (
+        <a className={expandLinkClassName} onClick={() => setIsExpanded(false)}>{t('collapse')}</a>
+      )}
+      {showExpander && !isExpanded && (
+        <a className={expandLinkClassName} onClick={() => setIsExpanded(true)}>{t('expand', { count: items.length - 10 })}</a>
```
adrianaluchian, line 18: Here `10` could be replaced with `initialItemCount`, if I get the purpose of this variable right.

ahelmberger: Thanks for spotting that

src/components/pages/search.less
ahelmberger: I replaced it anyway :-)

adrianaluchian: The first `margin-left` could still be removed, since you set `margin` afterwards, right?

ahelmberger: OMG, yes, of course. Removed it!

PR #297, opened 2022-03-08
EDU-420 Allow unassigning storage plans

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-420

### DIFF
--- a/src/components/used-storage.js
+++ b/src/components/used-storage.js
@@ -9,8 +9,8 @@ function UsedStorage({ usedBytes, maxBytes, showLabel }) {
   const { t } = useTranslation('usedStorage');
   const { uiLocale } = useLocale();
 
-  const percent = maxBytes ? usedBytes * 100 / maxBytes : 100;
-  const displayedPercent = `${Math.round(percent)} %`;
+  const percent = maxBytes ? usedBytes * 100 / maxBytes : 101;
+  const displayedPercent = percent > 100 ? '> 100 %' : `${Math.round(percent)} %`;
   const status = percent >= 95 ? 'exception' : 'normal';
 
   const maxSpace = prettyBytes(maxBytes, { locale: uiLocale });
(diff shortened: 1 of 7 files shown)


src/components/used-storage.js
```diff
@@ -9,8 +9,8 @@ function UsedStorage({ usedBytes, maxBytes, showLabel }) {
   const { t } = useTranslation('usedStorage');
   const { uiLocale } = useLocale();
 
-  const percent = maxBytes ? usedBytes * 100 / maxBytes : 100;
-  const displayedPercent = `${Math.round(percent)} %`;
+  const percent = maxBytes ? usedBytes * 100 / maxBytes : 101;
+  const displayedPercent = percent > 100 ? '> 100 %' : `${Math.round(percent)} %`;
```
adrianaluchian, line 13: Would it be maybe more useful to display the real percentage when > 100% ? Feels a bit hacky this way.
Of course, the bar would just be filled 100%, just the displayed value maybe could be the real one?

adrianaluchian, line 13: Would it be maybe more useful to display the real percentage when > 100% ? Feels a bit hacky this way.
Of course, the bar would just be filled 100%, just the displayed value maybe could be the real one?

ahelmberger: Myeah, I don't hink there's an easy solution for that. First of all, if you're admin has unassigned your recent storage plan (so you have a `maxSize` of 0 now), but you still have a file with 1 Byte in your private storage, how many percent would that be? Also: it could lead to strange huge numbers that destroy the UI, like "1000000000000000000000000000000000000 %". I think there has to be a generic "overused" label, but I agree that "> 100 %" is hard to distinguish from "100 %". Maybe we should display a word there like "exceedance"/"Überschreitung". What do you think?

adrianaluchian: I see, we'd open the door to having to treat several edge cases. Ok then let's keep it simple, I like your idea of having a clear text there instead of `> 100%` 👍

ahelmberger: Done, replaced with "Exceedance" / "Überschreitung".

adrianaluchian (approved):

PR #300, opened 2022-03-08
EDU-423 Add favorite icon to rooms, lessons and documents

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-423


src/components/pages/doc.js
```diff
@@ -34,6 +34,7 @@ import {
   confirmSectionDelete,
   confirmSectionHardDelete
 } from '../confirmation-dialogs.js';
+import FavoriteStar from '../favorite-star.js';
```
adrianaluchian, line 37: 😱 Anarchy!

src/components/pages/doc.js
```diff
@@ -383,7 +384,10 @@ function Doc({ initialState, PageTemplate }) {
     <Fragment>
       <PageTemplate alerts={alerts}>
         <div className="DocPage">
-          <MetadataTitle text={selectedHistoryRevision ? selectedHistoryRevision.title : doc.title} />
+          <MetadataTitle
+            text={selectedHistoryRevision ? selectedHistoryRevision.title : doc.title}
```
adrianaluchian, line 388: μ-thing to shorten this condition: `selectedHistoryRevision?.title || doc.title` ?

src/components/pages/lesson.js
```diff
@@ -32,6 +32,7 @@ import {
   removeItemAt,
   replaceItemAt
 } from '../../utils/array-utils.js';
+import FavoriteStar from '../favorite-star.js';
```
adrianaluchian, line 35: 🙈 my eeeeyes! 😂

src/server/user-controller.js
```diff
@@ -411,6 +426,20 @@ class UserController {
       [needsPermission(permissions.EDIT_USERS), validateParams(userIdParamsSchema)],
       (req, res) => this.handleDeleteAllUserStorageReminders(req, res)
     );
+
+    router.post(
+      '/api/v1/users/favorites',
+      jsonParser,
+      validateBody(favoriteBodySchema),
+      (req, res) => this.handlePostFavorite(req, res)
```
adrianaluchian, line 434: I think we should have some permission restriction here? If these endpoints would be called without a user, there would be an exception thrown through the `userService`, but I don't think it should reach that point.

src/stores/user-store.js
```diff
@@ -55,6 +55,16 @@ class UserStore {
...
   }
+
+  pushToUserFavorites({ userId, favoriteType, favoriteId, favoriteSetOn }, { session } = {}) {
+    const favorite = { type: favoriteType, id: favoriteId, setOn: favoriteSetOn };
+    return this.collection.updateOne({ _id: userId }, { $push: { favorites: favorite } }, { session });
+  }
+
+  pullFromUserFavorites({ userId, favoriteType, favoriteId }, { session } = {}) {
+    const filter = { type: favoriteType, id: favoriteId };
+    return this.collection.updateOne({ _id: userId }, { $pull: { favorites: filter } }, { session });
```
adrianaluchian, line 66: Could I suggest `addToUserFavorites` and `removeFromUserFavorites`? `pull` made me expect a return of the favorites list initially, till I understood the intention 🤔

src/components/pages/doc.js
ahelmberger: I think we wrote this back when titles were allowed to be empty string. Because it's not a fallback display in this case: if `selectedHistoryRevision` is defined then its title has to be shown no matter what and never the `doc.title` (as we are in `history` mode). That's why I don't feel too well having the construct with the OR as it would be actually logically "wrong" and potentially misleading.

adrianaluchian:
Looks super nice 😎 
Was wondering if maybe using our secondary color on the stars would look good too. What'cha think?

Screen Shot 2022-03-09 at 09 39 15

src/server/user-controller.js
ahelmberger: You're absolutely right. I went for `VIEW_DOCS` as it is more or less the most basic thing.

src/components/pages/doc.js
adrianaluchian: Oh, that's true! Though in theory there should be no empty titles, this fallback could hide a bug. Got it.

src/stores/user-store.js
ahelmberger: Sure, sounds a little bit more "high-level", changed it.

src/components/pages/lesson.js
ahelmberger: I'm grief-stricken. How could this happen? Multiple times? Order is restored now.

src/components/pages/doc.js
ahelmberger: Oh no, we wrote this later of course, but the same reason applies why I'd like to leave it like it is.

src/components/pages/lesson.js
adrianaluchian: 😂 I felt the tectonic plates settling back into place

ahelmberger:
> Looks super nice sunglasses Was wondering if maybe using our secondary color on the stars would look good too. What'cha think?
> 
> Screen Shot 2022-03-09 at 09 39 15

Hmm, not too shabby (and more consistent). I'll switch to the secondary color.

ahelmberger:
> > Looks super nice sunglasses Was wondering if maybe using our secondary color on the stars would look good too. What'cha think?
> > Screen Shot 2022-03-09 at 09 39 15
> 
> Hmm, not too shabby (and more consistent). I'll switch to the secondary color.

Done!

adrianaluchian (approved):

PR #301, opened 2022-03-08
EDU-432 Create filter and sort components

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-432


ahelmberger:
It looks super-nice :heart_eyes: 

I just found a few UI glitches:

* When the UI language is DE and I see 'Nach "Update-Datum" sortieren' and I switch the UI language to EN, the label appears in mixed language 'Sort by "Update-Datum"'
* The terms 'Nach "Erstellt am" sortieren' and 'Sort by "Created on"' sound a little bit artificial. I would propose to not do string concatenation here and have extra translations for each sorting criteria instead to make full sentences with correct casing like 'Sort by creation date' / 'Nach Erstellungsdatum sortieren'. It would be beneficial to only use nouns in the dropdown, I guess.
* The arrow for sort direction should look like a full arrow (not just the head) otherwise it looks like an indicator that ther will be somthing dropping down when clicking (which is also the case but not the purpose of the button). Maybe a tooltip would help there, too

adrianaluchian:
> It looks super-nice 😍
> 
> I just found a few UI glitches:
>  ...

Thanks alot for the detailed feedback! I had looked at it for too long to be able to distinguish what could be better there.
Will go through the list and improve 👍

adrianaluchian:
> It looks super-nice 😍
> 
> I just found a few UI glitches:
> 

By the way, you once mentioned that the date translations don't sound good here. If you remember, could you please mention which alternatives you would suggest?

ahelmberger:
> > It looks super-nice heart_eyes
> > I just found a few UI glitches:
> 
> By the way, you once mentioned that the date translations don't sound good here. If you remember, could you please mention which alternatives you would suggest? For the dropdown values I mean.

I would suggest: "Erstellungsdatum" and "Änderungsdatum" in German.

ahelmberger:
> > It looks super-nice heart_eyes
> > I just found a few UI glitches:
> 
> Updated!

:heart: Looks really good and feels even nicer to use :+1:

ahelmberger (approved):

PR #303, opened 2022-03-10
EDU-427 Add latest user activities to dashboard

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-427

and looks as follows:

Screen Shot 2022-03-10 at 11 40 23

Screen Shot 2022-03-10 at 11 40 41

### DIFF
--- a/src/components/news-tab.yml
+++ b/src/components/news-tab.yml
@@ -1,0 +1,33 @@
+latestActivitiesHeader:
+  en: Latest activities
+  de: Letzte Aktivitäten
+documentCreatedActivity:
+  en: New document created
+  de: Neues Dokument erstellt
+documentUpdatedActivity:
+  en: Document updated
+  de: Dokument bearbeitet
+documentMarkedFavoriteActivity:
+  en: Document marked as favorite
+  de: Dokument als Favorit hinzugefügt
+roomCreatedActivity:
+  en: New room created
+  de: Neuen Raum ertellt
+roomUpdatedActivity:
+  en: Room updated
+  de: Raum bearbeitet
+roomMarkedFavoriteActivity:
+  en: Room marked as favorite
+  de: Raum als Favorit hinzugefügt
+roomJoinedActivity:
+  en: Private room joined
+  de: Privater Raum eingetreten
+lessonCreatedActivity:
+  en: New lesson created
+  de: Neue Unterrichtseinheit erstellt
+lessonUpdatedActivity:
+  en: Lesson updated
+  de: Unterrichtseinheit bearbeitet
+lessonMarkedFavoriteActivity:
+  en: Lesson marked as favorite
+  de: Unterrichtseinheit als Favorit hinzugefügt
--- a/src/components/news-tab.yml
+++ b/src/components/news-tab.yml
@@ -21,7 +21,7 @@ roomMarkedFavoriteActivity:
   de: Raum als Favorit hinzugefügt
 roomJoinedActivity:
   en: Private room joined
-  de: Privater Raum eingetreten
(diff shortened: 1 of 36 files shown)


src/services/dashboard-service.js
```diff
@@ -0,0 +1,140 @@
...
+    this.documentStore = documentStore;
+  }
+
+  async getUserActivities({ userId, limit = 0 }) {
+    const user = await this.userStore.getUserById(userId);
+
+    const createdRooms = await this.roomStore.getRoomsCreatedByUser(userId);
+    const joinedRooms = await this.roomStore.getRoomsJoinedByUser(userId);
+    const updatedRooms = await this.roomStore.getRoomsUpdatedByUser(userId);
+    const createdDocuments = await this.documentStore.getDocumentsMetadataByCreatedBy(userId);
```
ahelmberger, line 24: I was thinking a little bit about the naming pattern, and although `byCreatedBy` is the technically 100% correct name, it sounds unnatural. We could have a similar naming like in the room store with `getDocumentsMetadataCreatedByUser(userId)`, what do you think?

src/services/dashboard-service.js
```diff
@@ -0,0 +1,140 @@
...
+    const updatedDocuments = await this.documentStore.getDocumentsMetadataByUpdatedBy(userId);
+    const lessons = await this.lessonStore.getLessonsMetadataByCreatedBy(userId);
```
ahelmberger, line 26: A second consideration: We query the complete db without any limits here multiple times as part of the browser requesting the page. I think there should be a limit in the db queries - and - even more important - an index on the exact fields used for the query on each collection.

src/components/news-tab.yml
```diff
@@ -0,0 +1,33 @@
...
+  de: Neuen Raum erstellt
+roomUpdatedActivity:
+  en: Room updated
+  de: Raum bearbeitet
+roomMarkedFavoriteActivity:
+  en: Room marked as favorite
+  de: Raum als Favorit hinzugefügt
+roomJoinedActivity:
+  en: Private room joined
+  de: Privater Raum eingetreten
```
ahelmberger, line 24: Either "Privatem Raum beigetreten" or "Privaten Raum betreten" or "In privaten Raum eingetreten", there's a lot of choices :-) (The first one sounds fanciest)

src/services/dashboard-service.js
adrianaluchian: Updated all such phrased functions, thanks!

src/services/dashboard-service.js
adrianaluchian: We could/should indeed limit each collection query to the provided `limit` (or a default value), that's true. However we'd need to introduces sorting too, as for these case we specifically need certain date fields sorted "desc". I'll make changes in this direction and see what I encounter.

src/services/dashboard-service.js
ahelmberger: That's true, the `DocumentMetadata` is crooked anyway.

src/components/news-tab.yml
adrianaluchian: Thanks, fancy it is then! But mostly because it's the only one that starts with the entity type ("private room", like the others do) **and** is in Partizip 2 form.

ahelmberger: They all are Partizip 2 btw., but I liked the first one best too.

src/services/dashboard-service.js
adrianaluchian: Done, tested and added indices, thanks for the review and please check again.

PR #304, opened 2022-03-11
EDU-428 Add favorites tab to dashboard

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-428

and looks like

Screen Shot 2022-03-11 at 12 12 30

### DIFF
--- a/src/server/dashboard-controller.js
+++ b/src/server/dashboard-controller.js
@@ -19,13 +19,12 @@ class UserController {
 
   async handleGetDashboardPage(req, res) {
     const { user } = req;
-
     let rooms = [];
     if (this.serverConfig.areRoomsEnabled) {
       rooms = await this.roomService.getRoomsOwnedOrJoinedByUser(user._id);
     }
     const activities = await this.dashboardService.getUserActivities({ userId: user._id, limit: 10 });
-    const favorites = await this.dashboardService.getUserFavorites(user._id);
+    const favorites = await this.dashboardService.getUserFavorites(user);
 
     const mappedRooms = await Promise.all(rooms.map(room => this.clientDataMappingService.mapRoom(room, user)));
     const mappedActivities = this.clientDataMappingService.mapUserActivities(activities);
--- a/src/server/dashboard-controller.js
+++ b/src/server/dashboard-controller.js
@@ -24,11 +24,14 @@ class UserController {
     if (this.serverConfig.areRoomsEnabled) {
       rooms = await this.roomService.getRoomsOwnedOrJoinedByUser(user._id);
     }
-    const mappedRooms = await Promise.all(rooms.map(room => this.clientDataMappingService.mapRoom(room, user)));
     const activities = await this.dashboardService.getUserActivities({ userId: user._id, limit: 10 });
-    const mappedActivities = this.clientDataMappingService.mapUserActivities(activities);
+    const favorites = await this.dashboardService.getUserFavorites(user._id);
 
-    const initialState = { rooms: mappedRooms, activities: mappedActivities };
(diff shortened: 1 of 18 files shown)


src/server/dashboard-controller.js
```diff
@@ -24,11 +24,14 @@ class UserController {
     if (this.serverConfig.areRoomsEnabled) {
       rooms = await this.roomService.getRoomsOwnedOrJoinedByUser(user._id);
     }
-    const mappedRooms = await Promise.all(rooms.map(room => this.clientDataMappingService.mapRoom(room, user)));
     const activities = await this.dashboardService.getUserActivities({ userId: user._id, limit: 10 });
+    const favorites = await this.dashboardService.getUserFavorites(user._id);
```
ahelmberger, line 28: I don't think we have to fetch them from the DB, they should be there by `user.favorites` (as the `req.user` is the DB user object).

src/services/dashboard-service.js
```diff
@@ -136,6 +136,34 @@ class DashboardService {
 
     return activities;
   }
+
+  async getUserFavorites(userId) {
```
ahelmberger, line 140: We could make this the user object itself (which is there in `req.user` when calling this service).

src/services/dashboard-service.js
```diff
@@ -136,6 +136,34 @@ class DashboardService {
...
+    const user = await this.userStore.getUserById(userId);
+
+    const documentIds = user.favorites.filter(f => f.type === FAVORITE_TYPE.document).map(d => d.id);
+    const roomIds = user.favorites.filter(f => f.type === FAVORITE_TYPE.room).map(r => r.id);
+    const lessonIds = user.favorites.filter(f => f.type === FAVORITE_TYPE.lesson).map(l => l.id);
+
+    const documents = documentIds.length ? await this.documentStore.getDocumentsMetadataByKeys(documentIds) : [];
+    const rooms = roomIds.length ? await this.roomStore.getRoomsByIds(roomIds) : [];
+    const lessons = lessonIds.length ? await this.lessonStore.getLessonsMetadataByIds(lessonIds) : [];
```
ahelmberger, line 149: We could parallelize this by `Promise.all`

src/stores/document-store.js
```diff
@@ -54,6 +54,10 @@ class DocumentStore {
     return this.collection.distinct('key');
   }
 
+  getDocumentsMetadataByKeys(keys, { session } = {}) {
+    return this.collection.find({ key: { $in: keys } }, { session }).toArray();
```
ahelmberger, line 58: There is no index on the `key` in documents, only on `_id` (which has the same value)

src/services/dashboard-service.js
adrianaluchian: Indeed, done, thanks!

src/services/dashboard-service.js
adrianaluchian: True, cause the `user` in the controller has the whole data 🤔 
Changed, thanks.

src/stores/document-store.js
adrianaluchian: Funny, did not realize that documents don't have a `key` index. I agree, maybe we can get around to removing this field eventually. Till then I decided to add the index, to make sure we don't suffer from this.

src/services/dashboard-service.js
ahelmberger: :+1:

ahelmberger (approved):

src/stores/document-store.js
ahelmberger: :+1:

PR #313, opened 2022-03-24
EDU-453 Distinguish room invitation invalidation reasons

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-453

### DIFF
--- a/src/services/room-service.js
+++ b/src/services/room-service.js
@@ -6,9 +6,9 @@ import RoomStore from '../stores/room-store.js';
 import LockStore from '../stores/lock-store.js';
 import UserStore from '../stores/user-store.js';
 import LessonStore from '../stores/lesson-store.js';
-import { ROOM_ACCESS_LEVEL } from '../domain/constants.js';
 import TransactionRunner from '../stores/transaction-runner.js';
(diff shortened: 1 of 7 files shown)


src/services/room-service.js
```diff
@@ -119,23 +119,35 @@ export default class RoomService {
...
+    }
+
+    if (invitation.email !== user.email) {
+      response.invalidInvitationReason = INVALID_ROOM_INVITATION_REASON.user;
+      logger.debug(`Invitation ${invitation._id} was sent to email '${invitation?.email}' but accessed by user with email '${user.email}'`);
+      return response;
+    }
+
+    const room = await this.roomStore.getRoomById(invitation.roomId);
+    if (!room) {
```
adrianaluchian, line 142: Technically this scenario can not occur, as deleting a room also deletes the invitations.
However I will leave the check here so that we have a clear message to show the user, in case we introduce a bug in the future.

ahelmberger (approved):

src/services/room-service.js
ahelmberger: :+1:

PR #314, opened 2022-03-28
EDU-434 Update documents page table

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-434

and looks like:

Screen Shot 2022-03-28 at 10 37 10

Screen Shot 2022-03-28 at 10 37 29

### DIFF
--- a/src/components/document-info-cell.yml
+++ b/src/components/document-info-cell.yml
@@ -1,0 +1,6 @@
+createdOn:
+  en: Created
+  de: Erstellt
+updatedOn:
+  en: Last update
+  de: Letzte Änderung
--- a/src/resources/common.yml
+++ b/src/resources/common.yml
@@ -226,3 +226,6 @@ sortedBySetOn:
 sortedByType:
   en: Sorted by type
   de: Nach Typ sortiert
+searchPlaceholder:
+  en: Search terms
+  de: Suchbegriffe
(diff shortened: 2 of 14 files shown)


src/resources/common.yml
```diff
@@ -226,3 +232,15 @@ sortedBySetOn:
...
   de: Nach Typ sortiert
+sortedByUser:
+  en: Sorted by user
+  de: Nach Benutzer sortiert
+sortedByOrigin:
+  en: Sorted by origin
+  de: Nach Quelle sortiert
+sortedByArchived:
+  en: Sorted by archived status
+  de: Nach Archiviert Status sortiert
```
ahelmberger, line 243: better: "Nach Archivierungsstatus sortiert"

adrianaluchian: Updated, thanks!

src/components/document-info-cell.yml
```diff
@@ -0,0 +1,6 @@
+createdOn:
+  en: Created
+  de: Erstellt
+updatedOn:
+  en: Last update
+  de: Letzte Änderung
```
ahelmberger, line 6: All their brothers and sisters mostly live in `common.yml` (but maybe those are used more often than once), I wouldn't mind having those two there, too, but just a "gut feeling", no urge.

ahelmberger (approved):

src/components/document-info-cell.yml
adrianaluchian: These two are more custom translations, they diverge from the common (they are shorter versions basically).

PR #315, opened 2022-03-28
EDU-447 Auto image resize

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-447

and looks like:

Screen Shot 2022-03-28 at 15 11 59
Screen Shot 2022-03-28 at 15 12 34


src/components/storage-upload-helper.js
```diff
@@ -0,0 +1,49 @@
+import { IMAGE_DOWN_SCALING_WIDTH } from '../domain/constants.js';
+
+const getScaledDimensions = (img, width) => {
+  const ratio = img.naturalWidth / width;
+  return { width, height: Math.round(((img.naturalHeight / ratio) + Number.EPSILON) * 100) / 100 };
+};
+
+const isImage = file => file && file.type.startsWith('image');
```
ahelmberger, line 8: We can only process raster images with the canvas, so files with type `image/svg` will result in an invalid file (if its size is > `IMAGE_DOWN_SCALING_WIDTH`).

src/components/storage-upload-helper.js
```diff
@@ -0,0 +1,49 @@
+import { IMAGE_DOWN_SCALING_WIDTH } from '../domain/constants.js';
```
ahelmberger, line 1: I wonder whether we should put this file outside of the `components` folder and into some of the `utils` (as it is not a react component?

ahelmberger:
Wow, what an elegant and simple solution without libraries etc. :cupid: :+1:

src/components/storage-upload-helper.js
adrianaluchian: Interesting, I hadn't considered that 🤔 Should we exclude SVGs from processing then?
I'm not sure if there's another way to resize images on the client side.

src/components/storage-upload-helper.js
adrianaluchian: I considered putting it in `src/utils`, but I had second thoughts because it's only for the client side. However I see we have `file-helper` already, only for the server side-so that's fine then. I'll move it 👍

src/components/storage-upload-helper.js
ahelmberger: SVGs do not grow in byte size when they grow in pixel size, so they can be left untouched (it will still show the warning if they're smaller than 1200). So, in my opinion we should do a positive test on well-known image types (`image/png`, `image/jpg`, `image/jpeg`, `image/webp`, so basically those that are supported by `canvas.toBlob()`, others will probably not work anyways).

src/components/storage-upload-helper.js
ahelmberger: :+1:

src/components/storage-upload-helper.js
adrianaluchian: 🤦‍♀️ of course, you are right, they are vector images. Will do a white-list check!

adrianaluchian: Done, thank you!

ahelmberger (approved):

PR #316, opened 2022-03-28
EDU-450 Support uploading video image poster

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-450

On a Prod dump, the report was:

```
Updated 17 lessons, 285 documents, 5052 documentRevisions
info [2022-03-28T18:03:28.041Z] [/src/stores/database.js] {
  event: 'migrated',
  name: 'educandu-2022-03-28-01-update-video-sections-content',
  durationSeconds: 18.555
}
```

### DIFF
--- a/src/plugins/video/info.js
+++ b/src/plugins/video/info.js
@@ -20,8 +20,8 @@ export default class Video {
 
   getDefaultContent() {
     return {
-      type: SOURCE_TYPE.internal,
-      url: '',
+      sourceType: SOURCE_TYPE.internal,
+      sourceUrl: '',
       text: '',
       width: 100,
       aspectRatio: {
@@ -41,6 +41,6 @@ export default class Video {
   }
 
   getCdnResources(content) {
-    return content.type === SOURCE_TYPE.internal && content.url ? [content.url] : [];
+    return content.sourceType === SOURCE_TYPE.internal && content.sourceUrl ? [content.sourceUrl] : [];
   }
 }
--- a/src/plugins/video/info.js
+++ b/src/plugins/video/info.js
@@ -41,6 +41,8 @@ export default class Video {
   }
 
   getCdnResources(content) {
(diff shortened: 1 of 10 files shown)


src/plugins/video/info.js
```diff
@@ -37,6 +41,6 @@ export default class Video {
   }
 
   getCdnResources(content) {
-    return content.type === SOURCE_TYPE.internal && content.url ? [content.url] : [];
+    return content.sourceType === SOURCE_TYPE.internal && content.sourceUrl ? [content.sourceUrl] : [];
```
ahelmberger, line 44: The poster image - if set - is also a CDN resource but is missing here.

src/plugins/video/video.yml
```diff
@@ -37,3 +37,6 @@ width:
 copyrightInfos:
   en: Copyright infos
   de: Copyright Infos
+posterImageUrl:
+  en: Preview image URL
+  de: Vorschau Bild URL
```
ahelmberger, line 42: We need hyphens between the words: "Vorschau-Bild-URL"

src/ui/mime-type-helper.js
```diff
@@ -222,7 +222,7 @@ function getCategory(pathOrExtension) {
       return CATEGORY_AUDIO;
     case 'image':
       return CATEGORY_IMAGE;
-    case 'vido':
+    case 'video':
```
ahelmberger, line 225: :scream:

src/plugins/video/video.yml
adrianaluchian: Ge-hyphenated, thank you

src/plugins/video/info.js
adrianaluchian: Good catch, thanks! Updated

src/ui/mime-type-helper.js
adrianaluchian: Seems we always showed "Unbekannt" for videos 😅 Well, now we **know**

ahelmberger (approved):

PR #318, opened 2022-03-29
EDU-442 Limit storage upload

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-442


src/components/storage-browser.js
```diff
@@ -323,6 +324,14 @@ class StorageBrowser extends React.Component {
       }
     }
 
+    if (requiredBytes > LIMIT_PER_STORAGE_UPLOAD_IN_BYTES) {
+      message.error(t('uploadLimitExceeded', {
+        uploadSize: prettyBytes(requiredBytes, { locale: this.props.uiLocale }),
+        uploadLimit: prettyBytes(LIMIT_PER_STORAGE_UPLOAD_IN_BYTES, { locale: this.props.uiLocale })
+      }));
+      return;
```
ahelmberger, line 332: Maybe let's also set the same limit in the middleware of the controller so the server doesn't accept more than `LIMIT_PER_STORAGE_UPLOAD_IN_BYTES`, what do you think?

adrianaluchian: Of course, added it on the server side too, thanks!

ahelmberger: Cool thanks :+1:

src/server/storage-controller.js
```diff
@@ -55,6 +57,12 @@ class StorageController {
       throw new BadRequest(`Invalid storage path '${prefix}'`);
     }
 
+    const requiredBytes = files.reduce((totalSize, file) => totalSize + file.size, 0);
```
ahelmberger, line 60: I had actually something different in mind: here the bytes are already saved on the server (and could potentially put it under pressure. We should set a limit in the `multipartParser` middleware to not even accept more than the number of bytes. Sorry for having been too vague about it in my last comment.

ahelmberger (approved):

PR #320, opened 2022-03-30
EDU-422 Add storage plan management

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-422


src/components/storage-plan-modal.js
```diff
@@ -0,0 +1,101 @@
...
+
+function StoragePlanModal({ isVisible, storagePlan, storagePlanNamesInUse, onOk, onCancel }) {
+  const [form] = Form.useForm();
+  const { t } = useTranslation('storagePlanModal');
+  const [loading, setLoading] = useState(false);
+
+  const nameValidationRules = [
+    {
+      required: true,
+      message: t('nameRequired'),
```
adrianaluchian, line 18: `nameRequired`, also known as "translation required" 😄

src/components/storage-plan-modal.js
```diff
@@ -0,0 +1,101 @@
...
+      setLoading(false);
+    } catch (error) {
+      setLoading(false);
+    }
+  };
+
+  const initialValues = {
+    _id: storagePlan?._id || '',
+    name: storagePlan?.name || t('newStoragePlan'),
+    maxBytes: storagePlan?.maxBytes || 100 * 1000 * 1000
```
adrianaluchian, line 61: Here we'd need to check `storagePlan ? storagePlan.maxBytes : 100 * 1000 * 1000` because it seems that when you don't set the bytes (or set them to zero) and then edit the plan, you see "100 MB" and not your given value.

src/domain/permissions.js
```diff
@@ -37,6 +39,9 @@ const rolesForPermission = {
   [RESTORE_DOC_REVISIONS]: [ROLE.admin],
   [MANAGE_ARCHIVED_DOCS]: [ROLE.admin],
   [MANAGE_IMPORT]: [ROLE.admin],
+  [MANAGE_EXPORT]: [], // This permission is restricted to a built-in user exclusively
```
adrianaluchian, line 42: Should we rather reflect the comment in the name of the permission? I for one prefer that over comments, when possible.
Maybe `MANAGE_EXPORT_WITH_BUILT_IN_USER` ?

src/components/admin/settings-tab.js
```diff
@@ -0,0 +1,170 @@
...
+}
+
+SettingsTab.propTypes = {
+  documents: PropTypes.arrayOf(documentMetadataShape).isRequired,
+  initialSettings: settingsShape.isRequired,
+  onDirtyStateChange: PropTypes.func.isRequired,
+  onSettingsSaved: PropTypes.func.isRequired
+};
+
+export default SettingsTab;
```
adrianaluchian, line 170: This tab looks so much more readable now with cards! 👍

src/components/storage-plan-modal.js
ahelmberger: That's so true :laughing: Added!

src/components/admin/storage-plans-tab.js
```diff
@@ -0,0 +1,162 @@
...
+      </a>
+    ];
+
+    if (!storagePlan.assignedUserCount) {
+      actions.push((
+        <a
+          key="delete"
+          onClick={() => handleDeleteClick(storagePlan)}
+          disabled={!!storagePlan.assignedUserCount}
+          >
```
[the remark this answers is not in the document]
ahelmberger: Unfortunately Antd hard-codes the color for these actions (including a blue-ish hover effect). Overriding it would require more overrides of ant styles. I initially tried (hence the superfluous `disabled` prop above, which I removed now), but then decided to leave it. Maybe I should have simply used a table here like we do on other pages. Should we change it (to table) maybe?

src/components/storage-plan-modal.js
ahelmberger: Good catch, thanks. I changed it to `??` instead of `||` which should do the trick.

src/domain/permissions.js
ahelmberger: Good idea, changed to `MANAGE_EXPORT_WITH_BUILT_IN_USER`.

src/components/admin/settings-tab.js
ahelmberger: Completely agree :grin:

src/components/admin/storage-plans-tab.js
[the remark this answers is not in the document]
adrianaluchian: Ok, I see. If you feel it's not a big effort and not overcomplicating the UI, then I'd vote for the table change. Up to you!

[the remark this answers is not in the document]
ahelmberger: Then I'll go for the table. The cards look nice but they'd introduce another type of UI we have nowhere else.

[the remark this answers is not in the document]
adrianaluchian: Poifect!

PR #323, opened 2022-03-31
EDU-458 Allow login with email address or username

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-458

### DIFF
--- a/src/components/login-form.js
+++ b/src/components/login-form.js
@@ -1,8 +1,7 @@
 import PropTypes from 'prop-types';
-import urls from '../utils/urls.js';
+import { Form, Input } from 'antd';
 import React, { useState } from 'react';
 import Logger from '../common/logger.js';
-import { Form, Input, Button } from 'antd';
 import { useSetUser } from './user-context.js';
 import { useTranslation } from 'react-i18next';
 import errorHelper from '../ui/error-helper.js';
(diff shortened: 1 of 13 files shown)


src/stores/user-store.js
```diff
@@ -11,8 +11,25 @@ class UserStore {
...
+    }).toArray();
+
+    switch (matches.length) {
+      case 0:
+        return null;
+      case 1:
+        return matches[0];
+      default:
+        // Always prefer matching emails over usernames:
+        return matches.find(match => match.email === emailOrUsername);
```
adrianaluchian, line 31: Micro thing, but 
```
const preferredMatch = matches.find(match => match.email === emailOrUsername);
return preferredMatch;
```
would speak for itself without a comment that risks getting out of sync with code :bowtie:

src/stores/user-store.spec.js
```diff
@@ -25,28 +25,37 @@ describe('user-store', () => {
...
+        result = await sut.findUserByLogin({ provider: user.provider, emailOrUsername: user.email });
+      });
+      it('should return the user', () => {
+        expect(result).toEqual(user);
+      });
+    });
+
+    describe('when provider matches and emailOrUsername matches the username', () => {
+      beforeEach(async () => {
+        result = await sut.findUserByLogin({ provider: user.provider, emailOrUsername: user.username });
```
adrianaluchian, line 58: Would be good to also add a test for both username and email matching.

src/components/login-form.js
```diff
@@ -1,8 +1,7 @@
 import PropTypes from 'prop-types';
```
adrianaluchian, line 1: Both the code here and the UI look alot better now 👍 
Maybe we could make a ticket to update registration in the same direction.

src/stores/user-store.js
ahelmberger: Changed it, my commentophobic friend :laughing:

adrianaluchian: 😄 thanks! I would say rather a deep dislike than a phobia, I rather have phobia of code that needs comments 😨

src/stores/user-store.spec.js
ahelmberger: Yep, good idea, added a test for this case.

src/stores/user-store.js
ahelmberger: :grin:

src/components/login-form.js
ahelmberger: Yes, I created a ticket: https://educandu.atlassian.net/browse/EDU-462

adrianaluchian (approved):

PR #326, opened 2022-04-01
EDU-455 Create timeline component

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-455
 and can be tested locally on the admin page -> tests tab.

### DIFF
--- a/src/components/pages/admin.yml
+++ b/src/components/pages/admin.yml
@@ -4,3 +4,6 @@ settingsTabTitle:
 storagePlansTabTitle:
   en: Storage plans
   de: Speicherpläne
+testsTabTitle:
+  en: Tests
+  de: Teste
--- a/src/components/pages/admin.yml
+++ b/src/components/pages/admin.yml
@@ -6,4 +6,4 @@ storagePlansTabTitle:
   de: Speicherpläne
 testsTabTitle:
   en: Tests
-  de: Teste
+  de: Tests
--- a/src/components/timeline.js
+++ b/src/components/timeline.js
@@ -139,12 +139,6 @@ function Timeline({ length, parts, onPartAdd, onPartDelete, onStartTimecodeChang
       return { leftMin, leftMax };
     }).filter(bound => bound);
 
-    console.log(`markers ${JSON.stringify(markers)}`);
-    console.log(`segments ${JSON.stringify(segments)}`);
-    console.log(`msToPxRatio ${JSON.stringify(msToPxRatio)}`);
-    console.log(`bounds ${JSON.stringify(bounds)}`);
-    console.log(`markerBounds ${JSON.stringify(markerBounds)}`);
-
     setNewMarkerBounds(markerBounds);
     setTimelineState({ markers, segments, msToPxRatio, bounds, minSegmentLength });
   }, [parts, length]);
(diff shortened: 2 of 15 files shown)


src/components/timeline.js
```diff
@@ -0,0 +1,253 @@
...
+      updateStates();
+    }
+  }, [timelineRef, updateStates]);
+
+  useEffect(() => {
+    window.addEventListener('resize', updateStates);
+    return () => {
+      window.removeEventListener('resize', updateStates);
+    };
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
ahelmberger, line 157: Wouldn't it work anymore when we put `updateStates` into the dependency array?

src/components/timeline.js
```diff
@@ -0,0 +1,253 @@
...
+        <FlagOutlined onMouseDown={handleMarkerMouseDown(marker, index)} />
+      </div>
+    );
+  };
+
+  const renderSegment = segment => (
+    <div key={segment.key} className="Timeline-segment" style={{ width: `${segment.width}px` }}>{segment.title}</div>
+  );
+
+  const renderButton = segment => {
```
ahelmberger, line 184: Maybe better: `renderDeleteButton`

src/components/timeline.js
```diff
@@ -0,0 +1,253 @@
...
+    };
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useEffect(() => {
+    if (!dragState || isTouchDevice()) {
+      return () => { };
+    }
+    window.addEventListener('mousemove', handleWindowMouseMove);
+    window.addEventListener('mouseup', handleWindowMouseUp);
```
ahelmberger, line 165: We should also remove this handler when the dragging is over, shouldn't we?

ahelmberger:
Nice and clean, and very compact given the complexity :heart_eyes:

src/components/pages/admin.yml
```diff
@@ -4,3 +4,6 @@ settingsTabTitle:
 storagePlansTabTitle:
   en: Storage plans
   de: Speicherpläne
+testsTabTitle:
+  en: Tests
+  de: Teste
```
ahelmberger, line 9: "Teste" is only used for medical tests, otherwise the plural is also "Tests" in German.

src/components/timeline.js
adrianaluchian: I actually moved this inside the one usage, was a temporary change that I hadn't reverted.

src/components/timeline.js
adrianaluchian: Instead of removing the listener I had removed the whole line at some point by mistake 😂 Added back, thanks.

src/components/timeline.js
adrianaluchian: Thanks for catching the bug behind this 🙏 Corrected

src/components/pages/admin.yml
adrianaluchian: I see 🤔 Corrected!

adrianaluchian:
> Nice and clean, and very compact given the complexity 😍

Well thank you! ☺️

ahelmberger (approved):

PR #328, opened 2022-04-01
EDU-459 Enable lesson cloning

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-459

### DIFF
--- a/src/server/lesson-controller.js
+++ b/src/server/lesson-controller.js
@@ -48,7 +48,7 @@ class LessonController {
       throw new NotFound();
     }
 
-    if (templateLessonId && lesson.roomId !== templateLesson.roomId) {
+    if (templateLesson && lesson.roomId !== templateLesson.roomId) {
       throw new BadRequest();
     }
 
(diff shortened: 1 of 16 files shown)


src/server/lesson-controller.js
```diff
@@ -37,16 +38,22 @@ class LessonController {
...
 
     const lesson = await this.lessonService.getLessonById(lessonId);
+    const templateLesson = templateLessonId ? await this.lessonService.getLessonById(templateLessonId) : null;
 
-    if (!lesson) {
+    if (!lesson || (templateLessonId && !templateLesson)) {
       throw new NotFound();
     }
 
+    if (templateLessonId && lesson.roomId !== templateLesson.roomId) {
```
adrianaluchian, line 51: Tiny thing: at this point would probably make more sense to use `templateLesson` instead of `templateLessonId` in this condition 🤔

adrianaluchian (approved):

src/server/lesson-controller.js
ahelmberger: True, thanks, changed it to `templateLesson`

PR #332, opened 2022-04-05
EDU-463 Adjust image optimization

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-463

Some non-optimized vs optimized results on different file types (file name reflects case):

Screen Shot 2022-04-05 at 10 26 29


src/utils/storage-helper.js
```diff
@@ -1,13 +1,21 @@
...
 
 const scalableFileTypes = ['image/png', 'image/jpg', 'image/jpeg', 'image/webp'];
 
-const scaleDownImage = ({ file, width }) => {
+const shouldOptimizeImage = ({ naturalSize, naturalWidth, width }) => {
+  const optimize = (naturalWidth > width && (naturalSize > 500 * 1000)) || (naturalSize > 1.5 * 1000 * 1000);
+  return optimize;
+};
+
+const optimizeImage = ({ file, width }) => {
```
ahelmberger, line 18: Just a small thing: I think it might better easier to understand immediately, if this param would be called something like e.g. `widthThreshold` or `optimalWidth`, what do you think?

ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-463
> 
> Some non-optimized vs optimized results on different file types (file name reflects case):
> 
> Screen Shot 2022-04-05 at 10 26 29

The result looks really good!

src/utils/storage-helper.js
```diff
@@ -1,13 +1,21 @@
...
+  }
   const ratio = img.naturalWidth / width;
   return { width, height: Math.round(((img.naturalHeight / ratio) + Number.EPSILON) * 100) / 100 };
 };
 
 const scalableFileTypes = ['image/png', 'image/jpg', 'image/jpeg', 'image/webp'];
 
-const scaleDownImage = ({ file, width }) => {
+const shouldOptimizeImage = ({ naturalSize, naturalWidth, width }) => {
+  const optimize = (naturalWidth > width && (naturalSize > 500 * 1000)) || (naturalSize > 1.5 * 1000 * 1000);
```
ahelmberger, line 14: It's maybe a little bit inconsistent that the `width` threshold is parameterized, but the `size` (in bytes) thresholds are hard-coded. We could put them also into `constants.js`.

adrianaluchian: That's true, I extracted all 3 limits and named them clearly, then used them directly in the `storage-helper`. Thanks for pointing this out.

src/utils/storage-helper.js
adrianaluchian: Ended up using the constant directly and no longer passing it as an argument.

src/utils/storage-helper.js
ahelmberger: Cool, thanks :+1:

ahelmberger (approved):

PR #334, opened 2022-04-06
EDU-468 Add clip effect to image plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-468


src/plugins/image/image.yml
```diff
@@ -16,6 +16,9 @@ hoverOption:
 revealOption:
   en: Reveal
   de: Auflösung
+clipOption:
+  en: Clip
+  de: Abschneiden
```
ahelmberger, line 21: "Zuschneiden" would be better

src/plugins/image/image.yml
```diff
@@ -31,6 +34,12 @@ copyrightInfos:
 maximumWidth:
   en: Maximum width
   de: Maximale Breite
+clippedWidth:
+  en: Clipping actual width
+  de: Echte abgeschnittene Breite
```
ahelmberger, line 39: I think it would be more understandable to the user if we'd say: "Resulting (width|height)" / "Resultierende (Breite|Höhe)"

src/plugins/image/image.yml
```diff
@@ -49,3 +58,6 @@ orientationOptionHorizontal:
 smallImageSizeWarning:
   en: The selected image has a smaller width than the page content and might therefore be displayed smaller than expected.
   de: Das von Ihnen gewählte Bild hat eine geringere Breite als der Seiteninhalt und wird deshalb unter Umständen kleiner als erwartet angezeigt.
+smallClippingSizeWarning:
+  en: The selected clipping has a smaller width than the page content and might therefore be displayed smaller than expected.
+  de: Das von Ihnen gewählte Abschneidung hat eine geringere Breite als der Seiteninhalt und wird deshalb unter Umständen kleiner als erwartet angezeigt.
```
ahelmberger, line 63: Either "Der von Ihnen gewählte Zuschnitt hat ..." or "Der ausgewählte Bereich hat ..." (I'd like the second one better, also in English I'd say "The selected area has ...")

src/plugins/image/display/image-display.js
```diff
@@ -9,18 +9,37 @@ import GithubFlavoredMarkdown from '../../../common/github-flavored-markdown.js'
...
+    const canvas = clipEffectCanvasRef.current;
+    const context = canvas.getContext('2d');
+    const width = img.naturalWidth * (effect.region.width / 100);
+    const height = img.naturalHeight * (effect.region.height / 100);
+    const x = img.naturalWidth * (effect.region.x / 100);
+    const y = img.naturalHeight * (effect.region.y / 100);
+    canvas.width = width;
+    canvas.height = height;
+    context.drawImage(img, x, y, width, height, 0, 0, width, height);
+  }, [clipEffectCanvasRef, effect]);
```
ahelmberger, line 35: I think we have to specify `clipEffectImageRef.current` and `clipEffectCanvasRef.current`?

src/plugins/image/editing/image-editor.js
```diff
@@ -36,79 +41,76 @@ function ImageEditor({ content, onContentChanged, publicStorage, privateStorage
...
+      return;
+    }
+
+    const img = clipEffectImageRef.current;
+    const clipWidth = Math.round(img.naturalWidth * (effect.region.width / 100));
+    const clipHeight = Math.round(img.naturalHeight * (effect.region.height / 100));
+
+    setClippingSizeInPx({ width: clipWidth, height: clipHeight });
+    setSmallImageSizeWarning(clipWidth < SMALL_IMAGE_WIDTH_THRESHOLD ? t('smallClippingSizeWarning') : null);
+  }, [currentImageSource, effectType, effect?.region, t]);
```
ahelmberger, line 75: Could we also put `clipEffectImageRef.current` here?

src/plugins/image/editing/image-editor.js
```diff
@@ -15,11 +16,15 @@ const RadioGroup = Radio.Group;
 const FormItem = Form.Item;
 const { TextArea } = Input;
 
+const defaultClipRegion = { x: 0, y: 0, width: 0, height: 0, data: {} };
```
ahelmberger, line 19: One suggestion and on question:

* Wouldn't it be nicer from a UX perspective to set the default to `{ x: 0, y: 0, width: 100, height: 100 }`? It's hard to spot the little square in the left top corner after updating the image source.
* `data` seems to be always an empty object, or did I miss something?

src/plugins/image/editing/image-editor.js
```diff
@@ -154,109 +197,149 @@ function ImageEditor({ content, onContentChanged, publicStorage, privateStorage
...
-                  <FormItem label={t('orientationLabel')} {...formItemLayout}>
-                    <RadioGroup value={effect.orientation} onChange={handleOrientationValueChanged}>
-                      <RadioButton value={ORIENTATION.horizontal}>{t('orientationOptionHorizontal')}</RadioButton>
-                      <RadioButton value={ORIENTATION.vertical}>{t('orientationOptionVertical')}</RadioButton>
-                    </RadioGroup>
-                  </FormItem>
+
+              {effect.type === EFFECT_TYPE.clip && (
+                <div className="Image-clipEffect">
+                  {!!currentImageSource && (
```
ahelmberger, line 316: Just a small idea: It would be nice to have a line of text above the clip editor saying something like "Please use the selection rectangle to specify the displayed area" or similar.

src/plugins/image/image.yml
adrianaluchian: Should we also change the above text in the same direction?
"Das von Ihnen gewählte Bild" -> "Das ausgewählte Bild"

ahelmberger: I like! (it sounds a little bit less accusing when we don't say 'your' or 'by you')

src/plugins/image/editing/image-editor.js
adrianaluchian: Good idea, I shall add that.
"Bitte der rechteckige Zuschnitt benutzen, um eine Bereich auszuwählen"?

src/plugins/image/image.yml
adrianaluchian: I struggled for a good wording here, this is better, thanks!
Updated

src/plugins/image/image.yml
adrianaluchian: Updated

src/plugins/image/editing/image-editor.js
ahelmberger: "Bitte benutzen Sie das Markierungsrechteck, um einen Bildbereich auszuwählen" comes to my mind

src/plugins/image/image.yml
adrianaluchian: Updated both cases!

src/plugins/image/editing/image-editor.js
adrianaluchian: - It's a good idea, we can set it to a default rectangle that is actually visible. 
Since the values are percentages, shall we go for something like `{ x: 0, y: 0, width: 15, height: 15 }` ?
- Yes, `data` is for us an empty object always, the plugin does need it to be present though

ahelmberger: As I consider "clipping" to be a process started from the original size, I'd go for 100, but on the other hand the rectangle would be more visible if it were inset by 10% of each side. Your call :-)

ahelmberger: Can we just feed the `data` property to the plugin, but not persist it as part of the data model?

src/plugins/image/display/image-display.js
adrianaluchian: Thanks for noticing this, added `clipEffectImageRef`.

src/plugins/image/editing/image-editor.js
adrianaluchian: Actually while doing this I tested a bit more and discovered some strange inconsistencies when switching images and reworked the logic a bit to fix the issues.

src/plugins/image/editing/image-editor.js
adrianaluchian: - Went for the "10% on each size" suggestion, I think that's a nice one. "100%" would have been just as invisible as "0%" I think, hard for the user to notice the rectangle.
- Indeed, I thought of `data` as the model of the plugin, what it needs to render correctly, but we don't need it in out own persisted model. Removed, thanks!

src/plugins/image/editing/image-editor.js
adrianaluchian: Would have not come to my mind 😅 Thanks, done

src/plugins/image/editing/image-editor.js
ahelmberger: Cool, thanks!

src/plugins/image/editing/image-editor.js
ahelmberger: OK, was on the wrong track anyways. Refs cannot be dependencies.

ahelmberger (approved):

PR #337, opened 2022-04-08
EDU-474 Make initial PDF page number configurable

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-474

### DIFF
--- a/migrations/educandu-2022-04-08-01-add-initial-page-number-to-pdf-viewer-plugin.js
+++ b/migrations/educandu-2022-04-08-01-add-initial-page-number-to-pdf-viewer-plugin.js
@@ -1,0 +1,62 @@
+/* eslint-disable camelcase, no-console */
+export default class Educandu_2022_04_08_01_add_initial_page_number_to_pdf_viewer_plugin {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async collectionUp(collectionName) {
+    const result = await this.db.collection(collectionName).updateMany(
+      {},
+      {
+        $set: {
+          'sections.$[sectionElement].content.initialPageNumber': 1
+        }
+      },
+      {
+        arrayFilters: [
+          {
+            'sectionElement.type': 'pdf-viewer',
+            'sectionElement.content': { $ne: null }
+          }
+        ],
+        multi: true
+      }
+    );
+
+    console.log(`Updated ${collectionName}: ${JSON.stringify(result)}`);
+  }
+
+  async collectionDown(collectionName) {
+    const result = await this.db.collection(collectionName).updateMany(
+      {},
+      {
+        $unset: {
+          'sections.$[sectionElement].content.initialPageNumber': null
+        }
+      },
+      {
+        arrayFilters: [
+          {
+            'sectionElement.type': 'pdf-viewer',
+            'sectionElement.content': { $ne: null }
+          }
+        ],
+        multi: true
+      }
+    );
+
+    console.log(`Updated ${collectionName}: ${JSON.stringify(result)}`);
+  }
+
+  async up() {
+    await this.collectionUp('lessons');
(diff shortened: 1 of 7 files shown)


migrations/educandu-2022-04-08-01-add-initial-page-number-to-pdf-viewer-plugin.js
```diff
@@ -0,0 +1,62 @@
...
+  async collectionUp(collectionName) {
+    const result = await this.db.collection(collectionName).updateMany(
+      {},
+      {
+        $set: {
+          'sections.$[sectionElement].content.initialPageNumber': 1
+        }
+      },
+      {
+        arrayFilters: [
```
adrianaluchian, line 16: 😱 great!!!

src/plugins/pdf-viewer/editing/pdf-viewer-editor.js
```diff
@@ -63,6 +68,9 @@ function PdfViewerEditor({ content, onContentChanged, publicStorage, privateStor
               />
           </div>
         </FormItem>
+        <Form.Item label={t('initialPageNumber')} {...formItemLayout}>
+          <InputNumber min={1} step={1} value={initialPageNumber} onChange={handleInitialPageNumberChange} />
```
adrianaluchian, line 72: Is there any way to identify the max too? If you set a number greater than the page count, an exception is thrown.

ahelmberger: It would complicate things too much because we would have to load the PDF first in order to find out. That's why I introduced the new page loading error in the display. I think we can leave the responsibility to set a correct value to the user (which would be consistent, I think, as we are always very forgiving in the plugins and allowing dead image links etc.).

migrations/educandu-2022-04-08-01-add-initial-page-number-to-pdf-viewer-plugin.js
ahelmberger: That's what I thought, too :laughing:

src/plugins/pdf-viewer/display/pdf-viewer-display.js
```diff
@@ -115,11 +124,11 @@ function PdfViewerDisplay({ content }) {
...
-          <a className="PdfViewer-pagerItem" disabled={pageNumber <= 1} onClick={handlePreviousPageButtonClick}><CaretLeftOutlined /></a>
+          <Button type="link" className="PdfViewer-pagerItem" disabled={pageNumber <= 1} onClick={handleFirstPageButtonClick}><BackwardOutlined /></Button>
+          <Button type="link" className="PdfViewer-pagerItem" disabled={pageNumber <= 1} onClick={handlePreviousPageButtonClick}><CaretLeftOutlined /></Button>
           <span className="PdfViewer-pagerItem">{pageNumber}&nbsp;/&nbsp;{pdf.numPages}</span>
-          <a className="PdfViewer-pagerItem" disabled={pageNumber >= pdf.numPages} onClick={handleNextPageButtonClick}><CaretRightOutlined /></a>
-          <a className="PdfViewer-pagerItem" disabled={pageNumber >= pdf.numPages} onClick={handleLastPageButtonClick}><ForwardOutlined /></a>
+          <Button type="link" className="PdfViewer-pagerItem" disabled={pageNumber >= pdf.numPages} onClick={handleNextPageButtonClick}><CaretRightOutlined /></Button>
+          <Button type="link" className="PdfViewer-pagerItem" disabled={pageNumber >= pdf.numPages} onClick={handleLastPageButtonClick}><ForwardOutlined /></Button>
```
ahelmberger, line 131: This fixes users previously being able to navigate out of bounds!

PR #342, opened 2022-04-11
EDU-489 Fix lessons sorting by schedule.startsOn

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-489

### DIFF
--- a/src/services/lesson-service.js
+++ b/src/services/lesson-service.js
@@ -22,7 +22,7 @@ class LessonService {
 
   async getLessonsMetadata(roomId) {
     const lessons = await this.lessonStore.getLessonsMetadataByRoomId(roomId);
-    return lessons.sort(by(l => l.schedule?.startsOn));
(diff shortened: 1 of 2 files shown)


src/services/lesson-service.js
```diff
@@ -22,7 +22,7 @@ class LessonService {
 
   async getLessonsMetadata(roomId) {
     const lessons = await this.lessonStore.getLessonsMetadataByRoomId(roomId);
-    return lessons.sort(by(l => l.schedule?.startsOn));
+    return lessons.sort(by(l => l.schedule?.startsOn || ''));
```
ahelmberger, line 25: Just a small thing: as we're comparing dates (which are compared as numbers), maybe `0` would be the better default value here than empty string?

adrianaluchian: Indeed, changed to `0`, thanks!

ahelmberger (approved):

PR #345, opened 2022-04-11
EDU-490 Compare email adresses in lower case

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-490

### DIFF
--- a/src/stores/user-store.js
+++ b/src/stores/user-store.js
@@ -12,10 +12,11 @@ class UserStore {
   }
 
   async findUserByLogin({ provider, emailOrUsername }, { session } = {}) {
+    const lowerCasedEmailOrUsername = emailOrUsername?.toLowerCase() || '';
     const matches = await this.collection.find({
       $and: [
         { provider },
-        { $or: [{ username: emailOrUsername }, { email: emailOrUsername }] }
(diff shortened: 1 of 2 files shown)


src/stores/user-store.js
```diff
@@ -12,10 +12,11 @@ class UserStore {
   }
 
   async findUserByLogin({ provider, emailOrUsername }, { session } = {}) {
+    const lowerCasedEmailOrUsername = emailOrUsername?.toLowerCase() || '';
```
ahelmberger, line 15: I realized this function has much more logic than the store should have (also with the `switch` below). I left it for the fix, but will create a ticket to put the decision logic into the service.

adrianaluchian: That's true, the best match decision does not belong here. Actually this method should rather be "getUserByEmailOrUsername" .

adrianaluchian (approved):

src/stores/user-store.js
adrianaluchian: Oh wait, but for username we allow upper case 🤔

ahelmberger: :+1:

adrianaluchian: Don't wait, I see you only used it for email 😆 sorry!

PR #352, opened 2022-04-13
EDU-472 Section cloning

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-472

I'm really sorry about the amount of changed files, I had to modify basically each plugin :-(


src/common/github-flavored-markdown.js
```diff
@@ -66,6 +76,28 @@ class GithubFlavoredMarkdown {
...
   }
+
+  redactCdnResources(markdown, cb) {
+    if (!markdown) {
+      return markdown;
+    }
+
+    const redact = url => {
+      if (url.startsWith('cdn://')) {
+        const pathOnly = url.slice(6);
```
adrianaluchian, line 87: Maybe we can extract
`const cdnUrlBasePath = 'cdn://'`
and use it in
`const pathOnly = url.slice(cdnUrlBasePath.length)`
so that this 6 is not a magic number and to reduce the risk of a nasty bug if the base url ever changes again 😬

src/common/github-flavored-markdown.js
```diff
@@ -4,6 +4,12 @@ import { escapeHtml } from '../utils/string-utils.js';
 const audioUrlPattern = new RegExp(`\\.(${['aac', 'm4a', 'mp3', 'oga', 'ogg', 'wav'].join('|')})$`, 'i');
 const videoUrlPattern = new RegExp(`\\.(${['mp4', 'm4v', 'ogv', 'webm', 'mpg', 'mpeg'].join('|')})$`, 'i');
 
+// Matches both URLs in e.g.: [![alt](cdn://image.png "image title")](cdn://some-target)
+const imageInsideOfHyperlinkPattern = /(\[!\[[^\]]*\]\()(\S*?)((?:\s+[^)]*)?\s*\)]\()(\S*?)((?:\s+[^)]*)?\s*\))(?!\])/g;
```
adrianaluchian, line 8: Jeeeesuuuus....

src/common/github-flavored-markdown.js
ahelmberger: Greate, idea, done, thanks!

src/common/github-flavored-markdown.js
ahelmberger: I'm so sorry, but I cannot think of anything better to do the job.

src/services/section-helper.js
```diff
@@ -110,3 +112,49 @@ export function extractCdnResources(sections, pluginInfoFactory) {
...
+function _clipboardDecode(clipboardText, encryptionKey) {
+  try {
+    const decryptedText = AES.decrypt(clipboardText, encryptionKey).toString(cryptoJsUtf8Encoder);
+    return JSON.parse(decryptedText);
+  } catch {
+    return null;
+  }
+}
+
+function _isValidClipboardObject(clipboardObject) {
```
adrianaluchian, line 130: Tiny thing, since this is specific for sections,  something like `_isValidClipboardSection` might be better.

ahelmberger: Sure, the "object" part is already in the param name anyway. Done.

adrianaluchian:
Do we want to make the icon fit in a bit better with a lighter grey, or will we ask for a custom icon and replace it? 
I think the ant one is fine, if we make it fit.

Screen Shot 2022-04-13 at 17 11 37

adrianaluchian:
Works really nicely! 👏

ahelmberger:
> Do we want to make the icon fit in a bit better with a lighter grey, or will we ask for a custom icon and replace it? I think the ant one is fine, if we make it fit.
> 
> Screen Shot 2022-04-13 at 17 11 37

I also think the Antd one is fine. I was thinking of a blue color first, but as we cannot colorize buttons the "paste" counterpart wouldn't fit, so I left it in the text color (which looks a lot darker because of the thicker stroke size). Should we maybe introduce our own "clipboard color"?

adrianaluchian:
> > Do we want to make the icon fit in a bit better with a lighter grey, or will we ask for a custom icon and replace it? I think the ant one is fine, if we make it fit.
> > Screen Shot 2022-04-13 at 17 11 37
> 
> I also think the Antd one is fine. I was thinking of a blue color first, but as we cannot colorize buttons the "paste" counterpart wouldn't fit, so I left it in the text color (which looks a lot darker because of the thicker stroke size). Should we maybe introduce our own "clipboard color"?

Honestly in this case I'd rather ask for a custom icon (that looks like this one basically) and have them all fit without a special grey for just one of them 🤔 
In a separate ticket, of course.

ahelmberger:
> > > Do we want to make the icon fit in a bit better with a lighter grey, or will we ask for a custom icon and replace it? I think the ant one is fine, if we make it fit.
> > > Screen Shot 2022-04-13 at 17 11 37
> > 
> > 
> > I also think the Antd one is fine. I was thinking of a blue color first, but as we cannot colorize buttons the "paste" counterpart wouldn't fit, so I left it in the text color (which looks a lot darker because of the thicker stroke size). Should we maybe introduce our own "clipboard color"?
> 
> Honestly in this case I'd rather ask for a custom icon (that looks like this one basically) and have them all fit without a special grey for just one of them thinking In a separate ticket, of course.

I asked Mani to provide us with a better one in the near future.

PR #353, opened 2022-04-13
EDU-492 Remove business logic from store

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-492

### DIFF
--- a/src/stores/user-store.js
+++ b/src/stores/user-store.js
@@ -11,30 +11,24 @@ class UserStore {
     return this.collection.find().toArray();
   }
 
-  async findUserByLogin({ provider, emailOrUsername }, { session } = {}) {
-    const lowerCasedEmailOrUsername = emailOrUsername?.toLowerCase() || '';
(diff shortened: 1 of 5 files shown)


src/stores/user-store.js
```diff
@@ -11,30 +11,24 @@ class UserStore {
...
+      queryFilters.push({ username });
+    }
+
+    if (!queryFilters.length) {
+      return [];
     }
 
-    return bestMatch;
+    return this.collection
+      .find({ $and: [{ provider }, queryFilters.length === 1 ? queryFilters[0] : { $or: queryFilters }] }, { session })
```
adrianaluchian, line 30: For the sake of simplification I think here you can simply say:
`.find({ $and: [{ provider },  { $or: queryFilters }] }, { session })`
even for the case of length being one

ahelmberger: True, that was a bit over-dogmatic. Simplified it, thanks!

PR #356, opened 2022-04-15
EDU-498 Allow private storage deletion for owners

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-498


src/components/pages/lesson.js
```diff
@@ -238,14 +247,19 @@ function Lesson({ PageTemplate, initialState }) {
...
+  const privateStorage = isPrivateRoom && !!room.owner.storage?.plan
     ? {
+      usedBytes: room.owner.storage.usedBytes,
+      maxBytes: room.owner.storagePlan.maxBytes,
       rootPath: `rooms/${room._id}/media`,
       initialPath: `rooms/${room._id}/media`,
-      uploadPath: `rooms/${room._id}/media`
+      uploadPath: `rooms/${room._id}/media`,
+      isDeletionEnabled: isRoomOwner || isRoomCollaborator,
+      onUsedBytesUpdated: handleUsedBytesUpdated
```
adrianaluchian, line 262: I opted for putting this callback here to be less intrusive with this patch fix 😬

src/domain/private-data.js
```diff
@@ -1,14 +0,0 @@
...
-
-export function getAllowedUserFields(forUser) {
-  return hasUserPermission(forUser, permissions.SEE_USER_EMAIL)
-    ? extendedFields
-    : basicFields;
-}
-
-export default {
-  getAllowedUserFields
-};
```
ahelmberger, line 14: :+1:

src/components/pages/lesson.js
ahelmberger: Good idea to not pollute all components in between!

src/server/lesson-controller.js
```diff
@@ -26,7 +26,7 @@ const { NotFound, BadRequest, Forbidden, Unauthorized } = httpErrors;
 
 const isRoomOwnerOrCollaborator = ({ room, userId }) => {
   const isOwner = room.owner === userId;
-  const isCollaborator = room.lessonsMode === ROOM_LESSONS_MODE.collaborative && room.members.find(m => m.userId === userId);
+  const isCollaborator = room.lessonsMode === ROOM_LESSONS_MODE.collaborative && room.members.some(m => m.userId === userId);
```
ahelmberger, line 29: The last place to use `find` would be line 67 in this file ;-)

ahelmberger (approved):

src/components/pages/lesson.js
adrianaluchian: Was not original :D

src/server/lesson-controller.js
adrianaluchian: As in 'it should be `some` there too'? If so, updated, if not, I need more info 🙏

ahelmberger: Yes exactly (although not really very important), thanks!

adrianaluchian: I think it's important cause we could end up with surprises if we rely on a boolean value. Thanks for pointing it out.

PR #357, opened 2022-04-15
EDU-499 Show searching headlines while fetching documents

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-499

### DIFF
--- a/src/components/pages/search.js
+++ b/src/components/pages/search.js
@@ -152,11 +152,14 @@ function Search({ PageTemplate }) {
     }
   ];
 
+  const showSearchingHeadline = isSearching && documents.length === 0;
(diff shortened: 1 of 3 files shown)


src/components/pages/search.js
```diff
@@ -152,11 +152,14 @@ function Search({ PageTemplate }) {
     }
   ];
 
+  const showSearchingHeadline = isSearching && documents.length === 0;
```
adrianaluchian, line 155: Added the `documents.length === 0` restriction as well so that we don't see the headline briefly changing each time the search text changes

ahelmberger (approved):

PR #363, opened 2022-04-20
EDU-476 Extend annotation plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-476

and looks like:

Screen Shot 2022-04-20 at 16 34 55


src/plugins/annotation/annotation-info.js
```diff
@@ -35,7 +36,10 @@ class AnnotationInfo {
     return {
       title: `[${t('common:title')}]`,
       text: `[${t('common:text')}]`,
-      renderMedia: false
+      renderMedia: false,
+      state: STATE.collapsed,
```
ahelmberger, line 40: Maybe `initialState` would be more logical, as 2 of these states can change later (without being persisted in the DB)? Other options would be `displayState` (or just `display`?) - or having an enum called `behavior` with values `static`, `expandable` (meaning: initially collapsed) and `collapsible` (meaning: initially expanded).

ahelmberger (approved):

src/plugins/annotation/annotation-info.js
adrianaluchian: Thanks for the suggestion,  I like the `behavior` approach as "initial" in `initialState` does not fit the `static` option.
Changed!

PR #365, opened 2022-04-22
EDU-519 Create first interactive media editor UI

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-519


src/plugins/abc-notation/abc-notation-editor.js
```diff
@@ -16,7 +16,7 @@ function AbcNotationEditor({ content, onContentChanged }) {
   };
 
   const changeContent = newContentValues => {
-    onContentChanged({ ...content, ...newContentValues });
+    onContentChanged({ ...content, ...newContentValues }, false);
```
adrianaluchian, line 19: I added this value for the second ("isInvalid") argument so that we see it when we need to add validation in a plugin

src/plugins/image-tiles/image-tiles-editor.js
```diff
@@ -48,8 +48,8 @@ function ImageTilesEditor({ content, onContentChanged, publicStorage, privateSto
     }
   });
 
-  const changeContent = newContentValues => {
-    onContentChanged({ ...content, ...newContentValues });
+  const changeContent = (newContentValues, isInvalid) => {
+    onContentChanged({ ...content, ...newContentValues }, isInvalid);
```
adrianaluchian, line 52: This partly works because:
- if you add a tile and invalidate it's sourceUrl
- then change anything outside the tile
- the page state is set to valid again

Anyway I think we should reconsider how we validate, because by the way we set the ant validation in plugins we cannot leverage the error status of the UI (`form.getFieldsError` does not work in this setup) and we end up having to duplicate the validation **and** re-validate on **any** change in the plugin. 

We might probably need to consider using the `rules` setup, I guess 🤔

src/plugins/abc-notation/abc-notation-editor.js
ahelmberger: :+1: Good idea!

src/plugins/image-tiles/image-tiles-editor.js
ahelmberger: As a solution for now I moved the validation into the main editor, so that on each change all tiles are validated. I agree that we should come up with a unified approach concerning plugin validation.

PR #368, opened 2022-04-29
EDU-505 Change several icons

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-505

With current/initial PR code, the annotations look as follows:
Screen Shot 2022-04-29 at 11 36 51

Alternatively annotation icons without background would look as follows:
Screen Shot 2022-04-29 at 11 39 06

With current/initial PR code, the copy+paste icon looks as follows:
Screen Shot 2022-04-29 at 11 50 00
Screen Shot 2022-04-29 at 11 50 10

Alternatively using the second copy+paste icon would look as follows:
Screen Shot 2022-04-29 at 11 57 37
Screen Shot 2022-04-29 at 11 57 55

What are your thoughts on the alternatives?


ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-505
> 
> With current/initial PR code, the annotations look as follows: Screen Shot 2022-04-29 at 11 36 51
> 
> Alternatively annotation icons without background would look as follows: Screen Shot 2022-04-29 at 11 39 06
> 
> With current/initial PR code, the copy+paste icon looks as follows: Screen Shot 2022-04-29 at 11 50 00 Screen Shot 2022-04-29 at 11 50 10
> 
> Alternatively using the second copy+paste icon would look as follows: Screen Shot 2022-04-29 at 11 57 37 Screen Shot 2022-04-29 at 11 57 55
> 
> What are your thoughts on the alternatives?

Honestly, I really don't like the light gray background, so I'd prefer the monotone version for the annotations. Concerning the clipboard icon, I would go with the one you proposed in the PR (the centipede-looking one) until we have a better solution.

adrianaluchian:
> > Closes https://educandu.atlassian.net/browse/EDU-505
> > [.....]
> > What are your thoughts on the alternatives?
> 
> Honestly, I really don't like the light gray background, so I'd prefer the monotone version for the annotations. Concerning the clipboard icon, I would go with the one you proposed in the PR (the centipede-looking one) until we have a better solution.

I agree. Will remove the background for the annotation ones and for clipboard let's then keep the icon we have from AntDesign now and replace when we get a better one, cause I think the current one is much more intuitive.

adrianaluchian:
> > > Closes https://educandu.atlassian.net/browse/EDU-505
> > > [.....]
> > > What are your thoughts on the alternatives?
> > 
> > 
> > Honestly, I really don't like the light gray background, so I'd prefer the monotone version for the annotations. Concerning the clipboard icon, I would go with the one you proposed in the PR (the centipede-looking one) until we have a better solution.
> 
> I agree. Will remove the background for the annotation ones and for clipboard let's then keep the icon we have from AntDesign now and replace when we get a better one, cause I think the current one is much more intuitive.

Done!

ahelmberger (approved):

PR #369, opened 2022-04-29
EDU-520 Render hover image in canvas of same size as the original image

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-520

### DIFF
--- a/src/plugins/image/image-display.js
+++ b/src/plugins/image/image-display.js
@@ -9,6 +9,8 @@ import { sectionDisplayProps } from '../../ui/default-prop-types.js';
 import { ReactCompareSlider, ReactCompareSliderImage } from 'react-compare-slider';
 
 function ImageDisplay({ content }) {
+  const mainImageRef = useRef();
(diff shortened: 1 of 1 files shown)


ahelmberger:
Strangly the image still flickers when the section is in preview mode (but the page is in edit mode)

adrianaluchian:
> Strangly the image still flickers when the section is in preview mode (but the page is in edit mode). It only happens once switched into edit and then back into preview mode 🤔

I can't reproduce it, I probably need to use your pictures for testing, because the different image sizes combinations might have a different effect. Let's maybe check this one together.

adrianaluchian:
> > Strangly the image still flickers when the section is in preview mode (but the page is in edit mode). It only happens once switched into edit and then back into preview mode 🤔
> 
> I can't reproduce it, I probably need to use your pictures for testing, because the different image sizes combinations might have a different effect. Let's maybe check this one together.

Thanks for finding this! 🙏

ahelmberger (approved):

PR #370, opened 2022-04-29
EDU-503 cell merging in table plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-503


src/plugins/table/table-designer-menu.js
```diff
@@ -58,7 +66,12 @@ function TableDesignerMenu({ canDeleteColumn, canDeleteRow, cell, dotType, onCel
     createMenuItem(t, DESIGNER_CELL_ACTION.deleteRow, cell, onCellAction, !canDeleteRow, true),
     createMenuItem(t, DESIGNER_CELL_ACTION.insertColumnBefore, cell, onCellAction, false),
     createMenuItem(t, DESIGNER_CELL_ACTION.insertColumnAfter, cell, onCellAction, false),
-    createMenuItem(t, DESIGNER_CELL_ACTION.deleteColumn, cell, !canDeleteColumn, false)
+    createMenuItem(t, DESIGNER_CELL_ACTION.deleteColumn, cell, onCellAction, !canDeleteColumn, true),
+    createMenuItem(t, DESIGNER_CELL_ACTION.connectToRowBefore, cell, onCellAction, cell.isFirstInColumn),
+    createMenuItem(t, DESIGNER_CELL_ACTION.connectToRowAfter, cell, onCellAction, cell.isLastInColumn),
+    createMenuItem(t, DESIGNER_CELL_ACTION.connectToColumnBefore, cell, onCellAction, cell.isFirstInRow),
+    createMenuItem(t, DESIGNER_CELL_ACTION.connectToColumnAfter, cell, onCellAction, cell.isLastInRow),
+    createMenuItem(t, DESIGNER_CELL_ACTION.disconnectCell, cell, onCellAction, !cell.isConnected)
```
adrianaluchian, line 74: In such long signature methods called so many times I find it's easy to introduce a call with wrong values.
I'd suggest to either use named arguments instead of positional ones, or always pass all arguments including the default ones. What do you think?

src/plugins/table/table.yml
```diff
@@ -19,3 +19,18 @@ cellAction_insertColumnAfter:
...
   en: Delete column
   de: Spalte löschen
+cellAction_connectToRowBefore:
+  en: Connect with row before
+  de: Mit Zeile davor verbinden
+cellAction_connectToRowAfter:
+  en: Connect with row after
+  de: Mit Zeile danach verbinden
+cellAction_connectToColumnBefore:
+  en: Connect with column before
```
adrianaluchian, line 29: How about "Connect with previous column", "Connect with following row" etc. for a more natural phrasing? 🤔

src/plugins/table/table.yml
```diff
@@ -19,3 +19,18 @@ cellAction_insertColumnAfter:
...
+  de: Mit Spalte davor verbinden
+cellAction_connectToColumnAfter:
+  en: Connect with column after
+  de: Mit Spalte danach verbinden
+cellAction_disconnectCell:
+  en: Revoke cell connection
```
adrianaluchian, line 35: "Disconnect cells" here maybe?

src/plugins/table/table-display.js
```diff
@@ -1,13 +1,13 @@
...
 import Markdown from '../../components/markdown.js';
-import { mapTwoDimensionalArray } from './table-utils.js';
 import { sectionDisplayProps } from '../../ui/default-prop-types.js';
 
 function TableDisplay({ content }) {
   const { rowCount, columnCount, cells, renderMedia } = content;
 
   const rows = useMemo(() => {
-    const fullCellMap = mapTwoDimensionalArray(rowCount, columnCount, () => null);
+    const fullCellMap = mapCellsNested(rowCount, columnCount, () => null);
```
adrianaluchian, line 10: I'm having a bit of a hard time understanding the "mapCellsFlat" and "mapCellsNested" ideas, I don't quite understand how they are mapped nested or flat

adrianaluchian: 💡 Ok, so they are both the methods that basically create the table, either flat or normal. 
How about something like:
- "createFlatTable({ rowCount, columnCount, createCell })"
- "createNormalTable({ rowCount, columnCount, createCell })"
to help understand the code easier?

src/plugins/table/table-designer-menu.js
ahelmberger: Agree, done. I also moved this function into the component, so it can be reduced to 3 arguments.

src/plugins/table/table.yml
ahelmberger: Actually I realized the "row" and "columns" might be confusing for the user. So I renamed them to "Connect with cell above/below/on the left/on the right" and the counterpart in the German (I left the enumeration values as they were though, I think technically they're not wrong)

src/plugins/table/table.yml
ahelmberger: Good idea, rephrased it.

src/plugins/table/table-display.js
ahelmberger: As the term "table" refers already to a bigger model (that does not only contain the cells collection), I went for `createTableCellsFlat` and `createTableCellsInRows`. What do you think?

src/plugins/table/table-designer-menu.js
adrianaluchian: Great, even better! 👍

src/plugins/table/table.yml
adrianaluchian: That's true, "cells" are what the user connects, but I agree that programmatically "rows" and "columns" are what we process. Sounds good.

src/plugins/table/table-display.js
adrianaluchian: Much clearer, thanks. The "map" part of it was confusing, especially since it wan't actually mapping but creating. Maybe another small thing, we could rename the 3rd argument from `mappingFunction` to `createCell`?

ahelmberger: Yeah, that should have been the logical consequence in the first place :roll_eyes: Renamed it, thanks!

adrianaluchian (approved):

PR #379, opened 2022-05-04
EDU-504 table plugin additional features

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-504


src/plugins/table/table-utils.js
```diff
@@ -24,8 +40,23 @@ export const DESIGNER_CELL_ACTION = {
   disconnectCell: 'disconnect-cell'
 };
 
+export function calculateEvenColumnWidthsInPercent(columnCount) {
```
adrianaluchian, line 43: If there's any risk `columnCount` could be zero, let's maybe default it to `1` here?

adrianaluchian, line 43: If it's not a huge hassle, some tests for this could come in handy.

ahelmberger: Should never happen, but I added an if clause that'll return empty array in this case.

adrianaluchian (approved):

PR #380, opened 2022-05-05
EDU-531 Get media duration

adrianaluchian:
https://educandu.atlassian.net/browse/EDU-531

### DIFF
--- a/src/plugins/interactive-media/interactive-media-editor.js
+++ b/src/plugins/interactive-media/interactive-media-editor.js
@@ -1,16 +1,21 @@
 import by from 'thenby';
+import ReactDOM from 'react-dom';
+import reactPlayerNs from 'react-player';
 import { useTranslation } from 'react-i18next';
+import uniqueId from '../../utils/unique-id.js';
 import validation from '../../ui/validation.js';
-import React, { Fragment, useState } from 'react';
 import Timeline from '../../components/timeline.js';
 import { Form, Input, Radio, Spin, Switch } from 'antd';
 import { MEDIA_TYPE, SOURCE_TYPE } from './constants.js';
 import { removeItemAt } from '../../utils/array-utils.js';
+import React, { Fragment, useRef, useState } from 'react';
 import ClientConfig from '../../bootstrap/client-config.js';
 import { useService } from '../../components/container-context.js';
 import { sectionEditorProps } from '../../ui/default-prop-types.js';
 import StorageFilePicker from '../../components/storage-file-picker.js';
 import ObjectMaxWidthSlider from '../../components/object-max-width-slider.js';
+
+const ReactPlayer = reactPlayerNs.default || reactPlayerNs;
 
 const FormItem = Form.Item;
 const RadioGroup = Radio.Group;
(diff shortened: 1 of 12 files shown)


src/plugins/interactive-media/interactive-media-editor.js
```diff
@@ -1,102 +1,111 @@
...
 
 function InteractiveMediaEditor({ content, onContentChanged, publicStorage, privateStorage }) {
   const clientConfig = useService(ClientConfig);
+  const hiddenPlayerContainerRef = useRef(null);
   const { t } = useTranslation('interactiveMedia');
 
-  const { sourceType, sourceUrl, text, width, aspectRatio, showVideo } = content;
+  const getAspectRatioText = givenAspectRatio => `${givenAspectRatio.h}:${givenAspectRatio.v}`;
+  const ensureChaptersOrder = chapters => chapters.sort(by(chapter => chapter.startTimecode));
+  const getDefaultChapter = () => ({ key: uniqueId.create(), title: t('defaultChapterTitle'), startTimecode: 0 });
```
ahelmberger, line 41: I think we should move this function into `interactive-media-info.js` (and also use it there when generating the whole default content).

src/plugins/interactive-media/interactive-media-editor.js
```diff
@@ -1,102 +1,111 @@
...
+  const getMediaType = path => {
+    const sanitizedPath = (path || '').trim();
+    const extensionMatches = sanitizedPath.match(/\.([0-9a-z]+)$/i);
+    const extension = extensionMatches && extensionMatches[1];
+
+    if (!extension) {
+      return MEDIA_TYPE.none;
+    }
+    if (audioTypes.includes(extension)) {
+      return MEDIA_TYPE.audio;
```
ahelmberger, line 56: I know this was there before, but I just realized we have (pretty much) the same code in `github-flavored-markdown.js`. Maybe you wanna check if we could easily unify those two places (there are a few more formats there, but they all should work)?

src/plugins/interactive-media/interactive-media-editor.js
adrianaluchian: Good idea, moved it there.

src/plugins/interactive-media/interactive-media-editor.js
adrianaluchian: Nice, I didn't know about this other place, thanks 👍 
Unified the media type identification.

src/utils/media-utils.js
```diff
@@ -0,0 +1,21 @@
+import { MEDIA_TYPE } from '../domain/constants.js';
+
+const audioUrlExtensions = ['aac', 'm4a', 'mp3', 'oga', 'ogg', 'wav', 'flac'];
+const videoUrlExtensions = ['mp4', 'm4v', 'ogv', 'webm', 'mpg', 'mpeg', 'mov', 'avi', 'mkv'];
+
+export const getMediaType = url => {
+  const sanitizedUrl = (url || '').trim();
+  const extensionMatches = sanitizedUrl.match(/\.([0-9a-z]+)$/i);
+  const extension = extensionMatches && extensionMatches[1];
```
ahelmberger, line 9: Just to be on the safe side I'd propose `const extension = extensionMatches?.[1]?.toLowerCase();` realizing that some applications produce files like `MYSOUND.MP3`

adrianaluchian: Oh of course, the regex that was in the markdown plugin was case insensitive 🤦‍♀️ 
Fixed, thanks for noticing it!

ahelmberger (approved):

PR #381, opened 2022-05-05
EDU-529 Create new media player and use it within plugins

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-529


src/components/media-player-controls.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { Button, Slider } from 'antd';
+import { AudioMutedOutlined, CaretRightOutlined, PauseOutlined, SoundOutlined } from '@ant-design/icons';
+
+function formatMillisecondsAsDuration(milliseconds) {
```
adrianaluchian, line 7: Now that we have the `media-utils` we could put this in there too and throw in a couple of tests.

src/plugins/anavis/anavis-display.js
```diff
@@ -174,28 +59,31 @@ function AnavisDisplay({ content }) {
...
-        </div>)}
-
-      {media.text && (
-        <div className="Anavis-text">
-          <Markdown>{media.text}</Markdown>
-        </div>
-      )}
+      <div className={`AnavisDisplay-players u-width-${width || 100}`}>
+        {sourceUrl && (
+          <MediaPlayer
```
adrianaluchian, line 75: Nice! Did not realize the player was also used here.

adrianaluchian:
Love the simplification in all related plugins! 😍

src/components/media-player-controls.js
ahelmberger: Good idea, thanks! It was put and tests were thrown :-)

adrianaluchian (approved):

PR #388, opened 2022-05-06
EDU-539 Disable link functionality on image tiles with no link set

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-539

### DIFF
--- a/src/plugins/image-tiles/image-tiles-display.js
+++ b/src/plugins/image-tiles/image-tiles-display.js
@@ -41,8 +41,8 @@ function ImageTilesDisplay({ content }) {
     const classes = classNames({
       'ImageTiles-tile': true,
       'ImageTiles-tile--noLink': !linkUrl,
-      'u-img-color-flip': content.hoverEffect === HOVER_EFFECT.colorizeZoom,
-      'u-img-color-flip-disabled': content.hoverEffect === HOVER_EFFECT.colorizeZoom && !linkUrl
+      'u-img-color-flip': content.hoverEffect === HOVER_EFFECT.colorizeZoom && linkUrl,
+      'u-img-color-flip-hover-disabled': content.hoverEffect === HOVER_EFFECT.colorizeZoom && !linkUrl
     });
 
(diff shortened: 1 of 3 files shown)


src/plugins/image-tiles/image-tiles-display.js
```diff
@@ -7,53 +7,50 @@ import { useService } from '../../components/container-context.js';
...
   const renderTile = (tile, index) => {
+    const linkUrl = getTileLinkUrl(tile);
+    const imageUrl = getTileImageUrl(tile, clientConfig.cdnRootUrl);
+
     const classes = classNames({
       'ImageTiles-tile': true,
-      'u-img-color-flip': content.hoverEffect === HOVER_EFFECT.colorizeZoom
+      'ImageTiles-tile--noLink': !linkUrl,
+      'u-img-color-flip': content.hoverEffect === HOVER_EFFECT.colorizeZoom,
+      'u-img-color-flip-disabled': content.hoverEffect === HOVER_EFFECT.colorizeZoom && !linkUrl
```
adrianaluchian, line 45: I guess it would not be enough to set
```
'u-img-color-flip': !!linkUrl && content.hoverEffect === HOVER_EFFECT.colorizeZoom
```
and not set `'u-img-color-flip-disabled'` at all, but I don't quite see why 🤔

adrianaluchian: Ah ok, cause only the hover has to be removed but the effect gets applied even when disabled.

ahelmberger: Because we still want the image to be grayscaled but not zoom on hover (so only one part of the style class). But now, as you asked, I think it might be a better idea to separate the current CSS class into two separately applicable classes (one for grayscale, one for zoom). I will do that, thanks for triggering the thought :smile:

ahelmberger: I followed your suggestion there, as it would have been more convoluted to separate the grayscale and the zoom.

adrianaluchian (approved):

PR #392, opened 2022-05-09
EDU-518 Create document picker

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-518


src/services/document-service.js
```diff
@@ -104,6 +104,19 @@ class DocumentService {
     return this.documentStore.getDocumentTagsMatchingText(searchString);
   }
 
+  async findDocumentsMetadata(query) {
+    const sanitizedQuery = query.trim();
+
+    const queryConditions = [
+      { archived: false },
+      { title: { $regex: `^${sanitizedQuery}`, $options: 'i' } }
```
ahelmberger, line 112: When I query "Film (Kopie" I get `name: 'MongoServerError', status: 500, message: 'Regular expression is invalid: missing )', code: 51091...`, because the opening parenthesis is interpreted as RegExp special character. We need to regex-encode the input here.

ahelmberger, line 112: A second thought: Have you tried removing the `^` from the regex, so users can search for any part of the title (it's hard to remember sometimes how the title exactly starts, I guess)? If that would work too, I'd prefer that.

src/components/document-picker.js
```diff
@@ -0,0 +1,86 @@
...
+import { useTranslation } from 'react-i18next';
+import React, { useEffect, useState } from 'react';
+import { useService } from './container-context.js';
+import DocumentApiClient from '../api-clients/document-api-client.js';
+
+const { Option } = Select;
+
+function DocumentPicker({ documentId, onChange }) {
+  const { t } = useTranslation('documentPicker');
+  const documentApiClient = useService(DocumentApiClient);
```
ahelmberger, line 12: Maybe better `useSessionAwareApiClient` as the endpoint requires to be authenticated

src/services/document-service.js
adrianaluchian: Indeed, good catch, thank you!
Fixed.

adrianaluchian: I thought that for users that actually type the title from the beginning it would be strange to not find it on the top. 
But I updated as you suggested and we can see what feedback we get (if any).

src/components/document-picker.js
adrianaluchian: Yes of course, changed, thanks!

adrianaluchian:
Now that the old `document-selector` is gone, I will rename this `document-picker` to `document-selector`, cause all our other such components are `selector`s 🤔

src/services/document-service.js
ahelmberger: :+1:

src/components/document-picker.js
```diff
@@ -0,0 +1,86 @@
...
+      className=""
+      allowClear
+      loading={loading}
+      filterOption={false}
+      value={selectedDocument}
+      onClear={handleClear}
+      onSearch={handleSearch}
+      onSelect={handleSelect}
+      notFoundContent={null}
+      placeholder={t('common:searchPlaceholder')}
```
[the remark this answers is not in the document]
ahelmberger: Sounds good, thanks!

src/services/document-service.js
```diff
@@ -104,6 +104,19 @@ class DocumentService {
     return this.documentStore.getDocumentTagsMatchingText(searchString);
   }
 
+  async findDocumentsMetadata(query) {
+    const sanitizedQuery = escapeStringRegexp(query.trim());
+
+    const queryConditions = [
+      { archived: false },
+      { title: { $regex: `${sanitizedQuery}`, $options: 'i' } }
```
ahelmberger, line 112: Just a tiny thing: there's no need for the template string anymore :-)

adrianaluchian: Gone!

ahelmberger (approved):

PR #398, opened 2022-05-11
EDU-551 Create files grid viewer component

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-551

**Note:**
Apologies for the escalations while sorting out some of the file.. Due to this and the fact that we risk overlapping at this point, I drew the line on this ticket on a very simplistic version.


src/components/files-grid-viewer.less
```diff
@@ -0,0 +1,42 @@
...
+
+    .FilesGridViewer-fileIcon {
+      font-size: 65px;
+      padding-top: 5px;
+    }
+  }
+}
+
+.FilesGridViewer-fileIcon {
+  font-size: 60px;
```
ahelmberger, line 40: We could put a `transition: all .2s ease-in-out;` here in order to make the hover effect less "jumpy".

src/components/files-grid-viewer.less
```diff
@@ -0,0 +1,42 @@
...
+@files-grid-viewer-file-container-total-width: @files-grid-viewer-file-container-base-width + 2 * @files-grid-viewer-file-icon-padding;
+
+.FilesGridViewer {
+  display: flex;
+  gap: 20px;
+  flex-wrap: wrap;
+  flex-direction: row;
+  justify-content: flex-start;
+  padding: @edu-content-padding;
+  border: 1px solid @edu-border-color-base;
```
ahelmberger, line 12: `text-align: center;` here would look better (in case of the short "..."), what dou you think?

PR #401, opened 2022-05-16
EDU-558 Fix UI issues with media player

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-558

Screen Shot 2022-05-16 at 09 22 37

### DIFF
--- a/src/plugins/ear-training/ear-training.less
+++ b/src/plugins/ear-training/ear-training.less
@@ -1,5 +1,10 @@
 .EarTraining {}
 
+.EarTraining-soundPlayer {
+  .u-width-50;
+  .m-horizontally-centered;
+}
+
 .EarTraining-testWrapper {
   margin: 0 auto;
   max-width: 100%;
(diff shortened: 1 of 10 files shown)


src/plugins/ear-training/ear-training.less
```diff
@@ -1,5 +1,10 @@
 .EarTraining {}
 
+.EarTraining-soundPlayer {
+  .u-width-50;
```
ahelmberger, line 4: I'm not sure if we use a utility class as a mixin anywhere else in the code, I'm kinda OK with it (although we also could have just `width: 50%;` here).

ahelmberger: EDIT: Little bit of history: Those `u-width-xy` classes exist primarily to be applied dynamically when plugins render (according what the user has set).

ahelmberger (approved):

src/plugins/ear-training/ear-training.less
ahelmberger: :+1:

PR #405, opened 2022-05-16
EDU-554 file preview component

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-554


src/components/credits-footer.js
```diff
@@ -29,22 +28,8 @@ function CreditsFooter({ doc, revision }) {
...
-          </Fragment>
-        ))}
-      </Fragment>
-    );
-  };
-
-  const renderUrl = () => (<a rel="noopener noreferrer" target="_blank" href={url}>{renderLongUrlText(url)}</a>);
-  const renderOriginalUrl = () => (<a rel="noopener noreferrer" target="_blank" href={originalUrl}>{renderLongUrlText(originalUrl)}</a>);
+  const renderUrl = () => <LiteralUrlLink rel="noopener noreferrer" target="_blank" href={url} />;
+  const renderOriginalUrl = () => <LiteralUrlLink rel="noopener noreferrer" target="_blank" href={originalUrl} />;
```
adrianaluchian, line 32: I guess the `rel` attribute could go inside the component, cause we always try to set it to "noopener noreferrer"

src/resources/common.yml
```diff
@@ -123,6 +123,15 @@ sectionCopiedToClipboard:
 clipboardPasteNotSupported:
   en: Pasting from the clipboard is not supported by this browser.
   de: Das Einfügen aus der Zwischenablage wird von diesem Browser nicht unterstützt.
+copyUrlToClipboard:
+  en: Copy URL into the clipboard
```
adrianaluchian, line 127: I would say "Copy ... to clipboard" sounds more natural than "Copy ... into the clipboard", in the English version. Sounds a bit translated from German to me 😋

src/components/file-preview.less
```diff
@@ -0,0 +1,68 @@
...
+  grid-row: 1;
+  grid-column-start: 1;
+  grid-column-end: 3;
+}
+
+.FilePreview-previewArea {
+  grid-row: 2;
+  grid-column: 1;
+  padding: 1em;
+  height: 24em;
```
adrianaluchian, line 18: I feel like we're getting pretty inconsistent with the units we use for dimensions and positioning. Do you think it's absolutely needed in this case to work with `em` (I understand the arguments for this, generally), or could we stick to `px` here? 🤔

src/components/file-preview.js
```diff
@@ -0,0 +1,214 @@
...
+import ClientConfig from '../bootstrap/client-config.js';
+import { getResourceType } from '../utils/resource-utils.js';
+import { FileOutlined, FileTextOutlined } from '@ant-design/icons';
+import CopyToClipboardIcon from './icons/general/copy-to-clipboard-icon.js';
+import React, { Fragment, useEffect, useMemo, useRef, useState } from 'react';
+import PdfDocument, { PDF_DOCUMENT_STRETCH_DIRECTION } from './pdf-document.js';
+
+const logger = new Logger(import.meta.url);
+
+function FilePreview({ lastModified, size, url }) {
```
adrianaluchian, line 23: Maybe `lastModified` could rather be `createdOn`, cause the label used to display it is `"common:createdOn"`.
Or maybe `uploadedOn` would actually be even better (with a corresponding label).

src/components/file-preview.js
```diff
@@ -0,0 +1,214 @@
...
+  );
+}
+
+FilePreview.propTypes = {
+  lastModified: PropTypes.instanceOf(Date).isRequired,
+  size: PropTypes.number.isRequired,
+  url: PropTypes.string.isRequired
+};
+
+export default FilePreview;
```
adrianaluchian, line 214: Super nice preview setup, seeing all the info there makes me realize this will be an extremely useful piece of transparency added to the CDN resources management 👍 

From UI perspective, I would have a suggestion for you to check out and see how you feel about it:
Screen Shot 2022-05-17 at 10 19 24

src/components/credits-footer.js
ahelmberger: It's a side effect of having `target="_blank"`, otherwise it wouldn't be there. But I created a `targetBlank` prop in the `LiteralUrlLink` component that puts either both or none there, so we do not have to specify it everywhere (we only use it so far with `target="_blank"` anyways).

src/resources/common.yml
ahelmberger: True, corrected (also in some old places)

src/components/file-preview.less
ahelmberger: Sure, no problem, switched most of them to `px` (except for font related rules)

src/components/file-preview.js
ahelmberger: Changed. The reason I used `lastModified` was that it is what we get from the server in the file object. But I find `createdOn` better, too (as long as we consider overriding a file with the same name in private storage a new "creation"). We should then change the server response accordingly once we got rid of the old browser.

src/utils/resource-utils.js
```diff
@@ -5,7 +5,7 @@ const extensionsGroups = [
   { type: RESOURCE_TYPE.pdf, extensions: ['pdf'] },
   { type: RESOURCE_TYPE.text, extensions: ['txt', 'doc', 'rtf', 'odt'] },
   { type: RESOURCE_TYPE.audio, extensions: ['aac', 'm4a', 'mp3', 'oga', 'ogg', 'wav', 'flac'] },
-  { type: RESOURCE_TYPE.image, extensions: ['jpg', 'jpeg', 'gif', 'png', 'tiff', 'raw', 'webp'] },
+  { type: RESOURCE_TYPE.image, extensions: ['jpg', 'jpeg', 'gif', 'png', 'tiff', 'raw', 'webp', 'svg'] },
```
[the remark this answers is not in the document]
ahelmberger: It would be beneficial to get rid of it, as it contributes a nice share to the bundle size (thanks to the `mime` package). The functionality of `mimeTypeHelper` I'd like keep though is to recognize the 20-30 most commonly used file extensions. I suggest we create a ticket to extract this into home-made code and then remove the package, whatcha think?

src/components/file-preview.js
ahelmberger: I like it better next to the label, thanks, good suggestion, I moved it.

src/components/credits-footer.js
adrianaluchian: Ah true, good idea 👍

src/components/file-preview.js
adrianaluchian: 👍

src/utils/resource-utils.js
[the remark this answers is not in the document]
adrianaluchian: Sounds good!

src/components/file-preview.js
adrianaluchian: Nice 👍 
What do you think about the other changes in the screenshot?
- more space between the preview and the text column
- removing the file name from the top
- inverting the background colors (white for preview and grey for text)
- centering the text column vertically

ahelmberger: Ha! Now I see the other changes :rofl:

Yeah, that looks great, I'll do them. I don't think we should set the gray background color for the preview component though, that might look weird when placed somewhere as it is not a confined block. We can always do that in the parent container (I'll make the preview box have an explicit white background though in order to stand out).

ahelmberger: Done, now it looks like this:

file-preview

adrianaluchian (approved):

PR #414, opened 2022-05-19
EDU-568 Show real image instead of icon

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-568

### DIFF
--- a/src/components/files-grid-viewer.js
+++ b/src/components/files-grid-viewer.js
@@ -45,7 +45,7 @@ function FilesGridViewer({
 
   const renderFile = file => {
     let fileDisplay;
-    if (!file.isDirectory && file.url && getResourceType(file.url) === RESOURCE_TYPE.image) {
+    if (getResourceType(file.url) === RESOURCE_TYPE.image) {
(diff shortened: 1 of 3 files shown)


src/components/files-grid-viewer.js
```diff
@@ -43,15 +44,21 @@ function FilesGridViewer({
   };
 
   const renderFile = file => {
-    const Icon = getResourceIcon({ filePath: file.path, isDirectory: file.isDirectory, filled: true });
+    let fileDisplay;
+    if (!file.isDirectory && file.url && getResourceType(file.url) === RESOURCE_TYPE.image) {
```
adrianaluchian, line 48: Here you can reduce the condition to `getResourceType(file.url) === RESOURCE_TYPE.image` because it will return `RESOURCE_TYPE.none` for any url you give it that does not end in a file extension (including undefined and null).

ahelmberger: True, thanks, I removed the second condition.

adrianaluchian (approved):

PR #416, opened 2022-05-19
EDU548 - Show overlay on chapter end

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-548

### DIFF
--- a/src/components/media-player.js
+++ b/src/components/media-player.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
-import React, { useRef, useState } from 'react';
 import MediaPlayerTrack from './media-player-track.js';
 import MediaPlayerControls from './media-player-controls.js';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { MEDIA_ASPECT_RATIO, MEDIA_PLAY_STATE } from '../domain/constants.js';
 
 function MediaPlayer({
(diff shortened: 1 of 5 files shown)


src/components/media-player.js
```diff
@@ -13,14 +13,33 @@ function MediaPlayer({
...
-  marks
+  marks,
+  onMarkReached,
+  onEndReached,
+  mediaPlayerRef
 }) {
   const trackRef = useRef();
   const [volume, setVolume] = useState(1);
   const [isMuted, setIsMuted] = useState(false);
+  const [lastReachedMark, setLastReachedMark] = useState();
```
ahelmberger, line 24: Not really a big issue, but I think we usually put something as a start value for `useState` (later on you use `null` to indicate absence of data), so maybe we could be more explicit and also initialize it with `null`?

src/plugins/interactive-media/interactive-media-display.js
```diff
@@ -1,14 +1,18 @@
...
 import { MEDIA_SOURCE_TYPE } from '../../domain/constants.js';
 import { useService } from '../../components/container-context.js';
 import { sectionDisplayProps } from '../../ui/default-prop-types.js';
+import { StepForwardOutlined, UndoOutlined } from '@ant-design/icons';
 import { formatMillisecondsAsDuration } from '../../utils/media-utils.js';
 
 function InteractiveMediaDisplay({ content }) {
+  const mediaPlayerRef = useRef();
   const clientConfig = useService(ClientConfig);
+  const [interactingChapterIndex, setInteractingChapterIndex] = useState();
```
ahelmberger, line 15: Maybe we can also have an initial value here.

ahelmberger:
Very nice and short solution :heart:

ahelmberger (approved):

PR #423, opened 2022-05-24
EDU-553 Prepare new storage browser

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-553

Sorry for the huge pull request. I started to modernize the server API and tried to keep it as tame as possible. There are more issues that should be fixed, but I created individual JIRA tickets for them.

### DIFF
--- a/src/api-clients/storage-api-client.js
+++ b/src/api-clients/storage-api-client.js
@@ -36,10 +36,10 @@ class StorageApiClient {
     return request.then(res => res.data);
   }
 
-  deleteCdnObject(fullPath) {
+  deleteCdnObject(path) {
     return this.httpClient
       .delete(
-        `/api/v1/storage/objects?path=${encodeURIComponent(fullPath)}`,
+        `/api/v1/storage/objects?path=${encodeURIComponent(path)}`,
         { responseType: 'json' }
       )
       .then(res => res.data);
--- a/src/api-clients/storage-api-client.js
+++ b/src/api-clients/storage-api-client.js
@@ -16,6 +16,15 @@ class StorageApiClient {
       .then(res => res.data);
   }
 
+  getCdnObjects(parentPath) {
+    return this.httpClient
+      .get(
+        `/api/v1/storage/cdn-objects?parentPath=${encodeURIComponent(parentPath)}`,
+        { responseType: 'json' }
+      )
+      .then(res => res.data);
+  }
+
   uploadFiles(files, prefix, { onProgress = () => {} } = {}) {
     const formData = new FormData();
 
--- a/src/api-clients/storage-api-client.js
+++ b/src/api-clients/storage-api-client.js
@@ -7,28 +7,19 @@ class StorageApiClient {
     this.httpClient = httpClient;
   }
 
-  getObjects(prefix) {
-    return this.httpClient
-      .get(
-        `/api/v1/storage/objects?prefix=${encodeURIComponent(prefix)}`,
-        { responseType: 'json' }
-      )
-      .then(res => res.data);
-  }
-
   getCdnObjects(parentPath) {
     return this.httpClient
       .get(
-        `/api/v1/storage/cdn-objects?parentPath=${encodeURIComponent(parentPath)}`,
+        `/api/v1/storage/objects?parentPath=${encodeURIComponent(parentPath)}`,
         { responseType: 'json' }
       )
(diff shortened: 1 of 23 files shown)


src/services/storage-service.js
```diff
@@ -7,25 +6,28 @@ import UserStore from '../stores/user-store.js';
...
+import { getPrivateStoragePathForRoomId, getPrefixFromStoragePath, getStoragePathType, STORAGE_PATH_TYPE, getPathSegments } from '../ui/path-helper.js';
 
 const { BadRequest } = httpErrors;
 
 export default class StorageService {
-  static get inject() { return [Cdn, RoomStore, RoomInvitationStore, LessonStore, StoragePlanStore, UserStore, LockStore, TransactionRunner]; }
+  static get inject() { return [ServerConfig, Cdn, RoomStore, RoomInvitationStore, LessonStore, StoragePlanStore, UserStore, LockStore, TransactionRunner]; }
 
-  constructor(cdn, roomStore, roomInvitationStore, lessonStore, storagePlanStore, userStore, lockStore, transactionRunner) {
+  // eslint-disable-next-line max-params
```
ahelmberger, line 23: I think the mapping should be done here and not in the controller, but then we need the server config as an additional argument) :-(

src/api-clients/storage-api-client.js
```diff
@@ -7,19 +7,19 @@ class StorageApiClient {
     this.httpClient = httpClient;
   }
 
-  getObjects(prefix) {
+  getCdnObjects(parentPath) {
```
adrianaluchian, line 10: Working with "path" instead of "prefix" makes it so much more understandable now I feel.

src/components/files-grid-viewer.js
```diff
@@ -1,69 +1,59 @@
...
     }
-    const overlayClasses = classNames('FilesGridViewer-fileOverlay', { 'is-visible': file.name === selectedFile?.name });
-    const actionsClasses = classNames('FilesGridViewer-actions', { 'are-visible': file.name === selectedFile?.name });
+    const overlayClasses = classNames('FilesGridViewer-fileOverlay', { 'is-visible': file.portableUrl === selectedFileUrl });
+    const actionsClasses = classNames('FilesGridViewer-actions', { 'are-visible': file.portableUrl === selectedFileUrl });
 
     return (
-      <div className="FilesGridViewer-fileContainer" key={file.path}>
-        <a className="FilesGridViewer-file" onClick={() => handleFileClick(file)}>
+      <div className="FilesGridViewer-fileContainer" key={file.portableUrl}>
```
adrianaluchian, line 48: Nice 👍  With this `portableUrl` we're one step closer to being able to remove the `sourceType` toggle, since we'll be able to deduce the source type wherever needed.

src/services/storage-service.js
adrianaluchian: I think that's fine, unless we make a dedicated method in the `client-data-mapping-service`. That way would be more aligned with all other client data we map.

src/components/files-viewer.less
```diff
@@ -0,0 +1,17 @@
...
+.FilesViewer-loadingOverlay {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  display: grid;
+  align-content: center;
+  justify-content: center;
+  background-color: fade(@edu-white, 50%);
```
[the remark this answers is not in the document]
ahelmberger: True, I adjusted the color and the opacity level to match the other overlay

src/api-clients/storage-api-client.js
adrianaluchian: Ok, I see! Then I agree that this is a better approach, you're basically pushing down the CDN model as much as possible 👍

ahelmberger:
> Nice to finally see the tests page functionality with a bit of navigation and actual real files heart_eyes
> 
> One tiny thing I noticed, maybe it's easy to fix? The "Type" for folders is now "Unbekannt", but only on the tests page, it's fine in the legacy `storge-browser`:
> 
> Screen Shot 2022-05-24 at 17 13 53
> 
> Screen Shot 2022-05-24 at 17 16 01

Thanks for noticing. In the legacy browser it comes from the mime type helper which we do not use here anymore, I realized. But it was easy to fix.

adrianaluchian (approved):

PR #426, opened 2022-05-24
EDU-579 Create custom media player progress bar

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-579

Screen Shot 2022-05-24 at 16 05 36

Audio only:

Screen Shot 2022-05-24 at 16 06 13


adrianaluchian:
@ahelmberger, included timecode tooltip on hover as you suggested. It is indeed very good to have!

ahelmberger:
> @ahelmberger, included timecode tooltip on hover as you suggested. It is indeed very good to have!

Perfect, it looks so much better, and we get nice tooltips for free on top :heart:

ahelmberger (approved):

PR #432, opened 2022-06-08
EDU-597 Add parent display name

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-597

### DIFF
--- a/src/services/storage-service.js
+++ b/src/services/storage-service.js
@@ -134,27 +134,65 @@ export default class StorageService {
   }
 
   async getObjects({ parentPath, recursive }) {
-    const prefix = parentPath ? `${parentPath}/` : '';
-    const objects = await this.cdn.listObjects({ prefix, recursive });
+    const currentDirectorySegments = parentPath.split('/').filter(seg => !!seg);
+    const encodedCurrentDirectorySegments = currentDirectorySegments.map(s => encodeURIComponent(s));
 
-    const mappedObjects = objects.map(obj => {
+    let currentDirectory;
+    if (currentDirectorySegments.length > 0) {
+      currentDirectory = {
+        displayName: currentDirectorySegments[currentDirectorySegments.length - 1],
+        parentPath: currentDirectorySegments.slice(0, -1).join('/'),
+        path: currentDirectorySegments.join('/'),
+        url: [this.serverConfig.cdnRootUrl, ...encodedCurrentDirectorySegments].join('/'),
+        portableUrl: `cdn://${encodedCurrentDirectorySegments.join('/')}`,
+        createdOn: null,
+        type: CDN_OBJECT_TYPE.directory,
+        size: null
+      };
+    } else {
+      currentDirectory = null;
+    }
+
+    let parentDirectory;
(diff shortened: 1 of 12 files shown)


src/services/storage-service.js
```diff
@@ -134,27 +134,65 @@ export default class StorageService {
...
+        portableUrl: `cdn://${encodedObjectSegments.join('/')}`,
         createdOn: isDirectory ? null : obj.lastModified,
         type: isDirectory ? CDN_OBJECT_TYPE.directory : CDN_OBJECT_TYPE.file,
         size: isDirectory ? null : obj.size
       };
     });
 
-    return mappedObjects;
+    return { parentDirectory, currentDirectory, objects };
   }
```
adrianaluchian, line 196: This method is decomposing and recomposing paths alot, feels like a test would be helpful here 🤔

adrianaluchian:
The upper level navigation actually looks nice in the list view, we'll just need a better icon so it's good for the grid view as well

ahelmberger:
> The upper level navigation actually looks nice in the list view, we'll just need a better icon so it's good for the grid view as well.
> 
> How would this look like to you? Screen Shot 2022-06-08 at 13 50 44

Agree, that looks a lot nicer. Did you create this as an SVG or just manipulate the pixels (i.e. can you provide me yours or should a re-create it :smiling_imp:)?

src/services/storage-service.js
ahelmberger: True, it's the one function in this service that has no tests, I will add some.

adrianaluchian:
> > The upper level navigation actually looks nice in the list view, we'll just need a better icon so it's good for the grid view as well.
> > How would this look like to you? Screen Shot 2022-06-08 at 13 50 44
> 
> Agree, that looks a lot nicer. Did you create this as an SVG or just manipulate the pixels (i.e. can you provide me yours or should a re-create it 😈)?

Sure, as you're ok with it I've committed it to the assets repo, along with a version for the files list view (haven't checked how that one looks but feel free to use, adjust or delete it).

ahelmberger:
> > [...]
> > Agree, that looks a lot nicer. Did you create this as an SVG or just manipulate the pixels (i.e. can you provide me yours or should a re-create it smiling_imp)?
> 
> Sure, as you're ok with it I've committed it to the assets repo, along with a version for the files list view (haven't checked how that one looks but feel free to use, adjust or delete it).

Thanks so much :pray:

src/services/storage-service.js
ahelmberger: Added, thanks for the suggestion, I actually found some edge case problems.

ahelmberger:
> > [...]
> > Agree, that looks a lot nicer. Did you create this as an SVG or just manipulate the pixels (i.e. can you provide me yours or should a re-create it smiling_imp)?
> 
> Sure, as you're ok with it I've committed it to the assets repo, along with a version for the files list view (haven't checked how that one looks but feel free to use, adjust or delete it).

Added and adjusted the new icons :+1:

PR #441, opened 2022-06-13
EDU-594 Support file deletion

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-594

### DIFF
--- a/src/components/storage-location.js
+++ b/src/components/storage-location.js
@@ -1,9 +1,11 @@
 import PropTypes from 'prop-types';
 import UsedStorage from './used-storage.js';
 import { useTranslation } from 'react-i18next';
-import React, { useEffect, useState } from 'react';
+import cloneDeep from '../utils/clone-deep.js';
 import { Alert, Button, message, Select } from 'antd';
+import { useSetStorageLocation } from './storage-context.js';
 import { useSessionAwareApiClient } from '../ui/api-helper.js';
+import React, { useCallback, useEffect, useState } from 'react';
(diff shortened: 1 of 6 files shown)


src/components/storage-location.js
```diff
@@ -12,15 +14,35 @@ import { getParentPathForStorageLocationPath, getStorageLocationPathForUrl } fro
...
+  const [filesViewerDisplay, setFilesViewerDisplay] = useState(FILE_VIEWER_DISPLAY.grid);
+
+  const fetchStorageContent = useCallback(async () => {
+    if (!currentLoadedDirectoryPath) {
+      return;
+    }
+
+    try {
+      let result = await storageApiClient.getCdnObjects(currentLoadedDirectoryPath);
+      if (!result.objects.length) {
```
ahelmberger, line 35: I think I don't completely understand the purpose of this fallback. It seems it makes opening an empty folder impossible? Or did I get it wrong?

adrianaluchian: We actually never have empty "folders", this doesn't exist on the CDN.
This check is for the case when one types in a wrong path, so we fall back onto the initial storage location path.
For example:
`cdn://media/documentThatDoesNotExist"`

ahelmberger: As discussed, we have to find a way how to show an actually valid empty folder (e.g. immediately after creating a new document). For now let's leave it this way until we have figured out a way to implement this.

ahelmberger (approved):

PR #443, opened 2022-06-14
EDU-592 Set up select button

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-592

### DIFF
--- a/src/components/pages/tests.js
+++ b/src/components/pages/tests.js
@@ -12,6 +12,11 @@ function Tests({ PageTemplate }) {
   const [initialUrl, setInitialUrl] = useState('');
   const [isResourceSelectorModalVisible, setIsResourceSelectorModalVisible] = useState(false);
 
+  const handleResourceSelectorSelect = filePath => {
+    console.log('selected file path: ', filePath);
+    setIsResourceSelectorModalVisible(false);
+  };
+
   return (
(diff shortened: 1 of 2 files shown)


src/components/resource-selector.js
```diff
@@ -42,7 +42,7 @@ function ResourceSelector({ allowedLocationTypes, initialUrl, onCancel, onSelect
             initialUrl={initialUrl}
             onEnterFullscreen={() => setIsFullscreen(true)}
             onExitFullscreen={() => setIsFullscreen(false)}
-            onSelect={onSelect}
+            onSelect={filePath => onSelect(filePath)}
```
ahelmberger, line 45: I don't understand the change here, it's equivalent :thinking:

adrianaluchian: I know, just wanted to make the argument transparent, cause I was not sure if this will be the value we stick with.
Should I remove it?

ahelmberger: Hm, we usually just propagate regardless of the arguments. I personally would remove it, but I'm not at all passionate about it. Up to you, I was just curious.

adrianaluchian: Ok, removed! It's true that if we'd change the method signature we'd have to change also here (which always leaves room for bugs).

PR #445, opened 2022-06-14
EDU-599-Markdown-inputs

ahelmberger:


src/components/markdown-help.less
```diff
@@ -0,0 +1,92 @@
...
+
+  &.MarkdownHelp--small {
+    font-size: 12px;
+  }
+
+  &.MarkdownHelp--block {
+    padding: 4px 12px;
+    border-radius: 5px;
+    background-color: @edu-secondary-color-light;
+    opacity: 0.25;
```
adrianaluchian, line 20: I guess this opacity could be extracted in a semantically meaningful variable, something like `@edu-markdown-[inactive/unfocused]-opacity` 🤔

ahelmberger: Good idea, as it is used in two places. Done!

src/components/markdown-input.less
```diff
@@ -0,0 +1,74 @@
+@markdown-input-block-min-height: 100px;
+
+.MarkdownInput {
+  display: grid;
+  gap: 5px;
+  grid-template-columns: 1fr;
```
adrianaluchian, line 6: We'd need to let this guy take `width: 100%` of its parent, for cases like this:

Screen Shot 2022-06-14 at 14 09 00
👇 
Screen Shot 2022-06-14 at 14 09 40

ahelmberger: True, thanks, now it takes 100%

adrianaluchian (approved):

PR #456, opened 2022-06-23
EDU-623 Reorganize preview and delete buttons

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-623


src/styles/global-variables.less
```diff
@@ -23,6 +23,7 @@
 
 // Layout
 @edu-page-content-padding: 30px;
+@edu-controls-group-gap: 10px;
```
ahelmberger, line 26: Nice

PR #462, opened 2022-06-24
EDU-626 Replace action links on docs page with icons

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-626

### DIFF
--- a/src/components/pages/docs.less
+++ b/src/components/pages/docs.less
@@ -17,6 +17,49 @@
   width: 100%;
 }
 
+.DocsPage-actions {
+  display: grid;
+  grid-auto-flow: column;
+  grid-auto-columns: max-content;
+  gap: 5px;
+}
+
+.DocsPage-action {
+  .m-box-shadow;
+  background-color: @edu-white;
+  width: @edu-control-height-sm;
+  height: @edu-control-height-sm;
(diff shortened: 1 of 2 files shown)


src/components/pages/docs.less
```diff
@@ -17,6 +17,49 @@
...
+  .m-box-shadow;
+  background-color: @edu-white;
+  width: @edu-control-height-sm;
+  height: @edu-control-height-sm;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  text-decoration: none;
+  opacity: 90%;
+  cursor: pointer;
```
adrianaluchian, line 37: Tiny thing: this is redundant.

adrianaluchian:
This small change makes such a big difference in the UI 😳 The page no longer looks like an "admin-only" page.

ahelmberger:
> This small change makes such a big difference in the UI flushed The page no longer looks like an "admin-only" page.

I thought the same, the page looks a lot less bureaucratic.

src/components/pages/docs.less
ahelmberger: True, removed. Thanks!

adrianaluchian (approved):

PR #464, opened 2022-06-24
EDU-624 Replace LinkPopover with standard Menu

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-624


src/plugins/table/table-designer-menu.js
```diff
@@ -35,78 +35,98 @@ export const menuItemInfos = {
...
     default:
       items = [];
       break;
   }
 
+  const menu = <Menu items={items} onClick={handleMenuItemClick} />;
+
   return (
-    <LinkPopover items={items} placement={placement} trigger="click" renderSeparator="onlyWhenSpecified" onVisibleChange={setIsPopoverVisible}>
+    <Dropdown overlay={menu} trigger={['click']} onVisibleChange={handleDropdownVisibleChange} arrow>
```
adrianaluchian, line 125: Seems you could set `arrow={{ pointAtCenter: true }}` and the arrow of the dropdown would be right on the dot in the table menus:
Screen Shot 2022-06-24 at 16 12 56
👇 
Screen Shot 2022-06-24 at 16 09 36
👌

ahelmberger: Wow, done, thanks. I didn't know about this setting. I wonder why anyone would ever want a different default behavior :thinking:

adrianaluchian: Possibly for some cases when the layout cannot afford the dropdown going out of that particular alignment bound of the parent or when it needs to be aligned with the parent 🤔 🤷‍♀️

PR #468, opened 2022-06-27
EDU-603 Fix rendering issues in table designer

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-603

### DIFF
--- a/src/plugins/table/table-designer.js
+++ b/src/plugins/table/table-designer.js
@@ -47,7 +47,7 @@ const getDesignerCellStyle = designerCell => {
   };
 };
 
-const getDeignerCellKey = designerCell => {
+const getDesignerCellKey = designerCell => {
   switch (designerCell.designerCellType) {
     case DESIGNER_CELL_TYPE.content:
       return [designerCell.rowIndex, designerCell.columnIndex, designerCell.rowSpan, designerCell.columnSpan, designerCell.cellType].join('|');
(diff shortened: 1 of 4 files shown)


src/plugins/table/table-designer.js
```diff
@@ -162,30 +195,31 @@ function TableDesigner({ content, onContentChange }) {
...
+      'TableDesigner-gridCell--firstInColumn': designerCell.isFirstInColumn,
+      'TableDesigner-gridCell--cellTypeHeader': designerCell.cellType === CELL_TYPE.header,
+      'TableDesigner-gridCell--cellTypeBody': designerCell.cellType === CELL_TYPE.body
+    });
 
-    const children = (
-      <Fragment>
+    return (
+      <div
+        key={getDeignerCellKey(designerCell)}
```
adrianaluchian, line 218: Small typo: `[..]Deigner[..]`

src/plugins/table/table.less
```diff
@@ -12,66 +15,76 @@
...
   }
 
-  &.TableDesigner-tableCell--columnHeader {
-    .TableDisplay-cell;
+  &.TableDesigner-gridCell--columnHeader {
+    padding: @table-plugin-cell-padding;
     text-align: center;
-    vertical-align: middle;
-    border: none;
     height: 20px;
```
adrianaluchian, line 76: Room for some small simplification here with merging the row and column CSS class variations into one, since they don't diverge.

adrianaluchian: Ah ok, I see 😬

src/plugins/table/table-designer.js
ahelmberger: Thanks, fixed.

PR #472, opened 2022-06-28
EDU-628 Support images in ear training plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-628


src/plugins/image/utils.js
```diff
@@ -1,16 +1,3 @@
-import { SOURCE_TYPE } from './constants.js';
```
ahelmberger, line 1: I deleted the second part of this module (in the already merged PR), so this file can be deleted now.

src/components/item-panel.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
```
ahelmberger, line 1: Nice unification :heart_eyes:

ahelmberger (approved):

src/components/item-panel.js
adrianaluchian: Danke! ☺️

PR #475, opened 2022-06-28
EDU-631 Remove color change effect on grid items hover

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-631

### DIFF
--- a/src/components/ui-language-dialog.less
+++ b/src/components/ui-language-dialog.less
@@ -11,6 +11,7 @@
 }
 
 .UiLanguageDialog-language {
+  cursor: pointer;
   padding: 10px 10px 5px 10px;
   border-radius: @edu-border-radius-base;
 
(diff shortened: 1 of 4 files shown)


src/components/ui-language-dialog.less
```diff
@@ -4,38 +4,21 @@
...
-      box-shadow: none;
-    }
-
-    &:hover {
-      .m-box-shadow;
-    }
-  }
+.UiLanguageDialog-language {
+  padding: 10px 10px 5px 10px;
+  border-radius: @edu-border-radius-base;
```
ahelmberger, line 15: We might want to add `cursor: pointer;` here, as it is gone now that this is not a radio button anymore.

adrianaluchian: True, added!

PR #487, opened 2022-07-01
EDU-656 Extend AMB response

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-656


adrianaluchian:
@ahelmberger, thanks for the remark that this `about` field should be configurable 👍 Changed.

PR #488, opened 2022-07-01
EDU-636 Make favorites deletable

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-636

### DIFF
--- a/src/components/favorites-tab.less
+++ b/src/components/favorites-tab.less
@@ -5,3 +5,36 @@
   justify-content: flex-end;
   margin-bottom: 20px;
 }
+
+.FavoritesTab-actions {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 5px;
+}
+
+.FavoritesTab-action {
+  .m-box-shadow;
+  background-color: @edu-white;
+  text-decoration: none;
+  display: flex;
+  width: 22px;
+  height: 22px;
+  align-items: center;
+  justify-content: center;
+
+  &:hover,
+  &:active {
+    .m-box-shadow-dark;
+    text-decoration: none;
+  }
+
+  &.FavoritesTab-action--delete {
+    color: @edu-error-color;
+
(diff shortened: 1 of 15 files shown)


src/services/dashboard-service.js
```diff
@@ -75,27 +84,42 @@ class DashboardService {
...
             type: USER_ACTIVITY_TYPE.documentMarkedFavorite,
             timestamp: favorite.setOn,
-            getData: async () => {
+            data: null,
+            isDeprecated: null,
+            [completionFunction]: async () => {
               const document = await this.documentStore.getDocumentMetadataByKey(favorite.id);
-              return { _id: favorite.id, title: document.title };
+              return {
+                data: { _id: favorite.id, title: document.title },
```
adrianaluchian, line 92: We get a `cannot read properties of undefined` exception here if `document` is `null`, and this can occur if the favorite document has source `external` because external documents can be hard deleted by admins.

src/components/favorites-tab.less
```diff
@@ -5,3 +5,36 @@
...
+
+.FavoritesTab-actions {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 5px;
+}
+
+.FavoritesTab-action {
+  .m-box-shadow;
```
adrianaluchian, line 17: What would you think about removing the shadow in such cases where the background is white and only adding the `m.-box-shadow` on hover?
This default shadow makes sense when the background is not white I feel, but it seems a bet strange on white.
If you feel the same, then maybe we can change it on the `documents` page too 🤔

adrianaluchian: Sure, good idea, we could extract them 👍

adrianaluchian (approved):

src/components/favorites-tab.less
ahelmberger: I'll make a ticket then :+1:

PR #495, opened 2022-07-06
EDU-637 ABC notation preview

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-637

**abc-notation plugin**

abc-notation

**ear-training plugin**

ear-training

### DIFF
--- a/src/styles/elements.less
+++ b/src/styles/elements.less
@@ -26,7 +26,3 @@ h5 {
 h6 {
   font-size: @edu-heading-6-size;
 }
-
-textarea {
-  resize: none;
-}
--- a/src/components/never-scrolling-text-area.less
+++ b/src/components/never-scrolling-text-area.less
@@ -6,6 +6,7 @@
   grid-auto-flow: row;
 
   textarea {
+    resize: none !important;
     border: none !important;
     outline: none !important;
     box-shadow: none !important;


src/components/input-and-preview.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function InputAndPreview({ input, preview }) {
```
adrianaluchian, line 4: Interesting simple wrapper component idea, keeps things cleaner than with mixins or CSS components 👍

src/styles/elements.less
```diff
@@ -26,3 +26,7 @@ h5 {
 h6 {
   font-size: @edu-heading-6-size;
 }
+
+textarea {
+  resize: none;
```
adrianaluchian, line 31: This feels a bit misplaced to me here. I would move it into the component that reflects this property setup, in
```
.NeverScrollingTextArea {
   ...
  textarea {
      ...
  }
}
```

adrianaluchian:
Super nice solution!! 👏

src/styles/elements.less
ahelmberger: We actually never want to see the ugly browser bars that appear for a fraction of a second when the server-rendered HTML is loaded. So I thought I'd put it into the global elements stylesheet like the `h1`-`h6` styles which fixed the problem. However, since this PR `textarea` is used in no other place than in `NeverScrollingTextArea`, so I moved it according to your suggestion

src/components/input-and-preview.js
ahelmberger: I also updated `MardownInput` to use this component now.

ahelmberger:
> Super nice solution!! clap

Thanks. Maybe you can check the last three commits, I basically:

* also used the `input-and-preview` container in Markdown input
* Moved the textarea styling (see above)
* Made the preview have a light silver background to indicate to the user that they cannot type there (as we talked about today)

adrianaluchian:
> > Super nice solution!! clap
> 
> Thanks. Maybe you can check the last three commits, I basically:
> 
> [...]

I have checked it out and made an additional small suggestion.

src/components/input-and-preview.less
```diff
@@ -14,6 +14,7 @@
 
 .InputAndPreview--preview {
   .m-control-border-base;
+  background-color: @edu-background-color-light;
```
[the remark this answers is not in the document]
adrianaluchian: Never mind, cause we have this ticket for extra changes: https://educandu.atlassian.net/browse/EDU-662

PR #497, opened 2022-07-08
EDU-640 content alignment in table plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-640

### DIFF
--- a/src/plugins/table/table-utils.js
+++ b/src/plugins/table/table-utils.js
@@ -161,7 +161,7 @@ function changeCellProps(tableModel, rowIndex, columnIndex, changedCellProps) {
   return {
     ...tableModel,
     cells: tableModel.cells.map(cell => {
-      return cell.rowIndex === rowIndex && cell.columnIndex === columnIndex
+      return isCellHit(cell, rowIndex, columnIndex)
         ? { ...cell, ...changedCellProps }
         : cell;
     })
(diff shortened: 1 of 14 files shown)


src/plugins/table/table-utils.js
```diff
@@ -36,9 +40,22 @@ export const DESIGNER_CELL_ACTION = {
...
+  disconnectCell: 'disconnect-cell',
+  setVerticalAlignmentToTop: 'set-vertical-alignment-to-top',
+  setVerticalAlignmentToMiddle: 'set-vertical-alignment-to-middle',
+  setVerticalAlignmentToBottom: 'set-vertical-alignment-to-bottom',
+  setHorizontalAlignmentToLeft: 'set-horizontal-alignment-to-left',
+  setHorizontalAlignmentToCenter: 'set-horizontal-alignment-to-center',
+  setHorizontalAlignmentToRight: 'set-horizontal-alignment-to-right'
 };
 
+export function isCellHit(cell, rowIndex, columnIndex) {
```
adrianaluchian, line 53: Small thing, how about `isCellAtCoordinates`? Took me a bit to get the `hit` part and then O directly thought of the battleships game 😂

adrianaluchian (approved):
Works super nice 😍 Can't imagine there is anything missing now from the table features, there's full control over everything.

src/plugins/table/table-utils.js
ahelmberger: Myeah, it comes from "hit testing" in graphics development when checking if an object is "hit" by a location. I wanted to avoid anything with "coordinates" in it, because the cell is also "hit" when e.g. `rowIndex` is `-1` (meaning: all cells that touch the column at `columnIndex`, no matter which row). So, as `rowIndex` and `columnIndex` do not necessarily constitute a real cell coordinate, I was searching for more abstract things like `isCellConcerned`, but as "concerned" sounded too human to me, I went with "hit". Would `isCellAffected` sound better to you as an alternative?

adrianaluchian: Ok I see, hadn't realized they are not quite coordinates 🤔 
`isCellAffected` sounds good though.

ahelmberger: Ok, I actually like it better, too (haven't thought of "affected" initially). Changed!

PR #502, opened 2022-07-12
EDU-663 update and restyle abc notation

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-663


src/components/media-player-controls.js
```diff
@@ -82,6 +78,12 @@ function MediaPlayerControls({
     return <Menu items={items} onClick={handleSettingsMenuItemClick} />;
   };
 
+  const renderTimeDisplay = () => {
+    return durationInMilliseconds
+      ? <Fragment>{formatMillisecondsAsDuration(playedMilliseconds)}&nbsp;/&nbsp;{formatMillisecondsAsDuration(durationInMilliseconds)}</Fragment>
+      : <Fragment>--:--&nbsp;/&nbsp;--:--</Fragment>;
```
adrianaluchian, line 84: Nice placeholder idea.

adrianaluchian:
Seems now we get an exception when trying to download anything from internal storage uploaded in video, audio, etc. plugins 🤔

ahelmberger:
> Seems now we get an exception when trying to download anything from internal storage uploaded in video, audio, etc. plugins thinking

Yeah, this is also happening on the main branch (but only on `localhost`). It's a CORS problem. I know the axios request is there as a workaround in order to always download instead of opening a new browser tab. But we would save ourself a some trouble if we'd just let the browser handle the download naturally (by just setting the `href` to the actual URL). Maybe we can discuss in the near future.

src/components/media-player-controls.js
ahelmberger: Thanks, I thought we should indicate somehow that the file is not "loaded" yet (and I found "0:00" total length confusing) :-)

adrianaluchian:
> > Seems now we get an exception when trying to download anything from internal storage uploaded in video, audio, etc. plugins thinking
> 
> Yeah, this is also happening on the main branch (but only on `localhost`). It's a CORS problem. I know the axios request is there as a workaround in order to always download instead of opening a new browser tab. But we would save ourself a some trouble if we'd just let the browser handle the download naturally (by just setting the `href` to the actual URL). Maybe we can discuss in the near future.

Oh yes, of course, it's CORS. Ok, we can live with that and we can always use the tunnel to test such cases.

adrianaluchian (approved):
Looks good and unified now 👏

PR #504, opened 2022-07-12
EDU-657 Extract and unify action buttons

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-657

### DIFF
--- a/src/components/action-button.js
+++ b/src/components/action-button.js
@@ -1,0 +1,62 @@
+import React from 'react';
+import { Tooltip } from 'antd';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export const ACTION_BUTTON_INTENT = {
+  default: 'default',
+  info: 'info',
+  success: 'success',
+  warning: 'warning',
+  error: 'error'
+};
+
+export function ActionButtonGroup({ children }) {
+  return <span className="ActionButtonGroup">{children}</span>;
+}
+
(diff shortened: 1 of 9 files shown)


src/components/action-button.js
```diff
@@ -0,0 +1,62 @@
...
+
+export const ACTION_BUTTON_INTENT = {
+  default: 'default',
+  info: 'info',
+  success: 'success',
+  warning: 'warning',
+  error: 'error'
+};
+
+export function ActionButtonGroup({ children }) {
```
ahelmberger, line 14: Not sure what you think about this. Normally we do not export multiple components per file. But as this is just a simple container I thought we could leave them together with the "items" component.

src/components/favorites-tab.less
```diff
@@ -7,35 +7,7 @@
...
-  }
-
-  &.FavoritesTab-action--delete {
-    color: @edu-error-color;
-
-    &:hover,
-    &:active {
-      color: @edu-error-color;
-    }
-  }
```
adrianaluchian, line 38: Hadn't noticed there was this amount of duplication between these action buttons. Nice that you reduced it 👍

adrianaluchian (approved):

src/components/action-button.js
adrianaluchian: I also find it fine in this case.

PR #508, opened 2022-07-18
EDU-675 Remove default tags

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-675


src/components/document-metadata-modal.yml
```diff
@@ -10,6 +10,9 @@ titleRequired:
 descriptionTooLong:
   en: The description cannot have more than {maxChars, plural, =1 {# character} other {# characters}}
   de: Die Beschreibung darf nicht länger als {maxChars} Zeichen sein
+tagsPlaceholder:
+  en: Enter keywords
+  de: Stickwörter angeben
```
ahelmberger, line 15: That would be "embroidery words" :rofl: We need "Stic**h**wörter"

ahelmberger (approved):

src/components/document-metadata-modal.yml
adrianaluchian: 😂 oh too bad! 👵

PR #510, opened 2022-07-18
EDU-681 Migrate document model to merged model

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-681

### DIFF
--- a/migrations/educandu-2022-07-18-02-migrate-document-models.js
+++ b/migrations/educandu-2022-07-18-02-migrate-document-models.js
@@ -1,0 +1,27 @@
+/* eslint-disable camelcase */
+
+export default class Educandu_2022_07_18_02_migrate_documents_models {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    const props = {
+      roomId: '',
+      dueOn: null,
+      accessLevel: 'public'
+    };
+    await this.db.collection('documentRevisions').updateMany({}, { $set: { ...props } });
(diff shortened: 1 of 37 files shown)


adrianaluchian:
Sorry, forgot the DB indices, will add!

migrations/educandu-2022-07-18-02-migrate-document-models.js
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable camelcase */
+
+export default class Educandu_2022_07_18_02_migrate_documents_models {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    const props = {
+      roomId: '',
```
ahelmberger, line 10: I would opt for `null` if intentionally if no room is assigned. I know that the `restoredFrom` field also is an empty string, but I'd consider that an "accident". In the sections e.g. `deletedBy` (which also is not an arbitrary string, but an ID reference) is `null` if not applicable (so are all date types). I would interpret empty string as "text intentionally left empty by the user" opposed to `null` (data value absent as part of the domain model). Of course, that's disputable. What do you think?

ahelmberger (approved):

migrations/educandu-2022-07-18-02-migrate-document-models.js
adrianaluchian: I think we should keep `null` (as we do for all other user data), as a lack of data, not as an intentionally set empty data by the user.

PR #511, opened 2022-07-19
EDU-651 expand interactive media chapter interaction

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-651


adrianaluchian:
Little UI glitch here:

Screen Shot 2022-07-19 at 18 46 13

migrations/educandu-2022-07-19-01-rename-question-to-text-in-interactive-media-plugin copy.js
```diff
@@ -0,0 +1,72 @@
...
+      if (docWasUpdated) {
+        updateCount += 1;
+        await this.db.collection(collectionName).replaceOne({ _id: doc._id }, doc);
+      }
+    }
+
+    return updateCount;
+  }
+
+  async collectionDown(collectionName) {
```
adrianaluchian, line 32: Just as a note, a nice little trick for these mirror up/down methods would be to extract a method like `migrate(collectionName, oldProp, newProp)`, and with this we really ensure we didn't forget to update both methods when we find an issue in the `collectionUp` (which happened to me). We did that in `educandu-2022-07-11-01-rename-text-to-copyrightNotice-in-all-plugins`.

src/plugins/interactive-media/interactive-media.yml
```diff
@@ -37,6 +37,9 @@ answers:
 addAnswer:
   en: Add answer
   de: Antwort hinzufügen
+addAnswerInfo:
+  en: "In order to allow the user to examine their knowledge, you can put the text above into a question and add possible answers here:"
```
adrianaluchian, line 41: Maybe "review" would work better than "examine" 🔍

src/components/media-player.js
```diff
@@ -99,6 +141,8 @@ function MediaPlayer({
```
adrianaluchian, line 141: Seems `trackRef.current` is not initialized on the initial setting of the video source and it throws on line 138:

Screen Shot 2022-07-20 at 09 44 36

I can reproduce this only when setting the video for the first time, on the newly created `interactive-media` section.

ahelmberger:
> Little UI glitch here:
> 
> Screen Shot 2022-07-19 at 18 46 13

Fixed it, they're 100% now

migrations/educandu-2022-07-19-01-rename-question-to-text-in-interactive-media-plugin copy.js
ahelmberger: I see. I did a similar thing. Maybe we should re-introduce the kind of migration helpers we once had (to process all documents in particular).

src/plugins/interactive-media/interactive-media.yml
ahelmberger: True, sounds better. Fixed.

src/components/media-player.js
ahelmberger: Very good catch, fixed it. Thanks!

ahelmberger:
> Looks very good and finally all plugins start having a common style heart_eyes Screen Shot 2022-07-20 at 09 39 02
> 
> And given all the many moving parts and possible scenarios the navigation, playback etc. works nicely +1

Thanks :bow: I also enjoy the UI getting more and more streamlined

adrianaluchian (approved):

PR #512, opened 2022-07-19
EDU-678 Add search to storage location view

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-678


src/utils/array-utils.js
```diff
@@ -106,3 +106,17 @@ export function range({ from, to, step }) {
 
   return [...Array(Math.floor((to - from) / actualStep) + 1)].map((_, i) => from + (i * actualStep));
 }
+
+export function unique(items, keyFunc = item => item) {
```
adrianaluchian, line 110: Tiny tiny thing, `ensureIsUnique` would match better the other helpers 😋

ahelmberger: True, good idea. Renamed!

adrianaluchian:
One more case that I stumbled upon:

- when you search and then delete one of the files that were found, the results list does not get updated

ahelmberger:
> One more case that I stumbled upon:
> 
> * when you search and then delete one of the files that were found, the results list does not get updated

Good catch, fixed, thanks!

ahelmberger:
> Using now this features makes me realize how good it is to have it, the resources get alot more accessible now!
> 
> Found a couple of glitches:
> 
> * uploading to private storage is disabled for me for most rooms (pre-existing or newly created, public or private alike) and it seems the CDN folder fails to create even; so it might be an already existing issue but you might know better at this point
> 
> Screen Shot 2022-07-20 at 10 33 01
> 
> * when you clear the typed in search text without applying it, the hint on the bottom is updated; we should probably keep the the last applied search text in the state to display it there, for consistency with the result the user is seeing
> 
> Screen Shot 2022-07-20 at 10 14 21
> 
> * maybe we could make the "Zurück zur Verzeichnisansicht" link styled the same as the "<< Zurück" link on the preview and upload (which is link styled, while this is button styled). or on top we could reuse your new text for the other too as well, would be better I think.

I took care of the second and third point

> Using now this features makes me realize how good it is to have it, the resources get alot more accessible now!
> 
> Found a couple of glitches:
> 
> * uploading to private storage is disabled for me for most rooms (pre-existing or newly created, public or private alike) and it seems the CDN folder fails to create even; so it might be an already existing issue but you might know better at this point
> 
> Screen Shot 2022-07-20 at 10 33 01
> 
> * when you clear the typed in search text without applying it, the hint on the bottom is updated; we should probably keep the the last applied search text in the state to display it there, for consistency with the result the user is seeing
> 
> Screen Shot 2022-07-20 at 10 14 21
> 
> * maybe we could make the "Zurück zur Verzeichnisansicht" link styled the same as the "<< Zurück" link on the preview and upload (which is link styled, while this is button styled). or on top we could reuse your new text for the other too as well, would be better I think.

I took care of point 2 and 3 (we cannot use my "Zurück zur Verzeichnisansicht" gereically though, because from the preview screen you could go back to either directory view or search view, so I left the original text for the header links).

Point 1 I could not reproduce. Upload into private storage works for me for existing rooms as well as for newly created rooms. We have to investigate deeper here :thinking:

ahelmberger:
> > Using now this features makes me realize how good it is to have it, the resources get alot more accessible now!
> > Found a couple of glitches:
> > 
> > * uploading to private storage is disabled for me for most rooms (pre-existing or newly created, public or private alike) and it seems the CDN folder fails to create even; so it might be an already existing issue but you might know better at this point
> > 
> > Screen Shot 2022-07-20 at 10 33 01
> > 
> > * when you clear the typed in search text without applying it, the hint on the bottom is updated; we should probably keep the the last applied search text in the state to display it there, for consistency with the result the user is seeing
> > 
> > Screen Shot 2022-07-20 at 10 14 21
> > 
> > * maybe we could make the "Zurück zur Verzeichnisansicht" link styled the same as the "<< Zurück" link on the preview and upload (which is link styled, while this is button styled). or on top we could reuse your new text for the other too as well, would be better I think.
> 
> I took care of point 2 and 3 (we cannot use my "Zurück zur Verzeichnisansicht" generically though, because from the preview screen you could go back to either directory view or search view, so I left the original text for the header links).
> 
> Point 1 I could not reproduce. Upload into private storage works for me for existing rooms as well as for newly created rooms. We have to investigate deeper here thinking

PR #514, opened 2022-07-20
EDU-683 Add migration logic to convert lessons to documents

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-683

### DIFF
--- a/src/services/lesson-convertor.js
+++ b/src/services/lesson-convertor.js
@@ -1,0 +1,46 @@
+/* eslint-disable no-console */
+import UserService from './user-service.js';
+import Database from '../stores/database.js';
+import DocumentService from './document-service.js';
+
+class LessonConvertor {
+  static get inject() {
+    return [Database, UserService, DocumentService];
+  }
+
+  constructor(db, userService, documentService) {
+    this.db = db;
+    this.userService = userService;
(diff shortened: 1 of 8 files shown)


src/services/lesson-convertor.js
```diff
@@ -0,0 +1,46 @@
...
+
+    for (const lesson of notConvertedLessons) {
+      // eslint-disable-next-line no-await-in-loop
+      const user = await this.userService.getUserById(lesson.createdBy);
+
+      const data = {
+        roomId: lesson.roomId,
+        createdOn: lesson.createdOn,
+        createdBy: lesson.createdBy,
+        dueOn: lesson.schedule?.startsOn,
```
ahelmberger, line 30: To avoid `undefined`/`null` mixtures in the  we might wanna add `|| null` at the end here.

adrianaluchian: I didn't set any fallbacks here for any of the fields, as `documentService` ensures a consistent fallback for everything in `_buildDocumentRevision`

ahelmberger: Ah, that's so true :man_facepalming:

ahelmberger (approved):

PR #522, opened 2022-07-25
EDU-692 Change username to display name

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-692

Screen Shot 2022-07-25 at 10 38 05


adrianaluchian:
@ahelmberger , so that I don't forget, please have a close look at the German texts and help me correct them ☺️

src/components/account-tab.yml
```diff
@@ -14,8 +14,8 @@ passwordResetEmailSent:
...
+  en: All your public contributions (documents, public rooms, files uploaded to the public storage) will remain on the website and your name will still be displayed as author or contributor. All other data (personal information, private rooms and files uploaded to the private storage) will be removed. You will no longer be able to log into the website.
+  de: Alle Ihre öffentlichen Beiträge (Dokumente, öffentliche Räume und Dateien, die Sie in den öffentlichen Speicher hochgeladen haben), bleiben auf der Website. Ihr Anzeigenamen wird weiterhin als Autor oder Bearbeiter angezeigt. Alle anderen Daten (persönliche Informationen, private Räume und Dateien, die Sie in den privaten Speicher hochgeladen haben), werden gelöscht. Sie können sich dann nicht mehr auf der Website anmelden.
```
ahelmberger, line 18: Anzeigename

src/components/displayName-form-item.yml
```diff
@@ -0,0 +1,9 @@
+enterDisplayName:
+  en: Please enter a display name
+  de: Bitte geben Sie einen Anzeigenamen an
+displayNameIsTooShort:
+  en: The display name must be at least {length} characters long
+  de: Der Anzeigenamen muss mindestens {length} Zeichen lang sein
+displayNameIsTooLong:
+  en: The display name must be at most {length} characters long
+  de: Der Anzeigenamen muss höchstens {length} Zeichen lang sein
```
ahelmberger, line 9: Anzeigename

src/components/login-form.yml
```diff
@@ -1,12 +1,9 @@
-enterEmailOrUsername:
-  en: Please enter your email address or user name
-  de: Bitte geben Sie Ihre E-Mail-Adresse oder Ihren Benutzernamen an
+enterEmail:
+  en: Please enter your email address
+  de: Bitte geben Sie Ihre E-Mail-Adresse
```
ahelmberger, line 3: ... an

src/components/pages/login.yml
```diff
@@ -1,3 +1,10 @@
 forgotPassword:
   en: Forgot your password?
   de: Kennwort vergessen?
+hintContent:
+  en: |
+    The username is no longer supported for logging in.
+    It continues to be shown as the name under which you author website content.
+  de: |
+    Der Benutzername kann nicht mehr für die Anmeldung benutzt werden.
+    Der bleibt als Autorennamen für alle von Ihnen bearbeiteten Website-Inhalt.
```
ahelmberger, line 10: Autorenname

src/resources/common.yml
```diff
@@ -213,9 +216,9 @@ id:
 user:
   en: User
   de: Benutzer
-username:
-  en: User name
-  de: Benutzername
+displayName:
+  en: Display name
+  de: Anzeigenamen
```
ahelmberger, line 221: Anzeigename

src/domain/schemas/user-schema.spec.js
```diff
@@ -9,14 +10,18 @@ import {
...
+const invalidDisplayNameCases = [
   {
-    description: 'username is shorter than min length',
-    body: { username: 'joedo' }
+    description: 'displayName is shorter than min length',
+    body: { displayName: [...Array(minDisplayNameLength - 1).keys()].join('') }
+  },
+  {
+    description: 'displayName is longer than max length',
+    body: { displayName: [...Array(maxDisplayNameLength + 1).keys()].join('') }
```
ahelmberger, line 24: As `maxDisplayNameLength` is > 9, the keys in this array start to be 2 characters long each. Probably better would be something like

~~~
Array.from({ length: maxDisplayNameLength + 1 }, () => 'x')

src/components/pages/login.yml
adrianaluchian: "Er", of course.. I formed the bad habit of saying "Der" in such context all the time.
Thanks, corrected all of the translations you mentioned 🙏

src/components/displayName-form-item.yml
adrianaluchian, line 9: Will change these 2 in Nominativ case as well.

ahelmberger: Actually, it should be "Der Anzeigename **darf** höchstens {length} Zeichen lang sein" (as oppoesed to "Der Anzeigenamen **muss** mindestens {length} Zeichen lang sein")

ahelmberger:
Looks good!

src/components/displayName-form-item.yml
adrianaluchian: Thanks for the correction and side-explanation! Updated

src/domain/schemas/user-schema.spec.js
adrianaluchian: That's a good point, I changed it as you mentioned and added the positive cases too (which would have caught the setup problem actually, if they had existed before)

adrianaluchian:
> Looks good! Hard to tell if there are some hidden problems.

Even with all my testing, anything hidden would come up later on anyway. Will ask Theresa to test everything.

src/domain/schemas/user-schema.spec.js
ahelmberger: Cool, thanks!

ahelmberger (approved):

PR #524, opened 2022-07-26
EDU-695 Change user profile data

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-695


migrations/educandu-2022-07-25-01-change-user-model.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable camelcase */
+
+export default class Educandu_2022_07_25_01_change_user_model {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    await this.db.collection('users').updateMany({}, { $unset: { profile: null } });
+    await this.db.collection('users').updateMany({}, { $set: { organization: null, introduction: null } });
```
ahelmberger, line 10: As these are text fields filled in by the user, I would make them empty string by default. `null` sounds more like there should be an object in this place that's missing. What do you think?

adrianaluchian: When creating a user default object, we default everything to `null`. I also agreed with your note in this other PR (https://github.com/educandu/educandu/pull/510#discussion_r924264917) and I'd like to continue making this distinction between `null` and empty string.

ahelmberger: If I understood correctly, you agree that `organization` and `introduction` should be empty string rather than `null` in this migration script?

adrianaluchian: I meant quite the opposite 😄  As we discussed in https://github.com/educandu/educandu/pull/510#discussion_r924264917

adrianaluchian: I think we should keep `null` as we do for all other user data in `userService._buildEmptyUser`, as meaning lack of data instead of intentionally set (by the user) empty data.

ahelmberger: I was waiting for a change here. As discussed in https://github.com/educandu/educandu/pull/510#discussion_r924264917, user provided fields (or "text intentionally left empty by the user") should be empty string rather than `null`. Doesn't this apply to `organization` and `introduction`?

ahelmberger (approved):

PR #529, opened 2022-07-27
EDU-688 Make timeline relative

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-688


migrations/educandu-2022-07-27-01-use-percentages-in-interactive-media-plugin.js
```diff
@@ -0,0 +1,76 @@
...
+        updateCount += 1;
+        await this.db.collection(collectionName).replaceOne({ _id: doc._id }, doc);
+      }
+    }
+
+    return updateCount;
+  }
+
+  async up() {
+    const lessonsCount = await this.processCollection('lessons');
```
adrianaluchian, line 66: I think it's safe to say you could drop this lessons migration now.

migrations/educandu-2022-07-27-01-use-percentages-in-interactive-media-plugin.spec.js
```diff
@@ -0,0 +1,149 @@
+import MigrationScript from './educandu-2022-07-27-01-use-percentages-in-interactive-media-plugin.js';
+
+describe('educandu-2022-07-27-01-use-percentages-in-interactive-media-plugin', () => {
```
adrianaluchian, line 3: Good that you added tests here! 👍 The script is a bit complex to follow

src/components/media-player-track.js
```diff
@@ -29,24 +29,28 @@ function MediaPlayerTrack({
...
 
-  const seekToStartIfNecessary = duration => {
-    if (duration && ((startTimecode && lastProgress < startTimecode) || (stopTimecode && lastProgress >= stopTimecode))) {
-      playerRef.current.seekTo((startTimecode || 0) / duration);
+  const seekToStartIfNecessary = newSourceDuration => {
+    const newTrackStartTimecode = playbackRange[0] * newSourceDuration;
+    const newTrackStopTimecode = playbackRange[1] * newSourceDuration;
+
+    if (newSourceDuration && ((lastProgressTimecode < newTrackStartTimecode) || (lastProgressTimecode >= newTrackStopTimecode))) {
+      playerRef.current.seekTo(newTrackStartTimecode / newSourceDuration);
```
adrianaluchian, line 53: Here I think it makes sense to early return if `!newSourceDuration`, cause the calculation of the 2 values and the `if` evaluation are not needed otherwise. Would make the code more clear too.

src/components/timeline.js
```diff
@@ -58,19 +64,19 @@ function Timeline({ length, parts, selectedPartIndex, onPartAdd, onPartDelete, o
...
 
   const handleSegmentsBarClick = () => {
-    if (isTouchDevice()) {
+    if (!timelineState.currentTimelineWidth || isTouchDevice()) {
       return;
     }
 
     if (newMarkerState?.isInBounds) {
-      const startTimecode = Math.round(newMarkerState.left / timelineState.msToPxRatio);
+      const startPosition = Math.max(0, Math.min(1, newMarkerState.left / timelineState.currentTimelineWidth));
```
adrianaluchian, line 77: Maybe this `Math.max(0, Math.min(1, <position>, <limit>))` could be extracted as a well named helper method, to make the code easier to grasp. Something like `getValueInUnitInterval` maybe, as in mathematics this [0,1] range is apparently called so.

adrianaluchian:
Works smoothly, could not spot any glitched 👏 And you've fit the changes nicely without major rewriting of the components involved, so it was easy enough to follow the logic.

What we could do in a second ticket, I realized now, would be to show the timecode above the marker, when it is being dragged. Right now the adjustment is a bit blind (unless you've selected the edited chapter and you keep an eye on the time intervals changing in it).

src/components/timeline.js
ahelmberger: Good idea! I went for `ensureValidMediaPosition` so I could put it into `media-utils` (I couldn't find a good place otherwise).

migrations/educandu-2022-07-27-01-use-percentages-in-interactive-media-plugin.js
ahelmberger: True, dropped!

src/components/media-player-track.js
ahelmberger: Good idea, we do that in most other places anyway, so it aligns. Done.

adrianaluchian (approved):

PR #537, opened 2022-08-01
EDU-712 Fix markdown input with preview UI case

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-712

### DIFF
--- a/src/components/input-and-preview.less
+++ b/src/components/input-and-preview.less
@@ -1,7 +1,7 @@
 .InputAndPreview {
   display: grid;
   gap: 5px;
-  grid-template-columns: 1fr 1fr;
+  grid-template-columns: 50% 50%;
 
   @media @edu-media-phone {
     grid-template-columns: 1fr;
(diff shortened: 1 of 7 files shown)


src/components/input-and-preview.less
```diff
@@ -17,6 +17,6 @@
   background-color: @edu-background-color-light;
   padding: 4px 11px;
   display: grid;
-  min-height: @edu-control-height-base;
+  min-height: 2 * @edu-control-height-base;
```
ahelmberger, line 20: This will probably break the inline preview (which is a one-line input), although we do not have any of those on the page.

src/components/markdown-input.less
```diff
@@ -39,6 +39,7 @@
 }
 
 .MarkdownInput-preview {
+  overflow-x: scroll;
```
ahelmberger, line 42: `scroll` shows permanent scrollbars on most systems, `auto` would be better.

src/components/input-and-preview.less
adrianaluchian: Good point, fixed with passing `inline` prop and adjusting min height accordingly.

src/components/markdown-input.less
adrianaluchian: Fixed with another CSS approach, thanks!

ahelmberger (approved):

PR #545, opened 2022-08-02
EDU-711 Reduce internal source url prefix

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-711

### DIFF
--- a/src/domain/constants.js
+++ b/src/domain/constants.js
@@ -155,6 +155,8 @@ export const HORIZONTAL_ALIGNMENT = {
   right: 'right'
 };
 
+export const CDN_URL_PREFIX = 'cdn://';
+
 export const AVATAR_SIZE = 110;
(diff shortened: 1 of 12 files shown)


src/domain/constants.js
```diff
@@ -155,6 +155,8 @@ export const HORIZONTAL_ALIGNMENT = {
   right: 'right'
 };
 
+export const CDN_URL_PREFIX = 'cdn://';
```
ahelmberger, line 158: This could also be used in `storage-service` and `storage-utils`.

adrianaluchian: Nice, thanks for spotting those. Used there too.

ahelmberger: :+1:

ahelmberger (approved):

PR #553, opened 2022-08-09
EDU-680 Extended file upload overview

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-680


src/components/image-editor.js
```diff
@@ -0,0 +1,178 @@
...
+
+function ImageEditor({ file, editorRef, onCrop }) {
+  const { t } = useTranslation('imageEditor');
+  const onCropRef = useRef(null);
+  const cropperImageRef = useRef(null);
+  const cropperInstanceRef = useRef(null);
+  const [cropperModule, setCropperModule] = useState(null);
+
+  editorRef.current = {
+    getCroppedFile: async (maxWidth = Infinity, quality = 0.5) => {
```
adrianaluchian, line 29: I guess here you could also use `IMAGE_OPTIMIZATION_QUALITY` as the default.

src/components/files-upload-overview.yml
```diff
@@ -7,4 +7,18 @@ uploadLimitExceeded:
...
   de: Nicht genügend Speicherplatz für die Datei verfügbar
-
+startUpload:
+  en: Start upload
+  de: Upload starten
+preprocessed:
+  en: edited
+  de: bearbeitet
+stage_uploadNotStarted:
+  en: The following file will be uploaded. You can edit images before upload by clicking on the pen icon.
```
adrianaluchian, line 17: The following file**s** [..]

adrianaluchian:
This feature is so cool!! Looks slick and elegant, works fast and brings sooo many tweaking options! 
Great job 👏

src/components/image-editor.js
```diff
@@ -0,0 +1,178 @@
...
+           && eventData.scaleX === 1
+           && eventData.scaleY === 1;
+
+          onCropRef.current?.({ ...eventData, isCropped: !isUnchainged });
+        }
+      });
+    }
+  }, [cropperModule, file, cropperInstanceRef, cropperImageRef, onCropRef, cleanUp]);
+
+  const handleCropperAction = (method, ...args) => {
```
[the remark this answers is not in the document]
ahelmberger: Sure, renamed.

src/components/image-editor.js
ahelmberger: Good idea, done.

src/components/files-upload-overview.yml
ahelmberger: I changed it using the fancy pluralization feature in our translations :-)

ahelmberger:
> The "Cancel" button should probably still be enabled on the screen before the user clicks "Start upload", for them to be able to cancel the process if they want to.

True, that was a tricky one, as it is in the parent component. I think we should do a refactoring of the whole `StorageLocation` component + children soon in order to unify the different screens we have now and figure out how to do the styling better to fit into the modal window. For now it works and looks OK, but it is already pretty convoluted.

ahelmberger:
> This feature is so cool!! Looks slick and elegant, works fast and brings sooo many tweaking options! Great job clap

Thank you :bowing_man: :relaxed:

adrianaluchian:
> > The "Cancel" button should probably still be enabled on the screen before the user clicks "Start upload", for them to be able to cancel the process if they want to.
> 
> True, that was a tricky one, as it is in the parent component. I think we should do a refactoring of the whole `StorageLocation` component + children soon in order to unify the different screens we have now and figure out how to do the styling better to fit into the modal window. For now it works and looks OK, but it is already pretty convoluted.

Agreed, now that we have the full picture so to say, with all screens that we wanted, we can reorganize it better.

adrianaluchian (approved):

PR #554, opened 2022-08-09
EDU-732 Make annotation titles support markdown

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-732

### DIFF
--- a/src/plugins/annotation/annotation-info.js
+++ b/src/plugins/annotation/annotation-info.js
@@ -50,6 +50,11 @@ class AnnotationInfo {
   redactContent(content, targetRoomId) {
     const redactedContent = cloneDeep(content);
 
+    redactedContent.title = this.gfm.redactCdnResources(
+      redactedContent.title,
+      url => isAccessibleStoragePath(url, targetRoomId) ? url : ''
+    );
+
     redactedContent.text = this.gfm.redactCdnResources(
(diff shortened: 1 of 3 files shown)


src/plugins/annotation/annotation-info.js
```diff
@@ -59,7 +64,12 @@ class AnnotationInfo {
   }
 
   getCdnResources(content) {
-    return this.gfm.extractCdnResources(content.text);
+    const cdnResources = [];
+
+    cdnResources.push(...this.gfm.extractCdnResources(content.title));
+    cdnResources.push(...this.gfm.extractCdnResources(content.text));
```
adrianaluchian, line 70: Oh this was missing too.. Which made me realize we should probably do a CDN resource regeneration and document regeneration on ELMU (before final import to OMA) and on OMA (after this change is deployed). 
We have had a few more scenarios in the past, so would be good to refresh the data 🤔

adrianaluchian (approved):

src/plugins/annotation/annotation-info.js
ahelmberger: I thought the same. Although it didn't render markdown before, so probably nobody put markdown there until now. Nevertheless, let's recreate them after releasing.

PR #556, opened 2022-08-09
EDU-731 Show solid red border around invalid sections

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-731

### DIFF
--- a/src/components/section-display.js
+++ b/src/components/section-display.js
@@ -50,7 +50,7 @@ function SectionDisplay({
 
   const [isEditing, setIsEditing] = useState(false);
   const [isInvalid, setIsInvalid] = useState(false);
-  const [isHelpModelVisible, setIsHelpModalVisible] = useState(false);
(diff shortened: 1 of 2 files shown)


src/components/section-display.js
```diff
@@ -48,19 +48,21 @@ function SectionDisplay({
 
   const isHardDeleteEnabled = canHardDelete && !section.deletedOn;
 
+  const [isEditing, setIsEditing] = useState(false);
+  const [isInvalid, setIsInvalid] = useState(false);
+  const [isHelpModelVisible, setIsHelpModalVisible] = useState(false);
```
ahelmberger, line 53: I know this was there before, but it'd be a good opportunity to rename the model to modal :laughing:

ahelmberger (approved):

src/components/section-display.js
adrianaluchian: Yes, we don't want these species mixed! 😱 Thanks!

PR #559, opened 2022-08-10
EDU-726 Include multitrack-media plugin in list

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-726

(Also switched around the table and interactive-media plugins, so that we'll have all the fancy media plugins at the end)

### DIFF
--- a/src/plugins/multitrack-media/multitrack-media.yml
+++ b/src/plugins/multitrack-media/multitrack-media.yml
@@ -1,6 +1,6 @@
 name:
   en: Multitrack media
-  de: Mehrspur Medien
+  de: Mehrspur-Medien
 mainTrack:
   en: Main track
   de: Hauptspur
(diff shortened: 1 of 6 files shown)


ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-726
> 
> (Also switched around the table and interactive-media plugins, so that we'll have all the fancy media plugins at the end)

:+1:

src/plugins/multitrack-media/multitrack-media.yml
```diff
@@ -1,3 +1,6 @@
+name:
+  en: Multitrack media
+  de: Mehrspur Medien
```
ahelmberger, line 3: Better would be to have a hyphen between the words, I guess

ahelmberger (approved):

src/plugins/multitrack-media/multitrack-media.yml
adrianaluchian: 👍 Added

PR #561, opened 2022-08-10
EDU-690 Fix reveal effect image loading order

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-690

### DIFF
--- a/src/plugins/image/image.yml
+++ b/src/plugins/image/image.yml
@@ -9,7 +9,7 @@ hoverOption:
   de: Hover
 revealOption:
   en: Reveal
-  de: Auflösung
+  de: Aufdeckung
 clipOption:
   en: Clip
   de: Zuschneiden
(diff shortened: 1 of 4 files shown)


src/plugins/image/image.yml
```diff
@@ -9,7 +9,7 @@ hoverOption:
   de: Hover
 revealOption:
   en: Reveal
-  de: Auflösung
+  de: Aufdeckung
```
ahelmberger, line 12: Maybe let's go with "Aufdecken" (similar to "Zuschneiden")

ahelmberger (approved):

src/plugins/image/image.yml
adrianaluchian: Also aligns with "Hover" then, changed 👍

ahelmberger: Poifect!

ahelmberger (approved):

PR #562, opened 2022-08-15
EDU-746 Remove public rooms

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-746

### DIFF
--- a/src/api-clients/room-api-client.js
+++ b/src/api-clients/room-api-client.js
@@ -47,7 +47,7 @@ class RoomApiClient {
       .then(res => res.data);
   }
 
-  deleteAllPrivateRoomsForUser({ ownerId }) {
+  deleteAllRoomsForUser({ ownerId }) {
     return this.httpClient
       .delete(
         `/api/v1/rooms?ownerId=${encodeURIComponent(ownerId)}`,
--- a/src/api-clients/room-api-client.js
+++ b/src/api-clients/room-api-client.js
@@ -1,5 +1,4 @@
 import HttpClient from './http-client.js';
-import { ROOM_ACCESS } from '../domain/constants.js';
 
 class RoomApiClient {
   static inject() { return [HttpClient]; }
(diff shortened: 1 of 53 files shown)


src/api-clients/room-api-client.js
```diff
@@ -51,7 +50,7 @@ class RoomApiClient {
   deleteAllPrivateRoomsForUser({ ownerId }) {
```
ahelmberger, line 50: We could even rename it to `deleteRoomsForUser`

src/components/pages/users.less
```diff
@@ -1,26 +1,11 @@
...
-
-  .UsersPage-roleTag--superUser {
-    color: @red-10;
-  }
-
-  .UsersPage-roleTag--editor {}
-
-  .UsersPage-roleTag--superEditor {}
-}
-
```
ahelmberger, line 15: Ha :rofl:

src/api-clients/room-api-client.js
adrianaluchian: Course, forgot the client-side, I had renamed just server-side 😵‍💫 
Thanks, done!

src/components/pages/users.less
adrianaluchian: Exactly my reaction 😂

src/server/document-controller.js
```diff
@@ -84,8 +80,8 @@ class DocumentController {
 
     const room = doc.roomId ? await this.roomService.getRoomById(doc.roomId) : null;
 
-    if (room?.access === ROOM_ACCESS.private && !isRoomOwnerOrMember({ room, userId: user._id })) {
-      throw new Forbidden();
+    if (!!room && !isRoomOwnerOrMember({ room, userId: user._id })) {
+      throw new Unauthorized();
```
ahelmberger, line 84: I think this was correct before. `Unauthorized` (401, bad naming for "unauthenticated") will simply redirect the user to the login page (and therefore into an endless loop), while we want `Forbidden` (403) :thinking:

adrianaluchian: Ah so this isn't actually `Unauthorized` 😱 Would have been better than `Forbidden`.. Damn, ok, will revert all these places where I changed it.

adrianaluchian: Reverted. Again what learned, thanks!

ahelmberger: Yeah, not optimal naming from the early days of the web :-) Thanks!

ahelmberger:
Such a joy to see a lot of conditions vanish :smile:

ahelmberger (approved):

adrianaluchian:
> Such a joy to see a lot of conditions vanish 😄

Yes! Glad we're able to simplify before we add some more complexity 😬

PR #564, opened 2022-08-15
EDU-753 Delete dueOn from documents

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-753

**Before migration**
Screen Shot 2022-08-15 at 15 19 15

**After migration**
Screen Shot 2022-08-15 at 15 19 54


src/components/document-metadata-modal.js
```diff
@@ -272,32 +218,21 @@ function DocumentMetadataModal({
...
-                <InputNumber style={{ width: '100%' }} disabled={!hasDueDate} min={1} max={100} />
-              </FormItem>
-            </CollapsePanel>
-          </Collapse>
+          <FormItem
+            name="sequenceCount"
+            rules={[{ type: 'integer', min: 1, max: 100 }]}
+            label={
+              <Fragment>
+                <span>{t('createSequence')}</span>
```
ahelmberger, line 227: This was once the header ("Reihe anlegen"), now it is a label for a number input. I'm not sure if it will be clear to the users that they have to enter the number of documents there. Maybe we should at least mention in the hover info text that they can type there how many documents should be generated? :thinking:

ahelmberger: On second thought, maybe it is clear. I'm not sure here and maybe over-concerned. You can also just ignore this comment.

ahelmberger (approved):

src/components/document-metadata-modal.js
adrianaluchian: I had a hard time deciding on this one too. Was not sure how to combine "Reihe anlegen" with "Anzahl". I think the info text does make it clear though. 
In lack of better ideas, let's just see what the others will think. I can actually leave a comment on the ticket for Theresa and ask her if she thinks it's clear enough for a user or has suggestions.

ahelmberger: :+1:

PR #571, opened 2022-08-24
EDU-724 Create multitrack media player

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-724


src/plugins/multitrack-media/multitrack-media.less
```diff
@@ -1,3 +1,13 @@
 .MultitrackMediaDisplay {}
 
+.MultitrackMediaDisplay-content {
+  margin: 0 auto;
+}
+
+.MultitrackMediaDisplay-errorMessage {
+  font-weight: bold;
```
adrianaluchian, line 8: From my side I think we can drop this, makes the UI a bit too aggressive 🤔

ahelmberger: True, especially as this is kind of the initial default view. Dropped.

adrianaluchian:
In edit mode, when we switch `sourceType` the `sourceUrl` is also reset, but the mixer still shows the old data

adrianaluchian:
Such a cool and fun plugin 😍 Good job!

adrianaluchian:
Small UI suggestion, if you feel it would be better, would be to write the times on the right hand side of the bar (as they represent the length, they might be more intuitive if placed there):

Screen Shot 2022-08-24 at 15 32 36

Not sure about the case of overflow to the right, if times should overflow a bit too:
Screen Shot 2022-08-24 at 15 31 57

ahelmberger:
> In edit mode, when we switch `sourceType` the `sourceUrl` is also reset, but the mixer still shows the old data

I'd like to fix this (and other issues in this panel) with https://educandu.atlassian.net/browse/EDU-761 as I think we might want to change the way the track lengths are determined there anyway.

ahelmberger:
> Such a cool and fun plugin heart_eyes Good job!

Thanks a lot :smile:

adrianaluchian (approved):

PR #574, opened 2022-08-24
EDU-756 adjust search relevance logic

ahelmberger:
https://educandu.atlassian.net/browse/EDU-756

This is a very naïve implementation making verified documents simply counting "3 matched tags more". For more sophisticated system we would have to also change the way we make tags count and open a bigger box. So maybe we can try and see how far we can get with this solution?

### DIFF
--- a/src/components/pages/search.js
+++ b/src/components/pages/search.js
@@ -44,7 +44,7 @@ function Search({ PageTemplate }) {
   ];
 
   const sorters = useMemo(() => ({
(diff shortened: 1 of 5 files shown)


adrianaluchian:
> https://educandu.atlassian.net/browse/EDU-756
> 
> This is a very naïve implementation making verified documents simply counting "3 matched tags more". For more sophisticated system we would have to also change the way we make tags count and open a bigger box. So maybe we can try and see how far we can get with this solution?

Sure, let's go with this first and gather some feedback, good idea.

adrianaluchian (approved):

PR #579, opened 2022-08-29
EDU-761 Finalize track mixer

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-761


src/utils/media-utils.js
```diff
@@ -61,10 +61,14 @@ export const determineMediaDuration = memoizee(url => {
       cleanedUp = true;
     }
   };
+
+  await Promise.resolve();
```
adrianaluchian, line 65: Why is this little hack here needed? 🤔

src/components/media-player/media-duration-cache.js
```diff
@@ -0,0 +1,116 @@
...
+      dispose: () => {
+        this._entries.clear();
+        this._subscribers.clear();
+        this._isDisposed = true;
+      }
+    };
+  }
+}
+
+export default MediaDurationCache;
```
adrianaluchian, line 116: Nice solution! 👍 
I thought initially that we might have issues when re-uploading the same video with an altered duration, but actually it will get a different name in the CDN so that's not an issue.

src/components/media-player/media-hooks.js
```diff
@@ -0,0 +1,27 @@
+import deepEqual from 'fast-deep-equal';
+import { useEffect, useState } from 'react';
+import { useService } from '../container-context.js';
+import MediaDurationCache from './media-duration-cache.js';
+
+export function useMediaDuration(urlOrUrls) {
```
adrianaluchian, line 6: Would it be ok for you if we always worked with arrays here?
I'm not a big fan of this 2-in-1 function signature pattern 😬

adrianaluchian:
I get a series of `propType` warnings in the console when I add a new `multitrack-media` section.

src/components/media-player/media-hooks.js
ahelmberger: Sure, I'll change it :+1:

ahelmberger: Done!

src/components/media-player/media-duration-cache.js
ahelmberger: Yes. Only for external resources we cannot "guarantee".

ahelmberger:
> Found a few glitches:
> 
> * I get a series of `propType` warnings in the console when I add a new `multitrack-media` section and switch the section to edit mode
> * When I set both main and secondary tracks from youtube, I can play the video in the preview but the durations are not shown, neither is the video progress
> 
> Screen Shot 2022-08-29 at 17 11 00
> 
> * After I've set both tracks from youtube, with the secondary track shorter than the main track, switch to preview and then back to edit, the main track gets a broken overflow styling
> 
> Screen Shot 2022-08-29 at 17 07 09

I could reproduce and fix the first and third point, but for me the second scenario was working correctly. Could you please provide me the steps and link URLs for which this happens?

adrianaluchian (approved):

PR #581, opened 2022-08-29
EDU-764 Create backend for comments

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-764

### DIFF
--- a/src/domain/permissions.js
+++ b/src/domain/permissions.js
@@ -9,6 +9,7 @@ const REVIEW_DOC = 'review-doc';
 const VERIFY_DOC = 'verify-doc';
 const VIEW_FILES = 'view-files';
 const EDIT_USERS = 'edit-users';
+const JOIN_ROOMS = 'join-rooms';
 const CREATE_FILE = 'create-file';
 const MIGRATE_DATA = 'migrate-data';
 const VIEW_BATCHES = 'view-batches';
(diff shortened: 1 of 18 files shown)


src/domain/permissions.js
```diff
@@ -115,5 +119,7 @@ export default {
   OWN_ROOMS,
   DELETE_FOREIGN_ROOMS,
   AUTORIZE_ROOMS_RESOURCES,
-  JOIN_PRIVATE_ROOMS
+  JOIN_ROOMS,
+  CREATE_DOCUMENT_COMMENT,
+  MANAGE_DOCUMENT_COMMENT
```
ahelmberger, line 124: In combination with db "entitites" we mostly used the plural so far (like "manage-storage-plans" etc.). I'm not insisting, just preferring :-)

src/domain/schemas/comment-schemas.js
```diff
@@ -0,0 +1,28 @@
...
+
+export const commentDBSchema = joi.object({
+  _id: idOrKeySchema.required(),
+  documentId: idOrKeySchema.required(),
+  createdOn: joi.date().required(),
+  createdBy: idOrKeySchema.required(),
+  deletedOn: joi.date().allow(null),
+  deletedBy: idOrKeySchema.allow(null),
+  topic: joi.string().max(maxCommentTopicLength).required(),
+  text: joi.string().max(maxCommentTextLength).allow(null)
```
ahelmberger, line 27: I think this is another one of the examples where you like `null` and I don't (as the user can hardly ever enter it), but we can postpone that for an eventual streamlining session.

src/server/comment-controller.js
```diff
@@ -0,0 +1,108 @@
...
+      '/api/v1/comments',
+      jsonParser,
+      needsPermission(permissions.CREATE_DOCUMENT_COMMENT),
+      validateBody(putCommentBodySchema),
+      (req, res) => this.handlePutComment(req, res)
+    );
+
+    router.post(
+      '/api/v1/comments/:commentId',
+      [needsPermission(permissions.MANAGE_DOCUMENT_COMMENT), jsonParser, validateBody(postCommentBodySchema)],
```
ahelmberger, line 95: Little thing: we should also validate the `:commentId` param

src/server/comment-controller.js
```diff
@@ -0,0 +1,108 @@
...
+  async handlePutComment(req, res) {
+    const { user } = req;
+    const data = req.body;
+
+    const document = await this.documentService.getDocumentById(data.documentId);
+
+    if (!document) {
+      throw new BadRequest(`Unknown document id '${data.documentId}'`);
+    }
+
```
ahelmberger, line 42: Maybe we could check here also if the document is an internal one and throw if it isn't. Just to make sure...?

src/domain/permissions.js
adrianaluchian: Nicely spotting, I of course aligned the namings, thanks!

src/domain/schemas/comment-schemas.js
adrianaluchian: I agree with the empty string default, changed!

src/server/comment-controller.js
adrianaluchian: Indeed, added validation.

ahelmberger: :+1:

ahelmberger (approved):

PR #584, opened 2022-08-31
EDU-773 Fix control panel width issue

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-773


ahelmberger:
Works even nicer than before :+1:

PR #585, opened 2022-08-31
EDU-771 Avoid gaps in flex layout because of older Safaris

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-771

### DIFF
--- a/src/components/collapsible.less
+++ b/src/components/collapsible.less
@@ -13,12 +13,10 @@
 }
 
 .Collapsible-header {
-  display: flex;
-  align-items: center;
+  .m-control-group-wide;
   overflow-x: hidden;
   text-decoration: none;
   color: @edu-text-color;
-  gap: @edu-content-padding;
   padding: 12px @collapsible-padding;
   border-radius: @edu-border-radius-base;
   border: 1px solid @edu-border-color-light;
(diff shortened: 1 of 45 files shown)


src/components/collapsible.less
```diff
@@ -13,12 +13,10 @@
 }
 
 .Collapsible-header {
```
adrianaluchian, line 15: The collapsible arrow is no longer aligned to the right, which I find looks not so good (especially when the title is omitted):

Screen Shot 2022-08-31 at 17 21 49

src/components/pages/room.less
```diff
@@ -5,25 +5,22 @@
...
+  padding: 0 @edu-content-padding;
 }
 
-.RoomPage-subtitleGroup {
-  display: flex;
-  gap: 7px;
-  align-items: center;
+.RoomPage-leaveRoomLink {
+  .m-control-group;
+  padding-left: @edu-control-spacing-wide;
```
adrianaluchian, line 15: The "leave room" link is no longer visible 🤔 

How it looks on Int:
Screen Shot 2022-08-31 at 17 29 09

How it looks locally:
Screen Shot 2022-08-31 at 17 30 02

ahelmberger: Ha! I reverted the condition in code in order to see the link while developing and forgot to set it back. Thanks for spotting it, it's fixed now.

src/components/collapsible.less
ahelmberger: Nice catch! I haven't realized. It's fixed now.

adrianaluchian (approved):

PR #591, opened 2022-09-01
EDU-768 Transform image-tiles to catalog plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-768


ahelmberger:
> Works nicely! +1 Just a thought though: what about setting the `sourceType` of the `link` to "document" ? Since for image we set it to "internal", it would look aligned, also feel aligned (as if we prioritize internal over external), and the user wouldn't directly see this orange warning message for the `sourceUrl`.

Sounds good to me. Actually we're gonna change more default values in the upcoming ticket (https://educandu.atlassian.net/browse/EDU-769), so I'll just add it there, ok?

adrianaluchian:
> > Works nicely! +1 Just a thought though: what about setting the `sourceType` of the `link` to "document" ? Since for image we set it to "internal", it would look aligned, also feel aligned (as if we prioritize internal over external), and the user wouldn't directly see this orange warning message for the `sourceUrl`.
> 
> Sounds good to me. Actually we're gonna change more default values in the upcoming ticket (https://educandu.atlassian.net/browse/EDU-769), so I'll just add it there, ok?

Sure thing 🤝

adrianaluchian (approved):

PR #596, opened 2022-09-05
EDU-782 Escape tag text when searching tags

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-782

### DIFF
--- a/src/components/document-metadata-modal.js
+++ b/src/components/document-metadata-modal.js
@@ -4,6 +4,7 @@ import { useUser } from './user-context.js';
 import cloneDeep from '../utils/clone-deep.js';
 import { useTranslation } from 'react-i18next';
 import { useLocale } from './locale-context.js';
+import escapeStringRegexp from 'escape-string-regexp';
 import { InfoCircleOutlined } from '@ant-design/icons';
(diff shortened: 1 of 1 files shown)


src/components/document-metadata-modal.js
```diff
@@ -117,11 +118,12 @@ function DocumentMetadataModal({
   }, [isVisible]);
 
   const handleTagSearch = async typedInTag => {
+    const sanitizedTypedInTag = escapeStringRegexp((typedInTag || '').trim());
```
ahelmberger, line 121: Wouldn't it be better to escape the string in the store (directly when constructing the RegExp)? We do this already in another place. Also, in my opinion, the frontend shouldn't have the knowledge how the DB query is done (I mean using RegExp or not).

adrianaluchian: You have a point, the store is a better, safer place for this correction. Will do!

adrianaluchian: Fixed, placed it in the service, that's where we have all sanitizations. Thanks for the suggestion!

ahelmberger (approved):

PR #600, opened 2022-09-07
EDU-784 Create media slideshow plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-784


adrianaluchian:
Found image overlay bug, closing PR to fix.

PR #601, opened 2022-09-07
EDU-784 Create media slideshow plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-784

Screen Shot 2022-09-07 at 16 48 06

Screen Shot 2022-09-07 at 16 48 28


src/components/media-player/chapter-selector.yml
```diff
@@ -0,0 +1,6 @@
+selectPreviousChapter:
+  en: Select previous chapter
+  de: Vorheriges Kapitel wählen
```
ahelmberger, line 3: "auswählen" matches a little bit better

src/components/media-player/main-track-editor.js
```diff
@@ -65,13 +65,16 @@ function MainTrackEditor({ content, onContentChanged }) {
 
   const handleSourceTypeChange = event => {
     const { value } = event.target;
-    changeContent({
+    const newContent = {
       sourceType: value,
       sourceUrl: '',
       range: [0, 1],
```
ahelmberger, line 71: BTW I discovered this wrongly named property. It should be `playbackRange`. It's fixed in my PR (https://github.com/educandu/educandu/pull/602), but you could fix it here to to avoid conflicts.

src/plugins/media-slideshow/media-slideshow-info.js
```diff
@@ -0,0 +1,110 @@
...
+    const cdnResources = [];
+
+    cdnResources.push(...this.gfm.extractCdnResources(content.copyrightNotice));
+
+    if (content.sourceType === MEDIA_SOURCE_TYPE.internal && content.sourceUrl) {
+      cdnResources.push(content.sourceUrl);
+    }
+
+    content.chapters.forEach(chapter => {
+      if (chapter.image?.sourceType === MEDIA_SOURCE_TYPE.internal && chapter.image?.sourceUrl) {
```
ahelmberger, line 99: I removed the Elvis from the image in other places as we assume it's always an object, so we can be confident it's defined.

src/plugins/media-slideshow/media-slideshow.yml
```diff
@@ -0,0 +1,3 @@
+name:
+  en: Media slideshow
+  de: Slideshow-Medien
```
ahelmberger, line 3: Better: "Medien-Slideshow"

src/utils/url-utils.js
```diff
@@ -49,15 +49,24 @@ function ensureIsFullyQualifiedUrl(pathOrUrl, fallbackBase) {
 
 function getImageUrl({ cdnRootUrl, sourceType, sourceUrl }) {
   switch (sourceType) {
-    case IMAGE_SOURCE_TYPE.external:
-      return sourceUrl || null;
     case IMAGE_SOURCE_TYPE.internal:
       return sourceUrl ? `${cdnRootUrl}/${sourceUrl}` : null;
+    case IMAGE_SOURCE_TYPE.external:
```
ahelmberger, line 54: Actually, we could make `return sourceUrl || null;` the default case (like in the function below). We only need to prepend in case of internal URLs for both images and video/audio.

src/components/media-player/chapter-selector.yml
adrianaluchian: Changed!

src/components/media-player/main-track-editor.js
adrianaluchian: Huh.. Ok, fixed it in this PR too.

src/components/media-player/main-track-editor.js
```diff
@@ -164,27 +172,31 @@ function MainTrackEditor({ content, onContentChanged }) {
...
+            onChange={handleAspectRatioChanged}
+            disabled={![RESOURCE_TYPE.video, RESOURCE_TYPE.none].includes(getResourceType(sourceUrl))}
+            >
+            {Object.values(MEDIA_ASPECT_RATIO).map(ratio => (
+              <RadioButton key={ratio} value={ratio}>{ratio}</RadioButton>
+            ))}
+          </RadioGroup>
+        </FormItem>
+      )}
+      { useShowVideo && (
```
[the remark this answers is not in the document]
adrianaluchian: Bastard!... 🚪 🚶 👈

src/plugins/media-slideshow/media-slideshow-info.js
adrianaluchian: That's true, we ensure object. Removed, thanks!

src/plugins/media-slideshow/media-slideshow.yml
adrianaluchian: Made it better!

src/utils/url-utils.js
adrianaluchian: Agreed, simplified 👍

ahelmberger (approved):

PR #602, opened 2022-09-07
EDU-770 streamline null usage

ahelmberger:
**Let's merge https://github.com/educandu/educandu/pull/601 first, then I will have to update this one with the new slideshow plugin**

EDIT: It's updated!

Closes https://educandu.atlassian.net/browse/EDU-770


src/domain/schemas/document-schemas.js
```diff
@@ -71,29 +71,29 @@ export const documentSectionDBSchema = joi.object({
...
-  review: joi.string().allow(null).allow(''),
-  verified: joi.boolean(),
-  allowedOpenContribution: joi.string().valid(...Object.values(DOCUMENT_ALLOWED_OPEN_CONTRIBUTION)),
+  review: joi.string().allow('').required(),
+  verified: joi.boolean().required(),
+  allowedOpenContribution: joi.string().valid(...Object.values(DOCUMENT_ALLOWED_OPEN_CONTRIBUTION)).required(),
   archived: joi.boolean().required(),
   origin: joi.string().required(),
-  originUrl: joi.string().allow(null).allow(''),
+  originUrl: joi.string().allow(null).required(),
```
ahelmberger, line 90: This is an edge case, the URL is not editable by the user, but set by the system to the import origin, so I went with value or `null`.

adrianaluchian: Makes sense. And yes, will be cleaned up soon enough!

src/domain/schemas/shared-schemas.js
```diff
@@ -1,12 +1,11 @@
 import joi from 'joi';
+import { slugValidationPattern } from '../validation-constants.js';
 
 export const idOrKeySchema = joi.string().alphanum().min(15).max(30);
 
 export const boolStringSchema = joi.any().valid('true', 'false', true, false);
 
-// Should use `slugValidationPattern` from `validation-constants.js`,
-// as soon as we implement https://educandu.atlassian.net/browse/EDU-308
```
adrianaluchian, line 8: This was quite the forgotten debt, it's been a while.

ahelmberger: Same thought :-)

src/domain/schemas/user-schemas.js
```diff
@@ -57,3 +57,37 @@ export const loginBodySchema = joi.object({
   email: joi.string().required(),
   password: joi.string().required()
 });
+
+const storageReminderDBSchema = joi.object({
+  timestamp: joi.date().required(),
+  createdBy: idOrKeySchema.required()
```
adrianaluchian, line 63: I ask myself if this will ever end up being used, as we seem to not continue to develop any more storage allocation concepts..

ahelmberger: I don't understand, it is used :thinking: You mean the reminder, right?

src/domain/schemas/user-schemas.js
```diff
@@ -57,3 +57,37 @@ export const loginBodySchema = joi.object({
...
+});
+
+const storageDBSchema = joi.object({
+  plan: joi.string().allow(null).required(),
```
adrianaluchian, line 67: Should we make a small separate ticket to turn this into `planId` ?

ahelmberger: Nice idea, I will create one! :+1:

src/domain/schemas/user-schemas.js
adrianaluchian: Yes, the reminder feature.

ahelmberger: It is used then. (in the sense that admins can set a marker for whom they've sent a reminder)

src/plugins/video/video-display.js
```diff
@@ -20,13 +20,12 @@ function VideoDisplay({ content }) {
   }
 
   let posterImageUrl;
-  switch (content.posterImage?.sourceType) {
-    case MEDIA_SOURCE_TYPE.internal:
+  switch (content.posterImage.sourceType) {
+    case IMAGE_SOURCE_TYPE.internal:
```
adrianaluchian, line 24: 👍

src/domain/schemas/user-schemas.js
adrianaluchian: I know they **can**, I just wandered if the **do** 😬

ahelmberger: I see. As they don't charge yet, I guess not. Questionable if ever :man_shrugging:

migrations/educandu-2022-09-07-02-streamline-nulls-in-plugins.js
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  tryUpdateSection(section) {
+    let sectionWasUpdated = false;
+
+    if (!section.content) {
+      return sectionWasUpdated;
+    }
+
+    if (section.type === 'markdown' && 'renderMedia' in section.content) {
```
adrianaluchian, line 19: Oh, leftover 🤦‍♀️ Nice cleanup.

migrations/educandu-2022-09-07-02-streamline-nulls-in-plugins.js
```diff
@@ -0,0 +1,227 @@
...
+    if (section.type === 'image') {
+      if (section.content.copyrightNotice === null) {
+        // Convert null -> empty string
+        section.content.copyrightNotice = '';
+        sectionWasUpdated = true;
+      }
+      if (section.content.effect?.type === EFFECT_TYPE.clip && 'copyrightNotice' in section.content.effect) {
+        // Delete superfluous property
+        delete section.content.effect.copyrightNotice;
+        sectionWasUpdated = true;
```
adrianaluchian, line 52: With these effects, maybe it would have been better to have:
- either dedicated objects
- and the `effect.type` outside of the `effect` object
- both of the above
Let's see if it is worth changing this at some point..

migrations/educandu-2022-09-07-02-streamline-nulls-in-plugins.js
```diff
@@ -0,0 +1,227 @@
...
+        section.content.copyrightNotice = '';
+        sectionWasUpdated = true;
+      }
+    }
+
+    if (section.type === 'catalog') {
+      for (const item of section.content.items) {
+        // Ensure property
+        if (!('documentId' in item.link)) {
+          item.link.documentId = null;
```
adrianaluchian, line 86: It seems in code we don't ensure this property. I think we should update that.

migrations/educandu-2022-09-07-02-streamline-nulls-in-plugins.js
```diff
@@ -0,0 +1,227 @@
...
+      }
+    }
+
+    if (section.type === 'ear-training') {
+      for (const test of section.content.tests) {
+        // Migrate leftover older data structure
+        if ('startAbcCode' in test) {
+          test.mode = TEST_MODE.abcCode;
+          test.questionAbcCode = test.startAbcCode;
+          test.answerAbcCode = test.fullAbcCode;
```
adrianaluchian, line 103: Interesting 😳 Is it possible that the migration was already done, but without the cleanup?
If so, we probably should prioritize the existing data (`test.questionAbcCode = test.questionAbcCode || test.startAbcCode` for safety?)

src/plugins/interactive-media/interactive-media-info.js
```diff
@@ -48,15 +49,37 @@ class InteractiveMediaInfo {
...
+        startPosition: joi.number().min(0).max(1).required(),
+        title: joi.string().allow('').required(),
+        text: joi.string().allow('').required(),
+        answers: joi.array().items(joi.string().allow('')).required(),
+        correctAnswerIndex: joi.number().min(-1).required()
+      })).required()
+    });
+
+    joi.attempt(content, schema, { convert: false, noDefaults: true });
+  }
```
adrianaluchian, line 81: Really good that you added these plugin validation! 👏

migrations/educandu-2022-09-07-02-streamline-nulls-in-plugins.js
adrianaluchian: Yes, the validation throws since it expects the property.

migrations/educandu-2022-09-07-02-streamline-nulls-in-plugins.js
ahelmberger: I thought the same (towards `effect` never being `null` but always an object (even if one containing only `type: 'none'`). Let's create a ticket for that, it will also make the editor less complicated!

ahelmberger: I created a ticket.

adrianaluchian: 👍

migrations/educandu-2022-09-07-02-streamline-nulls-in-plugins.js
ahelmberger: The data I checked on prod had both the old and new props, so I didn't want to "revive" the old values from their grave (since nobody has so far complained that they have been lost obviously some while ago)

adrianaluchian: Ok, then R.I.P 😄

src/plugins/interactive-media/interactive-media-info.js
ahelmberger: Yeah, it brought a few things into daylight. From now on we can be more confident that we do not save broken data.

adrianaluchian: You know what we could do as an ultimate test of these changes?
Run a full import from ELMU into OMA on Staging after this ticket is brought to both.
Then if there are any inconsistencies, the import report will point them out.

migrations/educandu-2022-09-07-02-streamline-nulls-in-plugins.js
ahelmberger: I checked, and now I remember the reason for the code. All sections that have `startAbcCode` set, do not have any of the new properties (meaning they were never migrated), so the migration script will actually fix this older issue. There is no need for the `test.questionAbcCode || test.startAbcCode` fallback.

adrianaluchian: I see. Yet another good find then. Glad this whole correction will happen now with this PR.

adrianaluchian (approved):

PR #603, opened 2022-09-07
EDU-786 Fix showing percentages instead of timecodes in timelines

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-786

### DIFF
--- a/src/plugins/interactive-media/interactive-media-editor.js
+++ b/src/plugins/interactive-media/interactive-media-editor.js
@@ -1,17 +1,18 @@
 import by from 'thenby';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
+import urlUtils from '../../utils/url-utils.js';
 import validation from '../../ui/validation.js';
 import cloneDeep from '../../utils/clone-deep.js';
 import { removeItemAt } from '../../utils/array-utils.js';
 import ClientConfig from '../../bootstrap/client-config.js';
-import React, { Fragment, useEffect, useState } from 'react';
 import DeleteButton from '../../components/delete-button.js';
+import React, { Fragment, useEffect, useState } from 'react';
+import { Button, Divider, Form, Input, Tooltip } from 'antd';
 import MarkdownInput from '../../components/markdown-input.js';
 import InteractiveMediaInfo from './interactive-media-info.js';
 import Timeline from '../../components/media-player/timeline.js';
 import { useService } from '../../components/container-context.js';
-import { Button, Divider, Form, Input, Spin, Tooltip } from 'antd';
 import { sectionEditorProps } from '../../ui/default-prop-types.js';
 import { useNumberFormat } from '../../components/locale-context.js';
 import MediaPlayer from '../../components/media-player/media-player.js';
@@ -19,6 +20,7 @@ import ObjectWidthSlider from '../../components/object-width-slider.js';
 import ChapterSelector from '../../components/media-player/chapter-selector.js';
 import MainTrackEditor from '../../components/media-player/main-track-editor.js';
 import { MEDIA_SCREEN_MODE, MEDIA_SOURCE_TYPE } from '../../domain/constants.js';
+import { useMediaDurations } from '../../components/media-player/media-hooks.js';
 import { formatMediaPosition, getFullSourceUrl } from '../../utils/media-utils.js';
 import { CheckOutlined, InfoCircleOutlined, PlusOutlined } from '@ant-design/icons';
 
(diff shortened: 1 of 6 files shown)


src/plugins/interactive-media/interactive-media-editor.js
```diff
@@ -42,13 +44,14 @@ function InteractiveMediaEditor({ content, onContentChanged }) {
...
   const { t } = useTranslation('interactiveMedia');
-  const [sourceDuration, setSourceDuration] = useState(0);
   const interactiveMediaInfo = useService(InteractiveMediaInfo);
   const [selectedChapterIndex, setSelectedChapterIndex] = useState(0);
   const [selectedChapterFraction, setSelectedChapterFraction] = useState(0);
-  const [isDeterminingDuration, setIsDeterminingDuration] = useState(false);
   const { sourceType, sourceUrl, playbackRange, chapters, width, showVideo } = content;
 
+  const [mediaDuration] = useMediaDurations([urlUtils.getMediaUrl({ cdnRootUrl: clientConfig.cdnRootUrl, sourceType, sourceUrl })]);
+  const sourceDuration = mediaDuration.duration;
```
ahelmberger, line 53: `mediaDuration.duration` could be `null` (if it is still in the progress of being determined), so we cannot use it unguarded (to achieve the former functionality, `|| 0` should do the trick).

src/plugins/interactive-media/interactive-media-editor.js
```diff
@@ -289,13 +281,6 @@ function InteractiveMediaEditor({ content, onContentChanged }) {
           </Fragment>
         )}
       </Form>
-
-      {isDeterminingDuration && (
```
ahelmberger, line 293: We could still show this when `mediaDuration.isResolved === false`

src/plugins/media-slideshow/media-slideshow-editor.js
```diff
@@ -39,14 +40,15 @@ function MediaSlideshowEditor({ content, onContentChanged }) {
...
-  const [sourceDuration, setSourceDuration] = useState(0);
   const mediaSlideshowInfo = useService(MediaSlideshowInfo);
   const [playingChapterIndex, setPlayingChapterIndex] = useState(0);
   const [selectedChapterIndex, setSelectedChapterIndex] = useState(0);
   const [selectedChapterFraction, setSelectedChapterFraction] = useState(0);
-  const [isDeterminingDuration, setIsDeterminingDuration] = useState(false);
   const { sourceType, sourceUrl, playbackRange, chapters, width } = content;
 
+  const [mediaDuration] = useMediaDurations([urlUtils.getMediaUrl({ cdnRootUrl: clientConfig.cdnRootUrl, sourceType, sourceUrl })]);
+  const sourceDuration = mediaDuration.duration;
```
ahelmberger, line 50: See above

src/plugins/interactive-media/interactive-media-editor.js
adrianaluchian: Actually I had checked that and it cannot be `null`, it return the default "unresolved entry", the same object as for resolved, with default values and `isResolved` set to `false`

adrianaluchian: Ah sorry, `duration` could be `null` you said, not the `mediaDuration`.
Honestly I would always default numbers to `0` unless they are IDs, like we said for the recent streamlining. Unless we saw that part differently actually.
What do you think?

src/plugins/interactive-media/interactive-media-editor.js
adrianaluchian: That's true. However do you think it's necessary? In most cases it would just be a very quick flash on the screen that the user wouldn't be able to read.
If the duration determination takes too long, nothing is actually broken in the UI anymore, everything is usable still and shown in percentages.
Thought?

src/plugins/interactive-media/interactive-media-editor.js
ahelmberger: For me that would be fine, too. We could change it that way.

src/plugins/interactive-media/interactive-media-editor.js
ahelmberger: Necessary, no. Just if we want the exact old behavior. But I think we do not indicate the loading in other components, so we could drop it here, too.

src/plugins/interactive-media/interactive-media-editor.js
adrianaluchian: Thanks, changed that way!

src/plugins/interactive-media/interactive-media-editor.js
adrianaluchian: 👍

src/plugins/interactive-media/interactive-media-editor.js
ahelmberger: :+1:

ahelmberger (approved):

PR #607, opened 2022-09-08
EDU-791 Make media-slideshow transition a bit smoother

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-791

### DIFF
--- a/src/utils/image-utils.js
+++ b/src/utils/image-utils.js
@@ -1,0 +1,26 @@
+import ReactDOM from 'react-dom';
+
+export const preloadImage = async url => {
+  if (!url) {
+    return;
+  }
+
+  const element = window.document.createElement('img');
+  element.style.display = 'none';
+  element.src = url;
+
+  await new Promise(resolve => {
+    element.onloadeddata = () => {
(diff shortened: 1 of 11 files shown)


adrianaluchian:
Found an issue though, with interactive media plugin. Fixing..

src/utils/image-utils.js
```diff
@@ -0,0 +1,26 @@
...
+    return;
+  }
+
+  const element = window.document.createElement('img');
+  element.style.display = 'none';
+  element.src = url;
+
+  await new Promise(resolve => {
+    element.onloadeddata = () => {
+      ReactDOM.unmountComponentAtNode(element);
```
ahelmberger, line 14: I wonder why this is necessary? There's no React component involved here, just a DOM node (outside the React tree), so `.remove()` should be enough :thinking: 

(Also: this function will be renamed in the next React version)

ahelmberger: Edit: I realized that the element is never put into the document tree, so also `.remove()` is not necessary.

ahelmberger (approved):

PR #608, opened 2022-09-08
EDU-799 Fix unexecuted migrations

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-799

This PR does two things:
* It makes the original migration script from https://educandu.atlassian.net/browse/EDU-770 execute completely
* It incorporates the no-ticket-changes from this commit: https://github.com/educandu/educandu/commit/b4bbe0302260c7c4cc74e4c3d8b241c08acd6684
* It adds a task processor to check the validity of all sections in the DB from the admin UI

**Sorry for the confusion. As these scripts were not run anywhere yet, I thought it would be easier to bring the code base to a clean state than to write yet another correction script** :disappointed:


src/services/document-service.js
```diff
@@ -447,6 +447,32 @@ class DocumentService {
...
 
+  async validateDocument(documentId) {
+    const doc = await this.documentStore.getDocumentById(documentId);
+    for (const section of doc.sections) {
+      try {
+        validateSection(section, this.pluginRegistry);
+      } catch (error) {
+        error.documentId = doc._id;
+        error.sectionKey = section.key;
+        throw error;
```
adrianaluchian, line 458: What would you say about collecting all validation errors on the document and revision and throwing an error at the end with the full report?
Maybe something like `validationError.cases.push({ documentId/documentRevisionId, sectionKey, error })`?
Otherwise the validation process has to be repeatedly run until all errors are fixed.

adrianaluchian (approved):

PR #610, opened 2022-09-09
EDU-793 Transform rooms domain model to keep list of included documents

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-793

### DIFF
--- a/src/services/document-service.js
+++ b/src/services/document-service.js
@@ -148,13 +148,17 @@ class DocumentService {
   }
 
   async createDocument({ data, user }) {
-    let lock;
+    let roomLock;
+    let documentLock;
     const documentId = uniqueId.create();
 
     await this.createUploadDirectoryMarkerForDocument(documentId);
 
     try {
-      lock = await this.lockStore.takeDocumentLock(documentId);
+      documentLock = await this.lockStore.takeDocumentLock(documentId);
+      if (data.roomId) {
+        roomLock = await this.lockStore.takeRoomLock(data.roomId);
+      }
 
       let newDocument;
       await this.transactionRunner.run(async session => {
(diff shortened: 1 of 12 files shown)


adrianaluchian:
One thing missing: there are no tests for `hardDeleteDocument`, so I will include them soon in this PR.

src/services/document-service.js
```diff
@@ -175,6 +176,12 @@ class DocumentService {
 
         await this.documentRevisionStore.saveDocumentRevision(newRevision, { session });
         await this.documentStore.saveDocument(newDocument, { session });
+
+        if (newDocument.roomId) {
```
ahelmberger, line 180: If the `roomId` is set, we should have also taken the room lock at the beginning of this `try` block, so no other concurrent room updates can take place.

src/services/document-service.js
```diff
@@ -255,9 +262,17 @@ class DocumentService {
     try {
       lock = await this.lockStore.takeDocumentLock(documentId);
 
-      logger.info(`Hard deleting external document '${documentId}'`);
+      logger.info(`Hard deleting document '${documentId}'`);
 
       await this.transactionRunner.run(async session => {
+        const doc = await this.documentStore.getDocumentById(documentId, { session });
+
+        if (doc.roomId) {
```
ahelmberger, line 270: I'd also lock the room for this operation

src/services/document-service.js
adrianaluchian: Good point, thanks! Took the lock.

src/services/document-service.js
adrianaluchian: Done here too.
And changed the logic for updating rooms as we discussed 🙏

ahelmberger: Nice! Looks good, thanks!

ahelmberger (approved):

PR #612, opened 2022-09-09
EDU-803 Avoid retrying failed tasks that will never succeed

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-803

### DIFF
--- a/src/services/task-processor.spec.js
+++ b/src/services/task-processor.spec.js
@@ -7,6 +7,7 @@ import { TASK_TYPE } from '../domain/constants.js';
 import ServerConfig from '../bootstrap/server-config.js';
 import DocumentImportTaskProcessor from './document-import-task-processor.js';
 import { destroyTestEnvironment, pruneTestEnvironment, setupTestEnvironment } from '../test-helper.js';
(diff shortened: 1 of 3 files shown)


src/services/task-processor.spec.js
```diff
@@ -156,6 +160,55 @@ describe('task-processor', () => {
...
+      it('should call lockStore.takeTaskLock', () => {
+        sinon.assert.calledOnce(lockStore.takeTaskLock);
+      });
+
+      it('should call taskStore.getUnprocessedTaskById', () => {
+        sinon.assert.calledWith(taskStore.getUnprocessedTaskById, taskId);
+      });
+
+      it('should call documentValidationTaskProcessor.process', () => {
+        sinon.assert.calledWith(documentValidationTaskProcessor.process, nextTask, batchParams, ctx);
```
adrianaluchian, line 189: Maybe throw in here also a `calledOnce` assert, even though it is implied. by the lock being called once, I feel this would be clearer?

adrianaluchian (approved):

src/services/task-processor.spec.js
ahelmberger: Done!

PR #614, opened 2022-09-12
EDU-794 Reorder room documents

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-794

Screen Shot 2022-09-12 at 09 36 00

### DIFF
--- a/src/services/room-service.js
+++ b/src/services/room-service.js
@@ -12,9 +12,9 @@ import TransactionRunner from '../stores/transaction-runner.js';
 import { getPathForPrivateRoom } from '../utils/storage-utils.js';
 import RoomInvitationStore from '../stores/room-invitation-store.js';
 import {
+  STORAGE_DIRECTORY_MARKER_NAME,
   INVALID_ROOM_INVITATION_REASON,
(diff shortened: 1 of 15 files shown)


src/services/room-service.js
```diff
@@ -117,6 +117,13 @@ export default class RoomService {
     return updatedRoom;
   }
 
+  async updateRoomDocuments(roomId, { documents }) {
+    await this.roomStore.updateRoomDocuments(roomId, { documents });
+    const updatedRoom = await this.roomStore.getRoomById(roomId);
+
+    return updatedRoom;
+  }
```
ahelmberger, line 125: I know it would add more complexity but I think we should ensure data integrity here by
* Get a room lock
* Fetch the current room
* Check that the new documents are still exactly the same as the old ones, just in different order
* Save the updated room
* Release the lock

adrianaluchian: That's true, we should ensure valid document ids here, not just id-type values.
I included this check, thanks!

ahelmberger: Great, thanks :+1:

PR #617, opened 2022-09-14
EDU-796 Show human readable folder names

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-796

Screen Shot 2022-09-14 at 12 01 33
Screen Shot 2022-09-14 at 12 01 55
Screen Shot 2022-09-14 at 12 01 05
Screen Shot 2022-09-14 at 12 01 17


src/components/resource-picker/storage-location-screens.js
```diff
@@ -26,7 +26,19 @@ const SCREEN = {
   filesUpload: 'files-upload'
 };
 
-function StorageLocationScreens({ storageLocation, initialUrl, onSelect, onCancel }) {
+const updateFileDisplayName = (cdnObject, documentsTitles, t) => {
+  if (cdnObject.type === CDN_OBJECT_TYPE.directory) {
+    const documentTitle = documentsTitles.find(doc => doc._id === cdnObject.displayName);
+    cdnObject.displayName = `${documentTitle?.title || t('common:unknown')} [${cdnObject.displayName}]`;
```
ahelmberger, line 32: This is called every time `t` changes (e.g. the user switching the display language). But the `displayName` property is already patched after the first call, so the condition in line 31 will never be true again). This is a theoretical problem only, as currently the file dialog has to be closed in order to switch the display language, and after reopening it  there will be another call with a freshly fetched file list. This also plays a little bit into the general feeling I have about doing the human-readable-name patching on the client side (see also my - yet to be written - comment at the end of the PR).

ahelmberger:
I found a few issues (not necessarily all of them directly connected to the PR but more to the feature in general) we have to make up our mind about:

* I kinda forgot what we decided about search, but if we want to search also inside human-readable directory names, it'll get more complicated when patching `displayName` of storage objects on the client side
* When I create a private document, the system also creates a public folder with the ID of that document, so far so good. Now with the human-readable directory names, the title of the private document seems to be disclosed to the public, don't know if this might be a problem, but I think rather yes then no. Actually if people open the dev console and check the request, they get a list of all document titles including all private documents of the entire DB.
* The grid view shows just a small portion of the directory name, so it is hard to find a "document" as it is also not ordered alphabetically.
* We still have the term "folder" in one of the English translations (opposed to the otherwise mostly used "directory"), but this was already in there before the PR. It would be a nice chance to change this one, too.

The TL;DR of this comment is: Maybe we want to think about extending the storage object by a property `cdnObject.humanReadableName` or `cdnObject.alternativeName` etc.) in order to make it an official part of the data structure,  and maybe we want to do this on the server-side in order to also make search possible/more logical. Also we have to decide if we want to really show other peoples' private document titles to the public.

adrianaluchian:
> I found a few issues (not necessarily all of them directly connected to the PR but more to the feature in general) we have to make up our mind about: [...]
>

All very good points, thanks for giving this more thought. 

Mentioning search, I think you're right that it would be better to actually do the CDN objects enrichment on the server side and likely as a separate property on the object. I briefly also considered if to separate the two (`displayName` and the human readable name) but thought maybe it just adds complexity. But the `t` change through language switch would indeed introduce/reveal a bug so we definitely need to fix this and separate the properties.

About whether to show public folders of private rooms, I wasn't sure with that one either, however the CDN does have publicly accessible resources for private documents and I do think they have to be accessible in browsing the CDN as well. If you as an author upload to public storage, then basically you only protect your document content through having it in a room, not your resources. If public users see IDs and title of private documents, they still cannot see the protected content of those documents because they cannot access those pages due to the authorization check.

Indeed, maybe we should sort the documents alphabetically on the server side, by the new `humanReadableName` so that by default the sorting initially matches between grid and list. Now the documents are in the same order in both grid and list view, but that order is not alphabetical.

I went with "folder" for consistency, yes. I don't know in English what people use more often, I for one am very familiar with "folder" more so than with "directory". But either is fine for me if we're consistent.

adrianaluchian:
> > I found a few issues (not necessarily all of them directly connected to the PR but more to the feature in general) we have to make up our mind about: [...]
> 
> All very good points, thanks for giving this more thought.
> 
I made the changes as we agreed. Please check if all looks good to you.

adrianaluchian:
> > > I found a few issues (not necessarily all of them directly connected to the PR but more to the feature in general) we have to make up our mind about: [...]
> > 
> > 
> > All very good points, thanks for giving this more thought.
> 
> I made the changes as we agreed. Please check if all looks good to you.

Made corrections to the accessibility determination logic and cleared title from payload when document is not owned/collaborated on.

ahelmberger (approved):

ahelmberger:
> > > > I found a few issues (not necessarily all of them directly connected to the PR but more to the feature in general) we have to make up our mind about: [...]
> > > 
> > > 
> > > All very good points, thanks for giving this more thought.
> > 
> > 
> > I made the changes as we agreed. Please check if all looks good to you.
> 
> Made corrections to the accessibility determination logic and cleared title from payload when document is not owned/collaborated on.

Nice! This feature is a really great improvement to the whole storage experience :+1:

adrianaluchian:
> > > > > I found a few issues (not necessarily all of them directly connected to the PR but more to the feature in general) we have to make up our mind about: [...]
> > > > 
> > > > 
> > > > All very good points, thanks for giving this more thought.
> > > 
> > > 
> > > I made the changes as we agreed. Please check if all looks good to you.
> > 
> > 
> > Made corrections to the accessibility determination logic and cleared title from payload when document is not owned/collaborated on.
> 
> Nice! This feature is a really great improvement to the whole storage experience 👍

Indeed it is! Thanks for helping with doing it the right way 🙏

PR #620, opened 2022-09-19
EDU-797 Search human readable directory names

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-797

### DIFF
--- a/src/components/debounced-input.js
+++ b/src/components/debounced-input.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useState } from 'react';
 import { useDebouncedCallback } from '../ui/hooks.js';
 
-function DebouncedInput({ elementType: Component, value, onChange, ...props }) {
+function DebouncedInput({ elementType: Component, value, onChange, onEnter, onSearch, ...props }) {
   const [actualValue, setActualValue] = useState(value);
   const debouncedOnChange = useDebouncedCallback(onChange, 250);
 
@@ -16,18 +16,32 @@ function DebouncedInput({ elementType: Component, value, onChange, ...props }) {
     debouncedOnChange(event.target.value);
   };
 
-  return <Component {...props} value={actualValue} onChange={handleChange} />;
+  const componentProps = { ...props };
+
+  if (onEnter) {
+    componentProps.onEnter = () => onEnter(actualValue);
+  }
+
+  if (onSearch) {
+    componentProps.onSearch = () => onSearch(actualValue);
+  }
+
+  return <Component {...componentProps} value={actualValue} onChange={handleChange} />;
 }
 
 DebouncedInput.propTypes = {
   elementType: PropTypes.elementType,
   onChange: PropTypes.func,
+  onEnter: PropTypes.func,
+  onSearch: PropTypes.func,
   value: PropTypes.string
 };
 
 DebouncedInput.defaultProps = {
   elementType: Input,
   onChange: () => {},
+  onEnter: null,
+  onSearch: null,
   value: ''
 };
 
--- a/src/components/resource-picker/storage-location-screens.js
+++ b/src/components/resource-picker/storage-location-screens.js
@@ -34,7 +34,6 @@ function StorageLocationScreens({ storageLocation, initialUrl, onSelect, onCance
 
   const isMounted = useRef(false);
   const [files, setFiles] = useState([]);
-  const [searchTerm, setSearchTerm] = useState('');
   const [isLoading, setIsLoading] = useState(false);
   const [uploadQueue, setUploadQueue] = useState([]);
   const [searchResult, setSearchResult] = useState([]);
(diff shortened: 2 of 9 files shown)


src/components/resource-picker/storage-location-screens.js
```diff
@@ -94,17 +87,11 @@ function StorageLocationScreens({ storageLocation, initialUrl, onSelect, onCance
...
   };
 
   const handleFileDoubleClick = newFile => {
-    if (newFile.type === CDN_OBJECT_TYPE.file) {
-      onSelect(newFile.portableUrl);
-    }
-    if (newFile.type === CDN_OBJECT_TYPE.directory) {
-      setIsLoading(true);
-      setCurrentDirectoryPath(newFile.path);
-    }
```
adrianaluchian, line 107: This was some leftover code that was actually never reached/executed.

src/services/storage-service.js
```diff
@@ -196,16 +196,36 @@ export default class StorageService {
...
+        includeEmptyObjects: false,
+        ignoreNonExistingPath: false
+      });
+
+      const matchingAllLevelsObjects = [
+        ...currentLevelObjects.filter(obj => (
+          obj.documentMetadata?.title.toLowerCase().includes(searchTerm.toLowerCase()) && obj.documentMetadata.isAccessibleToUser)
+          || obj.displayName.toLowerCase().includes(searchTerm.toLowerCase())),
+        ...innerLevelsObjects.filter(obj => obj.displayName.toLowerCase().includes(searchTerm.toLowerCase()))
+      ];
```
ahelmberger, line 216: It took me a while to completely grasp the conditions here, maybe you could extraxct some filter functions, like:

~~~js
const docTitleMatchesSearchTerm = obj => obj.documentMetadata?.title.toLowerCase().includes(searchTerm.toLowerCase()) && obj.documentMetadata.isAccessibleToUser;
const displayNameMatchesSearchTerm = obj => obj.displayName.toLowerCase().includes(searchTerm.toLowerCase());

const matchingAllLevelsObjects = [
  ...currentLevelObjects.filter(obj => docTitleMatchesSearchTerm(obj) || displayNameMatchesSearchTerm(obj)),
  ...innerLevelsObjects.filter(obj => displayNameMatchesSearchTerm(obj))
];
~~~

src/components/debounced-input.js
```diff
@@ -16,18 +16,32 @@ function DebouncedInput({ elementType: Component, value, onChange, ...props }) {
...
+
+  if (onEnter) {
+    componentProps.onEnter = () => onEnter(actualValue);
+  }
+
+  if (onSearch) {
+    componentProps.onSearch = () => onSearch(actualValue);
+  }
+
+  return <Component {...componentProps} value={actualValue} onChange={handleChange} />;
```
ahelmberger, line 29: Although this works in our scenarios, we should keep in mind that it's not a "multi-purpose" implementation, meaning that if it's used as a controlled component, the `value` value that's kept in state outside has still the old value and might be used. The expected behavior would be - in the event the `Component` fires `onEnter` or `onSearch` - to first cancel the debounce and trigger a regular `onChange`, so the outside world is informed and `value` (as seen from the outside) is updated, before triggering `onEnter` or `onSearch`. But that would either require us to add "force-updating" to the debounce hook or manually trigger the `onChange` callback first in this component (maybe in combination with `useDedupedCallback` so we don't call it again with the same value when the actual debounced value change kicks in), before any `onEnter` or `onSearch`. What do you think about leaving it for the purpose of this ticket and having an extra ticket for this? Even though it might not impact the way we use the component right now, we would prevent some possible surprises in the future.

adrianaluchian: That is true, that having `onSearch`/`onEnter` carry the updated value breaks the flow of using `onChange` for this purpose. Indeed we should first call `onChange`.
I can revert this change and make a ticket for this purpose then.
Thanks for pointing this out!

src/components/resource-picker/storage-location-screens.js
ahelmberger: :+1:

src/components/debounced-input.js
ahelmberger: You could also leave it as a temporary fix if we're gonna change it soon anyways, it's up to you. Extra ticket sounds good to me.

src/services/storage-service.js
adrianaluchian: Yes that's indeed much better, I had looked at it too often so that it made sense to me.
Changed!

src/components/debounced-input.js
adrianaluchian: Ok, will not revert then so that we at least have the fix for now.
Will describe the changes needed in the follow-up ticket https://educandu.atlassian.net/browse/EDU-811

ahelmberger: :+1:

ahelmberger (approved):

PR #621, opened 2022-09-19
EDU-795 Allow cloning of documents across rooms and public area

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-795


src/server/document-controller.js
```diff
@@ -74,19 +74,24 @@ class DocumentController {
       if (!templateDocument) {
         throw new NotFound();
       }
+
+      const templateDocumentRoom = templateDocument.roomId ? await this.roomService.getRoomById(templateDocument.roomId) : null;
+      if (templateDocumentRoom && !isRoomOwnerOrMember({ room: templateDocumentRoom, userId: user._id })) {
```
adrianaluchian, line 79: We should check for `isRoomOwnerOrCollaborator` here rather than `isRoomOwnerOrMember`, right?

src/components/document-metadata-modal.js
```diff
@@ -25,9 +29,121 @@ const logger = new Logger(import.meta.url);
...
+      return t('common:clone');
+    case DOCUMENT_METADATA_MODAL_MODE.create:
+      return t('common:create');
+    case DOCUMENT_METADATA_MODAL_MODE.update:
+      return t('common:save');
+    default:
+      throw new Error(`Invalid document metadata modal mode: '${mode}'`);
+  }
+}
+
```
adrianaluchian, line 146: Maybe we could move all these helper functions to a `document-metadata-modal-utils.js` module? It's alot of code in here.

adrianaluchian:
Super nice solution how you integrated the cloning options, very seamlessly 😍 
And it works very smoothly, this feature gives alot of flexibily!

adrianaluchian:
Noticed one difference from how it was before, in document creation. The `useTemplateDocument` checkbox is checked by default. Not sure if this was intentional. My feeling though is that it would be better not to check it?
In either case, since we're here, do you think we could throw in an info icon like in the `generateSequence` checkbox, explaining what this checkbox means?

src/server/document-controller.js
ahelmberger: I was thinking long about that, but then decided not to, because then we would have to also change line 88 further down to:

~~~
if (room && view === 'edit' && !isRoomOwnerOrCollaborator({ room, userId: user._id })) {
  throw new Forbidden();
}

if (room && view !== 'edit' && !isRoomOwnerOrMember({ room, userId: user._id })) {
  throw new Forbidden();
}
~~~

So I kept the paradigm: "If the user is allowed to see it (as oppposed to: edit it), they will see it. If they try to save changes, they will get an error message"

src/components/document-metadata-modal.js
ahelmberger: Yeah, good idea, it was bothering me, too. I'll do that!

src/server/document-controller.js
adrianaluchian: I saw the discrepancy there but I though one case is a guard for viewing ("read") and another for cloning ("write") 🤔 However I agree that the members still have a right so view even with cloning, just not to write, which is not the scope of this controller endpoint. In the end we're talking about a very edge case anyway, so I agree then, let's keep it so.

src/components/document-metadata-modal.js
ahelmberger: Done.

adrianaluchian: Nice! 👍

ahelmberger:
> Super nice solution how you integrated the cloning options, very seamlessly heart_eyes And it works very smoothly, this feature gives alot of flexibily!

:relaxed:

adrianaluchian (approved):

PR #622, opened 2022-09-20
EDU-809 Fix state change on reveal orientation value change

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-809

### DIFF
--- a/src/plugins/image/image-editor.js
+++ b/src/plugins/image/image-editor.js
@@ -183,7 +183,7 @@ function ImageEditor({ content, onContentChanged }) {
 
   const handleOrientationValueChanged = event => {
     const { value } = event.target;
-    const newEffect = { effect, orientation: value };
+    const newEffect = { ...effect, orientation: value };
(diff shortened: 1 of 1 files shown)


src/plugins/image/image-editor.js
```diff
@@ -183,7 +183,7 @@ function ImageEditor({ content, onContentChanged }) {
 
   const handleOrientationValueChanged = event => {
     const { value } = event.target;
-    const newEffect = { effect, orientation: value };
+    const newEffect = { ...effect, orientation: value };
```
adrianaluchian, line 186: Five month old bug, mind you...

ahelmberger: Phew :open_mouth: At least now you can't save that anymore due to validation.

ahelmberger (approved):

src/plugins/image/image-editor.js
adrianaluchian: True, good point!

PR #627, opened 2022-09-21
EDU-811 Add flush function to debounced callback hook

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-811

### DIFF
--- a/src/ui/hooks.js
+++ b/src/ui/hooks.js
@@ -28,6 +28,16 @@ export function useReloadPersistedWindow() {
   }, []);
 }
 
+function cancelTimeout(timeout) {
+  timeout.current && clearTimeout(timeout.current);
+  timeout.current = null;
+}
+
+function flushPendingArgs(pendingArgs, callback) {
+  pendingArgs.current && callback.current(...pendingArgs.current);
+  pendingArgs.current = null;
+}
+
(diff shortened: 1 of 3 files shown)


src/ui/hooks.js
```diff
@@ -38,21 +39,33 @@ export function useDebouncedCallback(callback, timeLimit = 250) {
...
+      lastPendingArgs.current = null;
     }, timeLimit);
   }, [timeLimit, storedCallback]);
+
+  if (!debouncedCallback.flush) {
+    debouncedCallback.flush = function flush() {
+      timeout.current && clearTimeout(timeout.current);
+      timeout.current = null;
+      lastPendingArgs.current && storedCallback.current(...lastPendingArgs.current);
+      lastPendingArgs.current = null;
```
adrianaluchian, line 64: To make the code a bit easier to follow and reduce this repetition in all 3 cases, maybe you could extract lines 63-64  into a `executePendingCallback` method. I think the name would help.

adrianaluchian:
Super nice solution! 👏

src/ui/hooks.js
ahelmberger: Good idea, extract them (and also the two lines above) into separate functions.

adrianaluchian: 👍

PR #629, opened 2022-09-22
EDU-806 Use original image sizes in ear training plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-806

This solution doesn't adjust the answer image's size to fit the question image size. We can leave that to the author, IMO.


adrianaluchian:
> This solution doesn't adjust the answer image's size to fit the question image size. We can leave that to the author, IMO.

Sure, now the authors can crop images themselves.
Nice fading effect on image change!

adrianaluchian (approved):

PR #633, opened 2022-09-22
EDU-820 Migrate invalid slugs

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-820

### DIFF
--- a/migrations/educandu-2022-09-22-02-fix-invalid-slugs.js
+++ b/migrations/educandu-2022-09-22-02-fix-invalid-slugs.js
@@ -1,0 +1,38 @@
+/* eslint-disable camelcase, no-console, no-await-in-loop */
+
+import slugify from '@sindresorhus/slugify';
+
+export default class Educandu_2022_09_22_02_fix_invalid_slugs {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async processCollection(collectionName) {
+    const invalidSlugCondition = {
+      $and: [
+        { slug: { $ne: '' } },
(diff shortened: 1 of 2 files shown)


migrations/educandu-2022-09-22-02-fix-invalid-slugs.js
```diff
@@ -0,0 +1,38 @@
...
+        { slug: { $ne: '' } },
+        { slug: { $not: { $regex: /^[a-z0-9-]+(\/[a-z0-9-]+)*$/ } } }
+      ]
+    };
+
+    const docsToUpdate = await this.db.collection(collectionName).find(invalidSlugCondition).toArray();
+
+    for (const doc of docsToUpdate) {
+      doc.slug = doc.slug.split('/').map(part => slugify(part)).filter(x => x).join('/');
+      await this.db.collection(collectionName).replaceOne({ _id: doc._id }, doc);
```
adrianaluchian, line 22: Maybe it would be good to log something like `Updated ${oldSlug} to ${doc.slug} for document ${doc._id}`;

ahelmberger: I ran it with a copy of prod data, and there were only 2 documents with invalid slugs (lots of older revisions though), so I decided to also extended the validation batch in this ticket to validate not only not only the section contents but the whole document/revision structure including the slug. This way we could just simply mark the ticket as "nontestable" and run the batch once after deploying in order to verify. What do you think?

adrianaluchian: True, basically the test now is the reimport. Ok then 👍

adrianaluchian (approved):

PR #640, opened 2022-09-26
EDU-818 Always use default quality metadata unless updating

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-818

### DIFF
--- a/src/components/document-metadata-modal.js
+++ b/src/components/document-metadata-modal.js
@@ -59,7 +59,7 @@ function DocumentMetadataModal({
   const [isSaving, setIsSaving] = useState(false);
   const [availableRooms, setAvailableRooms] = useState([]);
   const [isLoadingRooms, setIsLoadingRooms] = useState(false);
-  const [tagOptions, setTagOptions] = useState(composeTagOptions(initialDocumentMetadata?.tags));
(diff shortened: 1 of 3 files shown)


src/components/document-metadata-modal.js
```diff
@@ -138,25 +148,32 @@ function DocumentMetadataModal({
 
   const handleCancel = () => onClose();
 
+  const handleValuesChange = (_, { cloningStrategy, cloningTargetRoomId }) => {
+    const documentRoomId = determineDocumentRoomId({ mode, initialDocumentMetadata, documentToClone, cloningStrategy, cloningTargetRoomId });
+    if (documentRoomId) {
+      formRef.current.setFieldsValue({ allowedOpenContribution: DOCUMENT_ALLOWED_OPEN_CONTRIBUTION.metadataAndContent });
+    }
```
adrianaluchian, line 155: Good idea to reset this value!
I think the case from`renderAllowedOpenContributionFormItem` (`if (documentRoomId || noTargetRoomSelectedYet)`) applies here too though, we'd need to reset also if `noTargetRoomSelectedYet`.
But I guess we can't avoid this duplication and have the form reset as part of the `renderAllowedOpenContributionFormItem`..?

adrianaluchian (approved):

PR #641, opened 2022-09-26
EDU-826 Correct logic for determining if doc editing is allowed and redirect if not allowed

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-826

### DIFF
--- a/src/server/document-controller.js
+++ b/src/server/document-controller.js
@@ -107,7 +107,7 @@ class DocumentController {
     }
 
     if (view === DOC_VIEW_QUERY_PARAM.edit && !canEditDocContent({ user, doc, room })) {
-      return res.redirect(routes.getDocUrl({ id: doc._id, slug: doc.slug, templateDocumentId }));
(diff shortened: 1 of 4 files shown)


src/server/document-controller.js
```diff
@@ -105,6 +106,10 @@ class DocumentController {
       room = null;
     }
 
+    if (view === DOC_VIEW_QUERY_PARAM.edit && !canEditDocContent({ user, doc, room })) {
+      return res.redirect(routes.getDocUrl({ id: doc._id, slug: doc.slug, templateDocumentId }));
```
ahelmberger, line 110: As the `view` param is not (and cannot be) "edit" here, we should also ditch any possible `templateDocumentId` from the URL here, I guess.

adrianaluchian: I wanted to keep everything but the `view` param, but sure, I can also simplify.

adrianaluchian: Done!

ahelmberger: Cool, thanks!

ahelmberger (approved):

PR #645, opened 2022-09-27
EDU-830 Restore volume with unmutting and support solo

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-830

Screen Shot 2022-09-27 at 17 46 15
Screen Shot 2022-09-27 at 17 46 28

### DIFF
--- a/src/components/media-player/media-volume-slider.js
+++ b/src/components/media-player/media-volume-slider.js
@@ -10,16 +10,16 @@ export const MEDIA_VOLUME_SLIDER_ORIENTATION = {
 };
 
 function MediaVolumeSlider({ value, onChange, orientation }) {
-  const [lastValueBeforeMutting, setLastValueBeforeMutting] = useState(value);
+  const [lastValueBeforeMuting, setLastValueBeforeMuting] = useState(value);
 
   useEffect(() => {
     if (value > 0) {
-      setLastValueBeforeMutting(value);
+      setLastValueBeforeMuting(value);
     }
   }, [value]);
 
   const handleVolumeButtonClick = () => {
-    onChange(value ? 0 : lastValueBeforeMutting);
+    onChange(value ? 0 : lastValueBeforeMuting);
   };
 
   const handleSliderChange = sliderValue => {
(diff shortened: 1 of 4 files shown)


src/components/media-player/media-volume-slider.js
```diff
@@ -10,12 +10,21 @@ export const MEDIA_VOLUME_SLIDER_ORIENTATION = {
 };
 
 function MediaVolumeSlider({ value, onChange, orientation }) {
+  const [lastValueBeforeMutting, setLastValueBeforeMutting] = useState(value);
```
ahelmberger, line 13: Typo: `Muting`

ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-830
> 
> Screen Shot 2022-09-27 at 17 46 15
> 
> Screen Shot 2022-09-27 at 17 46 28

It would be cool if we had a speaker logo with an "S" in it (in most applications the "Mute" and "Solo" buttons are just labeled "M" and "S"). Maybe we can ask Mani to create one and exchange it then.

src/components/media-player/media-volume-slider.js
adrianaluchian: I appreciate you calling it a typo, given that I intentionally wrote it like that thinking it is correct 😶 
Corrected, thanks!

adrianaluchian:
> 
> It would be cool if we had a speaker logo with an "S" in it (in most applications the "Mute" and "Solo" buttons are just labeled "M" and "S"). Maybe we can ask Mani to create one and exchange it then.

After failing my attempt to create a dedicated icon, I went for the colored variation for now, but I agree we should get a proper icon. "M" would only make sense in combination with "S", but in most use cases of the volume slider there is no solo. 
"S" though makes sense on its own I'd say 👍

src/components/media-player/media-volume-slider.js
ahelmberger: :smile:

ahelmberger (approved):

PR #646, opened 2022-09-28
EDU-841 Improve media analysis display UI

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-841

**With track set:**
Screen Shot 2022-09-28 at 14 52 02
Screen Shot 2022-09-28 at 14 52 18
Screen Shot 2022-09-28 at 14 54 57

**Without track set:**
Screen Shot 2022-09-28 at 15 00 43


adrianaluchian:
I've set a double-color border between the segments, to always have contrast regardless of the segments colors. I think this is nicer than having a border all around the segments (which could in some case be the same color as the segment anyway).
What do you think?

ahelmberger:
> I've set a double-color border between the segments, to always have contrast regardless of the segments colors. I think this is nicer than having a border all around the segments (which could in some case be the same color as the segment anyway). What do you think?

Looks nice, also the more subtle border between the annotation segments is nicer than the solid border we had before :heart_eyes:

ahelmberger (approved):

adrianaluchian:
> > I've set a double-color border between the segments, to always have contrast regardless of the segments colors. I think this is nicer than having a border all around the segments (which could in some case be the same color as the segment anyway). What do you think?
> 
> Looks nice, also the more subtle border between the annotation segments is nicer than the solid border we had before 😍

Cool, thanks! ❤️

PR #647, opened 2022-09-30
EDU-831 Preload media tracks coming from the internal CDN

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-831

### DIFF
--- a/src/components/media-player/media-player-track-group.js
+++ b/src/components/media-player/media-player-track-group.js
@@ -1,8 +1,13 @@
 import PropTypes from 'prop-types';
 import deepEqual from 'fast-deep-equal';
+import { useService } from '../container-context.js';
 import MediaPlayerTrack from './media-player-track.js';
+import ClientConfig from '../../bootstrap/client-config.js';
+import { getResourceType } from '../../utils/resource-utils.js';
+import { getMediaSourceType } from '../../utils/media-utils.js';
+import PreloadingMediaPlayerTrack from './preloading-media-player-track.js';
 import React, { Fragment, useCallback, useEffect, useRef, useState } from 'react';
-import { MEDIA_ASPECT_RATIO, MEDIA_PLAY_STATE, MEDIA_SCREEN_MODE } from '../../domain/constants.js';
+import { MEDIA_ASPECT_RATIO, MEDIA_PLAY_STATE, MEDIA_SCREEN_MODE, MEDIA_SOURCE_TYPE, RESOURCE_TYPE } from '../../domain/constants.js';
 
 const createInitialTrackStates = sources => ({
   mainTrack: {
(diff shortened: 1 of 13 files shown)


src/components/media-player/media-player-track-group.js
```diff
@@ -230,9 +240,16 @@ function MediaPlayerTrackGroup({
     }));
   };
 
-  return (
-    <Fragment>
-      <MediaPlayerTrack
+  const getTrackComponent = sourceUrl => {
+    const isAudioFile = getResourceType(sourceUrl) === RESOURCE_TYPE.audio;
+    const isInternalFile = getMediaSourceType({ sourceUrl, cdnRootUrl }) === MEDIA_SOURCE_TYPE.internal;
+    return isMultitrack && isInternalFile && isAudioFile ? PreloadingMediaPlayerTrack : MediaPlayerTrack;
```
adrianaluchian, line 246: So using the `PreloadingMediaPlayerTrack` would not work for video files in the end then, right?

src/components/media-player/media-player-track-group.js
```diff
@@ -244,26 +261,37 @@ function MediaPlayerTrackGroup({
...
+    const TrackComponent = getTrackComponent(track.sourceUrl);
+    return (
+      <TrackComponent
+        key={`${track.sourceUrl}|${index.toString()}`}
+        trackRef={trackRefs.current.secondaryTracks[index]}
+        volume={volume * track.volume}
+        sourceUrl={track.sourceUrl}
+        screenMode={MEDIA_SCREEN_MODE.none}
+        playbackRate={playbackRate}
+        loadImmediately={isMultitrack || loadImmediately}
```
adrianaluchian, line 283: `isMultitrack` can be dropped from here to simplify the condition. It's always multitrack when a secondary track is rendered.

src/components/media-player/media-player-track-group.js
ahelmberger: It would, but we might end up preloading huge files into memory (with audio it's usually just MP3 with a few megabytes).

src/components/media-player/media-player-track-group.js
ahelmberger: Hm, that's true. So can the `loadImmediately`. Both are dropped and the prop set to `true` now.

src/utils/media-utils.js
```diff
@@ -142,10 +143,16 @@ export function getTrackDurationFromSourceDuration(sourceDuration, playbackRange
...
+export function getMediaSourceType({ sourceUrl, cdnRootUrl }) {
+  if (sourceUrl.startsWith(cdnRootUrl) || sourceUrl.startsWith(CDN_URL_PREFIX)) {
+    return MEDIA_SOURCE_TYPE.internal;
+  }
+
+  if (sourceUrl.startsWith('https://www.youtube.com/')) {
+    return MEDIA_SOURCE_TYPE.youtube;
+  }
+
+  return MEDIA_SOURCE_TYPE.external;
```
adrianaluchian, line 155: Nice! One step closer to dropping the `sourceType`.

src/components/media-player/media-player-track-group.js
adrianaluchian: Indeed. Though big files also carry the bigger risk for the tracks to get out of sync. But let's see what usage feedback comes in.

src/components/media-player/media-player-track-group.js
adrianaluchian: Yes I was just looking that the `PreloadingMediaPlayerTrack` actually throws for `loadImmediately === false` 👍

src/utils/media-utils.js
ahelmberger: :+1: yeah

adrianaluchian (approved):

PR #649, opened 2022-10-06
Create midi piano plugin

benewagner:
Can you please have a look at the code? The plugin still is not meant to be deployed, but it would be great if you can give me some feedback. Thanks in advance!


package.json
```diff
@@ -26,6 +26,9 @@
   "dependencies": {
     "@ant-design/icons": "^4.7.0",
     "@educandu/node-jsx-loader": "^2.2.0",
+    "@fortawesome/fontawesome-svg-core": "^6.2.0",
+    "@fortawesome/free-solid-svg-icons": "^6.2.0",
+    "@fortawesome/react-fontawesome": "^0.2.0",
```
adrianaluchian, line 31: Could you please remove these 3 libraries? As far as I can see, you're not using them anymore, right? 🤔

src/domain/constants.js
```diff
@@ -122,6 +122,11 @@ export const IMAGE_SOURCE_TYPE = {
   external: 'external'
 };
 
+export const MIDI_SOURCE_TYPE = {
```
adrianaluchian, line 125: You could move this into `src/plugins/midi-piano` in a `constants.js` file, as it is only used by this plugin. We do the same for the other plugins which have their own non-shared constants.

src/plugins/midi-piano/midi-piano-info.js
```diff
@@ -0,0 +1,61 @@
...
+import MidiPianoDisplay from './midi-piano-display.js';
+import { QuestionCircleOutlined } from '@ant-design/icons';
+import { MIDI_SOURCE_TYPE } from '../../domain/constants.js';
+
+class MidiPianoInfo {
+
+  static get typeName() { return 'midi-piano'; }
+
+  constructor(gfm) {
+    this.gfm = gfm;
```
adrianaluchian, line 12: `gfm` is not used in here, you could remove it.

src/plugins/midi-piano/midi-piano.yml
```diff
@@ -0,0 +1,16 @@
+name:
+  en: Midi piano
+  de: Midi-Klavier
+midiTrackTitle:
+  en: Track title
+  de: Stück-Titel
```
adrianaluchian, line 6: For cases where we deal with tracks, the translation we use in German is "Spur". Would this match here too?

src/plugins/midi-piano/midi-piano.yml
```diff
@@ -0,0 +1,16 @@
...
+pianoKeyRange:
+  en: Key range
+  de: Tastenumfang
+confirm:
+  en: Confirm
+  de: Bestätigen
```
adrianaluchian, line 12: You could remove the translation for "confirm" from here, as there is a shared translation in `common.yml`. This means that in `src/plugins/midi-piano/midi-piano-editor.js` on line 160, you can replace `t('midiPiano:confirm')` with `t('common:confirm')`

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+
+  const getData = () => {
+    try {
+      axios.get(src, { responseType: 'arraybuffer' })
+        .then(response => {
+          midiData.current = response.data;
+        });
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log(error);
```
adrianaluchian, line 185: Here would be better to use our custom handler and do:
```
  errorHelper.handleApiError({ error, logger, t });
```

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+  const renderControls = () => (
+    <div style={{ display: 'flex', justifyContent: 'center', paddingBottom: '1rem' }}>
+      <Button onClick={startMidiPlayer} icon={<PlayIcon />} />
+      <Button onClick={pauseMidiPlayer} icon={<PauseIcon />} />
+      <Button onClick={stopMidiPlayer} icon={<QuestionCircleOutlined />} />
+    </div>
+  );
+
+  const renderMidiTrackTitle = () => (
+    <div style={{ display: 'flex', justifyContent: 'center' }}>{midiTrackTitle}</div>
```
adrianaluchian, line 198: Let's use CSS classes for styling, instead of inline styling.

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+  const getEventTypefromMidiCommand = (command, velocity) => {
+    switch (command) {
+      case 144:
+        if (velocity > 0) {
+          return 'Note on';
+        }
+        return 'Note off';
+      case 128:
+        return 'Note off';
+      case 176:
```
adrianaluchian, line 43: These numbers look magical, it would be good to extract a constant in the `constants.js` of the plugin, something like:
```
const MIDI_COMMAND = {
    <clearName>: 144,
    <clearName>: 128,
    <clearName>: 176
}
```
This way when you read the code of `getEventTypefromMidiCommand` you can follow it nicely, like you'd read a book, and understand the logic :) 

Btw, small typo in the handler name -> getEventType**F**romMidiCommand

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+import PauseIcon from '../../components/icons/media-player/pause-icon.js';
+
+export default function MidiPianoDisplay({ content }) {
+
+  const keys = useRef([]);
+  const player = useRef(null);
+  const sampler = useRef(null);
+  const midiData = useRef(null);
+  const midiAccessObj = useRef(null);
+  const samplerHasLoaded = useRef(false);
```
adrianaluchian, line 22: `samplerHasLoaded` should be a state variable, so you can use:
`const [samplerHasLoaded, setSamplerHasLoaded] = useState(false)`

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+    switch (command) {
+      case 144:
+        if (velocity > 0) {
+          return 'Note on';
+        }
+        return 'Note off';
+      case 128:
+        return 'Note off';
+      case 176:
+        return 'Sustain off';
```
adrianaluchian, line 44: As far as I can see, this "Sustain off" value is never used, just "Note on" and "Note off", so you could just remove it.
You could extract the 2 values that are actually used into a const (and naming them a bit clearer):
```
const PIANO_EVENT = {
   notePressed: 'note-pressed',
   noteReleased: 'note-released',
   other: 'other'
}
```

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+  function onMIDISuccess(midiAccess) {
+    midiAccessObj.current = midiAccess;
+    for (const input of midiAccessObj.current.inputs.values()) {
+      input.onmidimessage = handleMidiDeviceEvent;
+    }
+  }
+
+  function onMIDIFailure() {
+    // eslint-disable-next-line no-console
+    console.log('Could not access your MIDI devices.');
```
adrianaluchian, line 95: Let's use the `errorHandler.handleError` method here and not `console.log`.
Disabling an ESLint rule should be done as an exception, for the most part we want to respect the linter rules so that the code is nice and streamlined.

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+      <PianoComponent
+        noteRange={noteRange}
+        playNote={playNote}
+        stopNote={stopNote}
+        keys={keys}
+        />
+      <div style={{ paddingTop: '1.5rem' }}>
+        {hasMidiFile
+          && renderControls()}
+        {hasMidiTrackTitle && hasMidiFile
```
adrianaluchian, line 283: Small thing, since these 2 variables are used just here and the expressions they evaluate are very basic, you can reduce the code by inlining the variable expressions:
```
{!!sourceUrl && renderControls()}
{!!sourceUrl && !!midiTrackTitle && renderMidiTrackTitle() }
```
Not that less code is always better, but here it's understandable without the variables :)

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+      sampler.current.triggerRelease(getNoteNameFromMidiValue(midiValue));
+    }, 200);
+  };
+
+  const getData = () => {
+    try {
+      axios.get(src, { responseType: 'arraybuffer' })
+        .then(response => {
+          midiData.current = response.data;
+        });
```
adrianaluchian, line 182: As we have the `HttpClient` wrapping the `axios` usages, let's please use the `HttpClient` for this request and not `axios` directly.
The reason for this is that, if needed, we could easily replace `axios` with a different module with very little code change.

src/utils/url-utils.js
```diff
@@ -63,6 +63,17 @@ function getGravatarUrl(userEmail) {
 
 }
 
+function getMidiUrl({ cdnRootUrl, sourceType, sourceUrl }) {
+  switch (sourceType) {
+    case IMAGE_SOURCE_TYPE.external:
+      return sourceUrl || null;
+    case IMAGE_SOURCE_TYPE.internal:
```
adrianaluchian, line 70: You need to use `MIDI_SOURCE_TYPE` here.

src/domain/constants.js
adrianaluchian: Sorry, actually I see that this should actually be used outside of the plugin folder.
See my comment here: https://github.com/educandu/educandu/pull/649/files#r989035573

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+import { sectionDisplayProps } from '../../ui/default-prop-types.js';
+import PlayIcon from '../../components/icons/media-player/play-icon.js';
+import PauseIcon from '../../components/icons/media-player/pause-icon.js';
+
+export default function MidiPianoDisplay({ content }) {
+
+  const keys = useRef([]);
+  const player = useRef(null);
+  const sampler = useRef(null);
+  const midiData = useRef(null);
```
adrianaluchian, line 20: As far as I can see, this also should be a state object (`const [midiData, setMidiData] = useState(null)`)

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+    if (player.current !== null && player.current.isPlaying()) {
+      return;
+    }
+    if (player.current === null) {
+      instantiatePlayer();
+      player.current.play();
+      return;
+    }
+    player.current.play();
+  };
```
adrianaluchian, line 147: You could rewrite this method in a simpler clearer manner like this:
```
const startMidiPlayer = () => {
    if (player.current === null) {
      instantiatePlayer();
    }
    if (!player.current.isPlaying()) {
       player.current.play();
    }
  };
```

src/plugins/midi-piano/midi-piano-editor.js
```diff
@@ -0,0 +1,197 @@
...
+  );
+
+  const renderExternalSourceTypeInput = (value, onChangeHandler) => (
+    <FormItem label={t('common:externalUrl')} {...formItemLayout} {...validation.validateUrl(value, t)} hasFeedback>
+      <Input value={value} onChange={onChangeHandler} />
+    </FormItem>
+  );
+
+  const renderMidiTrackTitleInput = (value, onChangeHandler) => (
+    <FormItem label={t('midiPiano:midiTrackTitle')} {...formItemLayout} hasFeedback>
```
adrianaluchian, line 125: Because the translation here is instantiated like this `const { t } = useTranslation('midiPiano');` you don't need to specify the "midiPiano" namespace anymore, as it is the current namespace.
This means you can write for example `t('midiTrackTitle')}` and only keep the namespace when it's a different one (like for `t('common:externalUrl')` for example)

src/plugins/midi-piano/piano-component.js
```diff
@@ -0,0 +1,115 @@
+import PropTypes from 'prop-types';
+import { Piano } from 'react-piano';
```
adrianaluchian, line 2: I don't see where "react-piano" comes from, it's not in the `package.json` 🤔

src/plugins/midi-piano/piano-component.js
```diff
@@ -0,0 +1,115 @@
...
+    return function cleanUp() {
+      window.removeEventListener('resize', handleWindowResize);
+    };
+  }, []);
+
+  useEffect(() => {
+    if (renderCount.current !== 2) {
+      return;
+    }
+    const keyElems = document.querySelectorAll('.ReactPiano__Key');
```
adrianaluchian, line 86: We should not use classes top get content like this. Does the piano player not have a way of exposing which keys it contains through props?
I wanted to check the documentation of `react-piano` but I'm not sure which package this is (please see https://github.com/educandu/educandu/pull/649/files#r989082791)

src/plugins/midi-piano/piano-component.js
```diff
@@ -0,0 +1,115 @@
...
+import React, { useEffect, useState, useRef } from 'react';
+
+export default function PianoComponent(props) {
+
+  const defaultKeyWidth = 29.2;
+  const renderCount = useRef(0);
+  const renderPiano = useRef(false);
+  const containerDiv = useRef(null);
+  const containerWidth = useRef(null);
+  const pianoWrapperDiv = useRef(null);
```
adrianaluchian, line 12: I can't find a usage of `pianoWrapperDiv`, can probably be removed.

src/plugins/midi-piano/piano-component.js
```diff
@@ -0,0 +1,115 @@
+import PropTypes from 'prop-types';
+import { Piano } from 'react-piano';
+import React, { useEffect, useState, useRef } from 'react';
+
+export default function PianoComponent(props) {
+
+  const defaultKeyWidth = 29.2;
+  const renderCount = useRef(0);
+  const renderPiano = useRef(false);
```
adrianaluchian, line 9: This should be a state variable and could be called more like a flag (now it's name sounds like it is a method), maybe `const [canRenderPiano, setCanRenderPiano] = useState(false);`

src/plugins/midi-piano/piano-component.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+    if (containerDiv.current.clientWidth > containerWidth.current && neededWidth < containerDiv.current.clientWidth) {
+      return;
+    }
+    const obj = getPianoWrapperDimensions(containerDiv.current.clientWidth);
+    setPianoWrapperDimensions(obj);
+  };
+
+  useEffect(() => {
+    renderCount.current += 1;
```
adrianaluchian, line 62: Hmm, this looks like a workaround to try and delay executing the code from line 86 onward, right? 
Or what was the situation you were trying to get around of by doing this?

ahelmberger:
I checked out the branch and tried to use the plugin, but couldn't get it to work (don't know how much it already supposed to work, though).

* The range selector for the key range looks like it's selecting random keys (never where the user actually clicks)
* After uploading a MIDI file the website crashes

I made a small screen recording of my test:

https://user-images.githubusercontent.com/2676617/194555451-441a58d3-fdc3-4228-a92c-59fec828ffaa.mp4

src/plugins/midi-piano/midi-piano.yml
benewagner: It's the title of the whole piece like "Edvard Grieg: Piano Concerto op. 16, I." Maybe better "Title of piece"? What would you suggest?

src/plugins/midi-piano/midi-piano-display.js
benewagner: Yes, should have commented it out, because it will be used soon, so that when you play the piano with a midi device you can actually use the pedal.

src/plugins/midi-piano/midi-piano-display.js
benewagner: Great!

src/plugins/midi-piano/piano-component.js
benewagner: That's strange, it's in line 97 of my package.json:  "react-piano": "^3.1.3".

src/plugins/midi-piano/piano-component.js
benewagner: It's this package: [https://www.npmjs.com/package/react-piano]

The keys to be shown have to be passed in via the prop `noteRange={ first: midiValueStart, last: midiValueEnd  }` so you would always know which keys are set to be shown. The problem is:

When playing back a midi file or when playing with a midi input device, I'd like to visualize the keys played by changing the color of the plugin-piano keys. The react-piano accepts the native prop `activeKeys` where you can pass in an array with the midi values of the keys to be changed in color. Since this is changing the state, the piano would rerender after every note which is played or released. It does kind of work, but because of the vast amount of rerenders in short time, the playback of the midi file or the playback of the keys you play gets cracks and missing notes and everything. This is why I changed the key visualization to DOM manipulation to get around these rerenders. Now the playback works as it's supposed to.
Is this an allowed thing to do? I can imagine that working with react you should prefer managing the state over of manipulating DOM elements, but this doesn't seem to be the right solution for playing music notes.

src/plugins/midi-piano/midi-piano.yml
adrianaluchian: Ok, I see. So then since there aren't multiple tracks/pieces, I'd go simply just with "Title", like in the other plugins.

src/plugins/midi-piano/midi-piano-display.js
adrianaluchian: I thought this was used 🤔 
I'd recommend not having either unused or commented out code in your pull request and keeping it clean. Any new functionality you want to add later can then be built on top.

src/plugins/midi-piano/piano-component.js
adrianaluchian: I see, it's just in your local code, not part of the code you created the pull request with 😬
Screen Shot 2022-10-10 at 09 41 10

adrianaluchian: Ok, mystery solved, I was not aware that this package is already imported because of this other plugin that we've not touched in a long time 🤦‍♀️ 
Sorry, I overlooked it.

src/plugins/midi-piano/piano-component.js
ahelmberger: There are basically two ways to work with such components: either they're 100% managed by React - or - they are 0% managed by React (meaning it's pure Vanilla DOM manipulation with only a React wrapper `div`). This mix doesn't really work well, because you never know when React will override your manual manipulations of the tree that is under React's control. I would first go the 100% pure React way. If the re-rendering gets too inefficient, there's always the possibility to either fix `react-piano` itself, or, if not possible/feasible, create our own piano component (it's just a bunch of `div` elements anyway, not that much magic).

benewagner: That makes sense! So for going 100% React, it depends on whether the package can be fixed so that the rerenders don't obstruct the audio playback. If this is possible, I must say that it surpasses what I can do. Although it doesn't mean too much, I can't imagine how that could be accomplished. Playing a chord consisting of only three notes on a midi device can be enough to cause problems in the audio playback.
As much as I would like to work with React, I would enjoy making the piano with Vanilla JS from ground up just the same. Maybe it would even be a good way to make sure that the functionality can be extended in the future without the risk of new issues coming up.
What would you suggest as the next steps?

adrianaluchian: My 2 cents would be to try and address the other comments and low hanging fruit to clear up the code and get it to work so we could try it out too (see https://github.com/educandu/educandu/pull/649#issuecomment-1271538587). Then we could think a bit and talk how to best proceed. Or? What do you think, @ahelmberger ?

ahelmberger: I agree. Let's make it run first (using the current approach), so we can get a better picture of what the plugin will do. Then we can decide how to replace the piano (I think a vanilla custom-written version might be the most lightweight while still performant solution).

benewagner: Ok I'll go on and keep you informed on the progress.

src/plugins/midi-piano/midi-piano-display.js
benewagner: This I tried to implement, but I don't know what to set `t` to. Please have a look at what I did, thank you.

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
...
+    return NOTES[midiValue];
+  };
+
+  const getEventTypefromMidiCommand = (command, velocity) => {
+    switch (command) {
+      case 144:
+        if (velocity > 0) {
+          return 'Note on';
+        }
+        return 'Note off';
```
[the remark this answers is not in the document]
benewagner: Oh, seems like I skipped this one by accident. I will include it in the next commit then.

src/plugins/midi-piano/midi-piano-display.js
benewagner: I removed the "Sustain off" and kept the other two named the same just for now, still need to think about it based on how the plugin evolves. Thanks however for the suggestion, definitely open to it!

src/plugins/midi-piano/midi-piano-display.js
benewagner: This I tried to implement, but I don't know what to set `t` to. Please have a look at what I did, thank you.
Also, if I may ask just for me to learn: Why is the logger instantiated outside the function beneath the imports?

src/plugins/midi-piano/midi-piano-editor.js
```diff
@@ -0,0 +1,197 @@
...
+      });
+      validNoteRange = {
+        first: array[0],
+        last: array[1]
+      };
+      changeContent({ activeNotes: array, noteRange: validNoteRange });
+      return;
+    }
+    changeContent({ activeNotes: array, noteRange: validNoteRange });
+  };
```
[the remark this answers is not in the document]
benewagner: I got rid of the key range selector for now, since there were problems with the mouse event handlers of the react-piano. I will remake it once we've decided on how to go on with the plugin. That is why I also got rid of this method for now.

src/plugins/midi-piano/midi-piano-display.js
```diff
@@ -0,0 +1,292 @@
+import axios from 'axios';
+import * as Tone from 'tone';
```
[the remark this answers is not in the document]
benewagner: I got rid of the key range selector for now. The strange behaviour came from the react-piano mouse event handlers. It actually selected the notes you clicked on, but the visualization was not correct.
The crash was due to a bug I missed since I didn't actually save right after using the key selector.

ahelmberger:
As the code on both sides has changed a lot in the meantime, we'll close this PR for now. We can create another one when the topic is ready again.

PR #653, opened 2022-10-06
EDU-840 Import CSV into media analysis plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-840

### DIFF
--- a/src/plugins/media-analysis/media-analysis-editor.js
+++ b/src/plugins/media-analysis/media-analysis-editor.js
@@ -359,7 +359,7 @@ function MediaAnalysisEditor({ content, onContentChanged }) {
               onPartDelete={handleChapterDelete}
               onStartPositionChange={handleChapterStartPositionChange}
               />
-            {chapters.length && (
+            {!!chapters.length && (
             <Fragment>
               <ChapterSelector
                 chaptersCount={chapters.length}
--- a/src/plugins/media-analysis/media-analysis-editor.js
+++ b/src/plugins/media-analysis/media-analysis-editor.js
@@ -1,10 +1,14 @@
 import by from 'thenby';
+import classNames from 'classnames';
+import Logger from '../../common/logger.js';
 import { useTranslation } from 'react-i18next';
 import urlUtils from '../../utils/url-utils.js';
 import { COLOR_SWATCHES } from './constants.js';
 import cloneDeep from '../../utils/clone-deep.js';
+import * as reactDropzoneNs from 'react-dropzone';
 import { Button, Form, Input, Tooltip } from 'antd';
 import ItemPanel from '../../components/item-panel.js';
+import { handleApiError } from '../../ui/error-helper.js';
 import ColorPicker from '../../components/color-picker.js';
 import ClientConfig from '../../bootstrap/client-config.js';
 import { MEDIA_SCREEN_MODE } from '../../domain/constants.js';
@@ -13,7 +17,6 @@ import { formatMediaPosition } from '../../utils/media-utils.js';
 import Timeline from '../../components/media-player/timeline.js';
 import { useService } from '../../components/container-context.js';
 import { sectionEditorProps } from '../../ui/default-prop-types.js';
-import { InfoCircleOutlined, PlusOutlined } from '@ant-design/icons';
 import React, { Fragment, useEffect, useRef, useState } from 'react';
 import { useNumberFormat } from '../../components/locale-context.js';
 import TrackMixer from '../../components/media-player/track-mixer.js';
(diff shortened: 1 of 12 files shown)


src/plugins/media-analysis/media-analysis-utils.js
```diff
@@ -54,6 +65,65 @@ export function createDefaultContent(t) {
...
 }
 
+export function exportChaptersToCsv(chapters) {
+  return [
+    'startPosition,title,color,text',
+    ...chapters.map(chapter => [
+      chapter.startPosition.toString(),
+      `"${chapter.title.replace(/"/g, '""')}"`,
+      `"${chapter.color.replace(/"/g, '""')}"`,
+      `"${chapter.text.replace(/"/g, '""')}"`
```
adrianaluchian, line 75: Maybe it's worth extracting an "escapeText" method here.

src/plugins/media-analysis/media-analysis-utils.js
```diff
@@ -54,6 +65,65 @@ export function createDefaultContent(t) {
...
+    ''
+  ].join('\n');
+}
+
+export function importChaptersFromCsv(csvStringOrFile) {
+  return new Promise((resolve, reject) => {
+    parseCsv(csvStringOrFile, {
+      header: true,
+      skipEmptyLines: true,
+      error: error => reject(Object.assign(new Error(), error)),
```
adrianaluchian, line 86: Is `error` not instance of `Error` then here, to just `reject(error)` directly? 🤔

src/plugins/media-analysis/media-analysis-utils.js
```diff
@@ -54,6 +65,65 @@ export function createDefaultContent(t) {
...
+            const chapter = {
+              key: uniqueId.create(),
+              startPosition: Number(row.startPosition),
+              color: row.color,
+              title: row.title,
+              text: row.text
+            };
+            joi.attempt(chapter, chapterSchema, { abortEarly: false, convert: false, noDefaults: true });
+            if (index === 0 && chapter.startPosition !== 0) {
+              throw new Error('First chapter has to start at position 0');
```
adrianaluchian, line 106: Should we maybe pass `t` in here to throw translated error messages?

src/plugins/media-analysis/media-analysis-editor.js
```diff
@@ -272,61 +333,77 @@ function MediaAnalysisEditor({ content, onContentChanged }) {
...
-                onChange={handleChapterColorChange}
-                />
-            </FormItem>
-            <FormItem label={t('chapterTextLabel')} {...formItemLayout}>
-              <MarkdownInput
-                preview
-                disabled={!selectedChapterFraction}
-                onChange={handleChapterTextChange}
-                value={chapters?.[selectedChapterIndex].text || ''}
+            {chapters.length && (
```
adrianaluchian, line 362: I know this code was here before, but we should cast `chapters.length` to boolean 😬

src/plugins/media-analysis/media-analysis-editor.js
```diff
@@ -165,6 +175,51 @@ function MediaAnalysisEditor({ content, onContentChanged }) {
     changeContent({ volumePresets: newVolumePresets });
   };
 
+  const handleFileDrop = async fs => {
```
adrianaluchian, line 178: It's cool that drag and drop is supported, but it's not visible to the user 🤔 
Should we maybe have a tooltip over the menu option saying that "this can be done via drag and drop too"?

adrianaluchian:
Works super smooth 😎 
Just a small glitch I found, when you export, then import, then try to export again, the download is no longer triggered (destination picker dialog doesn't open).

src/plugins/media-analysis/media-analysis-utils.js
ahelmberger: True. Actually it was pretty straight forward to extract generic `(to|from)Csv` functions, so I created a `csv-utils` module.

src/plugins/media-analysis/media-analysis-utils.js
ahelmberger: Exactly. Unfortunately they are plain objects (that contain - amongst others - a `message` property), so I had to assign those properties to a "real" error instance while not loosing all the error details.

src/plugins/media-analysis/media-analysis-utils.js
ahelmberger: Do you think it's worth it? These are just three out of many things that can go wrong (and have only English error messages coming from the libraries). Also they can never happen as long as users use Anavis to create the CSV. They could only occurr when the file has been meddled with manually in which case I assume a "power user". I didn't want to bloat our translation file just because of this super edge case (probably not used by more than 2-3 people) feature. What do you think?

src/plugins/media-analysis/media-analysis-utils.js
adrianaluchian: Ah, great! 👍

src/plugins/media-analysis/media-analysis-utils.js
adrianaluchian: Ok then, you're right, let's not invest so much in this relatively hidden feature.

src/plugins/media-analysis/media-analysis-editor.js
ahelmberger: True. I added a small icon with tooltip at the lower right corner of the box

ahelmberger:
> Works super smooth sunglasses Just a small glitch I found, when you export, then import, then try to export again, the download is no longer triggered (destination picker dialog doesn't open).

Hm, I cannot reproduce that on my machine. Is it still the case after my changes (and are there any hints of an error)?

src/plugins/media-analysis/media-analysis-editor.js
adrianaluchian: Nice idea!

adrianaluchian:
> > Works super smooth sunglasses Just a small glitch I found, when you export, then import, then try to export again, the download is no longer triggered (destination picker dialog doesn't open).
> 
> Hm, I cannot reproduce that on my machine. Is it still the case after my changes (and are there any hints of an error)?

It still happens, basically with exporting twice in a row (not even importing in between). When you try to export the second time, nothing happens and there are no errors.

adrianaluchian (approved):

ahelmberger:
> > > Works super smooth sunglasses Just a small glitch I found, when you export, then import, then try to export again, the download is no longer triggered (destination picker dialog doesn't open).
> > 
> > 
> > Hm, I cannot reproduce that on my machine. Is it still the case after my changes (and are there any hints of an error)?
> 
> It still happens, basically with exporting twice in a row (not even importing in between). When you try to export the second time, nothing happens and there are no errors.

As agreed upon, the multiple downloads are restricted by some browser settings. There's nothing we can do about it now.

PR #658, opened 2022-10-10
EDU-864 Make text content in media slideshows scroll if too big

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-864

### DIFF
--- a/src/plugins/media-slideshow/media-slideshow.less
+++ b/src/plugins/media-slideshow/media-slideshow.less
@@ -31,10 +31,10 @@
   width: 100%;
   padding: @edu-content-padding;
   background-color: @edu-background-color-light;
-  display: flex;
+  display: grid;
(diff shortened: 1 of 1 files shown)


src/plugins/media-slideshow/media-slideshow.less
```diff
@@ -31,10 +31,10 @@
   width: 100%;
   padding: @edu-content-padding;
   background-color: @edu-background-color-light;
-  display: flex;
+  display: grid;
   align-items: center;
-  justify-content: center;
-  flex-direction: column;
+  justify-items: stretch;
+  overflow-y: auto;
```
adrianaluchian, line 37: Should we maybe allow auto overflow on both axis?

adrianaluchian (approved):

PR #659, opened 2022-10-10
EDU-868 Keep pre-determined media duration on reset

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-865

### DIFF
--- a/src/components/media-player/media-player-track-group.js
+++ b/src/components/media-player/media-player-track-group.js
@@ -29,25 +29,36 @@ const createInitialTrackStates = sources => ({
   }))
 });
 
-const haveMediaTracksChanged = (currentTrackStates, newSources) => {
-  if (
-    newSources.mainTrack.sourceUrl !== currentTrackStates.mainTrack.sourceUrl
-    || newSources.mainTrack.playbackRange[0] !== currentTrackStates.mainTrack.playbackRange[0]
-    || newSources.mainTrack.playbackRange[1] !== currentTrackStates.mainTrack.playbackRange[1]
+const hasMainTrackMediaChanged = (currentTrackStates, sources) => {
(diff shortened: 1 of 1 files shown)


src/components/media-player/media-player-track-group.js
```diff
@@ -29,25 +29,36 @@ const createInitialTrackStates = sources => ({
...
+
+const hasAnyMediaChanged = (currentTrackStates, newSources) => {
+  return hasMainTrackMediaChanged(currentTrackStates, newSources)
     || newSources.secondaryTracks.length !== currentTrackStates.secondaryTracks.length
-  ) {
-    return true;
+    || currentTrackStates.secondaryTracks.some((track, index) => hasSecondaryTrackMediaChanged(currentTrackStates, newSources, index));
+};
+
+const resetTrackStates = (currentTrackStates, sources) => {
```
adrianaluchian, line 48: A bit of help and context on these changes:

When **any** track changes its media data (`sourceUrl` or `playbackRange`), we were resetting the  `trackState`. This meant the `duration` was lost for **all** tracks too. However the `duration` only gets determined again if the `sourceUrl` of a track is changed.

And due to this code
```
    if (!!mainTrackDuration && secondaryTrackDurations.every(x => !!x)) {
      onDuration(mainTrackDuration);
    }
```
we only called the `onDuration` callback when all durations were there.

So basically for the situation in the JIRA ticket, this was never called after the range was changed.

So within the reset of the `trackStates`, I keep the `durations` if the track "media" data is not changed. Everything else makes sense to be reset and kept in sync between the tracks.

ahelmberger: Thanks, this info was very helpful!

ahelmberger (approved):

PR #663, opened 2022-10-11
EDU-874 Fix lack of propagation of range and duration changes in certain cases

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-874

### DIFF
--- a/src/components/media-player/media-player-track-group.js
+++ b/src/components/media-player/media-player-track-group.js
@@ -174,7 +174,16 @@ function MediaPlayerTrackGroup({
     }
   }, [sources, trackStates, reinitialize]);
 
-  const setDurations = (mainTrackDuration, secondaryTrackDurations) => {
+  useEffect(() => {
+    const mainTrackDuration = trackStates.mainTrack.duration;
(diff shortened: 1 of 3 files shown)


src/components/media-player/media-player-track-group.js
```diff
@@ -174,7 +174,16 @@ function MediaPlayerTrackGroup({
     }
   }, [sources, trackStates, reinitialize]);
 
-  const setDurations = (mainTrackDuration, secondaryTrackDurations) => {
+  useEffect(() => {
+    const mainTrackDuration = trackStates.mainTrack.duration;
+    const secondaryTrackDurations = trackStates.secondaryTracks.map(track => track.duration);
+
+    if (!mainTrackDuration || (!!mainTrackDuration && secondaryTrackDurations.every(d => !!d))) {
+      onDuration(mainTrackDuration);
```
adrianaluchian, line 182: I've done this because we also need to trigger `onChange` when the main track's media data gets changed (`sourceUrl` or `range`) thus resetting it's duration to `0`.
This case is reached not via `setDeterminedDurations`, but through  `resetTrackStates`.

ahelmberger (approved):

PR #667, opened 2022-10-12
EDU-869 Ensure tracks do not play on pause in Firefox

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-869

### DIFF
--- a/src/components/media-player/media-player-track.js
+++ b/src/components/media-player/media-player-track.js
@@ -207,6 +207,9 @@ function MediaPlayerTrack({
     { 'is-hidden': screenMode === MEDIA_SCREEN_MODE.overlay }
   );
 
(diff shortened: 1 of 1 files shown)


src/components/media-player/media-player-track.js
```diff
@@ -207,6 +207,9 @@ function MediaPlayerTrack({
     { 'is-hidden': screenMode === MEDIA_SCREEN_MODE.overlay }
   );
 
+  const isBufferingWhilePlaying = trackPlayState.current === MEDIA_PLAY_STATE.buffering && trackPlayState.beforeBuffering === MEDIA_PLAY_STATE.playing;
+  const shouldPlay = trackPlayState.current === MEDIA_PLAY_STATE.playing || isBufferingWhilePlaying;
```
adrianaluchian, line 211: Great find and nice solution!

ahelmberger: Thanks :-) Let's hope we don't have to find much more stuff in the future

PR #670, opened 2022-10-13
EDU-846 Expand markdown-input to include possibility for CDN URL injection

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-846


src/components/resource-picker/resource-picker.js
```diff
@@ -1,17 +1,21 @@
...
 
 // This can later be made configurable by individual plugins
 // (when we have more locations than just from the public/private CDN).
 const allowedLocationTypes = [STORAGE_LOCATION_TYPE.public, STORAGE_LOCATION_TYPE.private];
 
-function ResourcePicker({ url, onUrlChange }) {
+function ResourcePicker({ isOpen, isButtonVisible, url, onUrlChange }) {
   const { t } = useTranslation();
-  const [isModalVisible, setIsModalVisible] = useState(false);
+  const [isModalVisible, setIsModalVisible] = useState(isOpen);
```
ahelmberger, line 14: The fact that we copy the prop into the internal state makes the component "half-controlled", which works fine if the open/close state is always triggered by the same source (either the prop or the internal state). But when it is mixed they get out of sync.

Example: The markdown field opens the dialog by setting `isOpen` to true, then the user clicks on "Cancel" which only sets the internal state `isModalVisible` to `false` while `isOpen` still remains true. The consequence is that the dialog cannot be opened a second time as the outside code still thinks it's open and sets `isOpen` from `true`to `true` which does not trigger the effect hook again as there is no change.

While we could add a callback here of course to correct this I think it makes the component hard to reason about. I'd suggest to extract a `ResourcePickerDialog` from here that is 100% controlled by the outside via `isOpen`, `onUrlChange`, onCancel` and use that in here as well as in the markdown field, what do you think?

ahelmberger:
Looks and works really cool (except for the dialog problem) :heart_eyes:

src/components/resource-picker/resource-picker.js
adrianaluchian: Done!

ahelmberger (approved):

adrianaluchian:
> Looks and works really cool (except for the dialog problem) 😍
🙇‍♀️

PR #672, opened 2022-10-17
EDU-855 Fix crash due to quick state change when typing fast into multitrack sourceUrl fields

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-885

### DIFF
--- a/src/components/media-player/media-duration-cache.js
+++ b/src/components/media-player/media-duration-cache.js
@@ -1,7 +1,7 @@
 import Logger from '../../common/logger.js';
+import { isBrowser } from '../../ui/browser-helper.js';
 import { determineMediaDuration } from '../../utils/media-utils.js';
 import { getDisposalInfo, DISPOSAL_PRIORITY } from '../../common/di.js';
-import { isBrowser } from '../../ui/browser-helper.js';
 
 const logger = new Logger(import.meta.url);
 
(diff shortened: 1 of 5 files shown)


src/utils/media-utils.js
```diff
@@ -68,11 +68,18 @@ export const determineMediaDuration = memoizee(async url => {
...
+    let validUrl;
+    if (!url) {
+      resolve(null);
+    }
+
+    try {
+      validUrl = new URL(url).href;
+    } catch {
+      resolve(null);
+    }
```
adrianaluchian, line 80: Just added this to remove the console errors logged while typing an external URL. Basically I think we should treat a missing value the same as a partial value (aka URL while typing).

src/components/media-player/media-duration-cache.js
```diff
@@ -1,7 +1,7 @@
 import Logger from '../../common/logger.js';
+import { isBrowser } from '../../ui/browser-helper.js';
 import { determineMediaDuration } from '../../utils/media-utils.js';
 import { getDisposalInfo, DISPOSAL_PRIORITY } from '../../common/di.js';
-import { isBrowser } from '../../ui/browser-helper.js';
```
ahelmberger, line 4: :scream:

ahelmberger (approved):

PR #674, opened 2022-10-18
EDU-884 Add url-input component

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-884

### DIFF
--- a/src/components/url-input.js
+++ b/src/components/url-input.js
@@ -13,13 +13,13 @@ import ResourcePicker from './resource-picker/resource-picker.js';
 import { GlobalOutlined, WarningOutlined, YoutubeOutlined } from '@ant-design/icons';
 import { ensurePortableUrlIfStorageUrl, storageLocationPathToUrl } from '../utils/storage-utils.js';
 
-function UrlInput({ sourceUrl, onChange }) {
+function UrlInput({ value, onChange }) {
   const { t } = useTranslation('urlInput');
   const clientConfig = useService(ClientConfig);
 
   const sourceType = useMemo(
-    () => getSourceType({ sourceUrl, cdnRootUrl: clientConfig.cdnRootUrl })
-    , [clientConfig.cdnRootUrl, sourceUrl]
+    () => getSourceType({ sourceUrl: value, cdnRootUrl: clientConfig.cdnRootUrl })
+    , [clientConfig.cdnRootUrl, value]
   );
 
   const inputPrefixIcon = useMemo(() => {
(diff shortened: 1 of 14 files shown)


src/components/url-input.js
```diff
@@ -0,0 +1,90 @@
...
+        return <WarningOutlined />;
+    }
+  }, [sourceType]);
+
+  const handleUrlChange = selectedPortableUrl => {
+    onChange({ sourceUrl: selectedPortableUrl, sourceType });
+  };
+
+  const handleInputValueChange = event => {
+    onChange({ sourceUrl: event.target.value, sourceType });
```
adrianaluchian, line 47: Reconsidered the return value, will return the url only and the type can be identified wherever needed. I overcomplicated it this way.

ahelmberger:
I already love this component, it's plain simple but intelligent :heart:

src/components/url-input.yml
```diff
@@ -0,0 +1,15 @@
...
+  de: Interner öffentlicher Speicher URL
+tooltip_internal-private:
+  en: Internal private storage URL
+  de: Interner privater Speicher URL
+tooltip_external:
+  en: External URL
+  de: Externer URL
+tooltip_unsupported:
+  en: Unsupported value
+  de: Nicht unterstützten Wert
```
ahelmberger, line 15: Maybe better "Nicht unterstützte URL"?

ahelmberger, line 15: Also, although technically, URL should be masculine, it is used by 99,9% percent of people as feminine, meaning you should remove the "r"s in "Interner" and "Externer". And also, it's more correct to connect "URL" with the noun before by a hyphen in German.

adrianaluchian: But it could be that the user did type something abstract like "abc". That's why I didn't want to call it "URL".

adrianaluchian: Indeed, in `common.yml` we write "Externe URL" 🤦‍♀️ 
But there we don't hyphenate either. I'll update in `common.yml` as well then.

adrianaluchian: How should the German version for `tooltip_internal-private` / `tooltip_internal-public` be actually?

ahelmberger: > Indeed, in `common.yml` we write "Externe URL" woman_facepalming But there we don't hyphenate either. I'll update in `common.yml` as well then.

"Externe" is an adjective, so no hyphen there. Only between noun (for example "YouTube") and "URL" there should be a hyphen.

ahelmberger: Hey, another thought: I think it would sound easier in both languages if we drop the URL completely and simply have these tooltips:

~~~
tooltip_youtube:
  en: Youtube
  de: Youtube
tooltip_internal-public:
  en: Internal public storage
  de: Interner öffentlicher Speicher
tooltip_internal-private:
  en: Internal private storage
  de: Interner privater Speicher
tooltip_external:
  en: External
  de: Extern
tooltip_unsupported:
  en: Unsupported
  de: Nicht unterstützt
~~~

Or maybe with the word "Source: " / "Quelle: " in front of each?

What do you think?

adrianaluchian:
> I already love this component, it's plain simple but intelligent ❤️

Thanks! I think it might morph a few times till I get it right, after actually using it in plugins 😄

src/components/url-input.yml
adrianaluchian: Perfect, will drop. Thanks!

adrianaluchian: Done

ahelmberger: Cool!

PR #676, opened 2022-10-19
EDU-888 Use url-input component in audio plugin

adrianaluchian:
**Branched out of:** https://github.com/educandu/educandu/pull/674
Closes https://educandu.atlassian.net/browse/EDU-888

### DIFF
--- a/migrations/educandu-2022-10-18-01-remove-sourceType-from-audio-plugin.js
+++ b/migrations/educandu-2022-10-18-01-remove-sourceType-from-audio-plugin.js
@@ -1,0 +1,37 @@
+/* eslint-disable camelcase, no-await-in-loop, no-console */
+export default class Educandu_2022_10_18_01_remove_sourceType_from_audio_plugin {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async collectionUp(collectionName) {
+    const result = await this.db.collection(collectionName).updateMany(
+      {},
+      {
+        $unset: {
+          'sections.$[sectionElement].content.sourceType': null
+        }
+      },
+      {
+        arrayFilters: [
+          {
+            'sectionElement.type': { $in: ['audio'] },
+            'sectionElement.content': { $ne: null }
+          }
+        ],
+        multi: true
+      }
+    );
+
+    console.log(`Updated ${collectionName}: ${JSON.stringify(result)}`);
+  }
+
+  async up() {
+    await this.collectionUp('documents');
+    await this.collectionUp('documentRevisions');
+  }
+
+  down() {
+    throw Error('Not supported');
+  }
+}
--- a/migrations/educandu-2022-10-18-01-remove-sourceType-from-audio-plugin.js
+++ b/migrations/educandu-2022-10-18-01-remove-sourceType-from-audio-plugin.js
@@ -15,7 +15,7 @@ export default class Educandu_2022_10_18_01_remove_sourceType_from_audio_plugin 
       {
         arrayFilters: [
           {
-            'sectionElement.type': { $in: ['audio'] },
+            'sectionElement.type': 'audio',
             'sectionElement.content': { $ne: null }
           }
         ],
--- a/src/domain/constants.js
+++ b/src/domain/constants.js
@@ -175,6 +175,11 @@ export const DOCUMENT_ALLOWED_OPEN_CONTRIBUTION = {
   none: 'none'
 };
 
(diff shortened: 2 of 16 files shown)


src/utils/source-utils.js
```diff
@@ -0,0 +1,79 @@
...
+    return SOURCE_TYPE.external;
+  }
+
+  return SOURCE_TYPE.unsupported;
+}
+
+export function isInternalSourceType({ url, cdnRootUrl }) {
+  const sourceType = getSourceType({ url, cdnRootUrl });
+  return sourceType === SOURCE_TYPE.internalPrivate || sourceType === SOURCE_TYPE.internalPublic;
+}
```
adrianaluchian, line 79: My idea/plan is to gather here all helpers dealing with formatting the `sourceUrl` for the:
- plugin editors
       - which show the "portable version" in the input (aka `cdn://media/...`)
       - but persist the "persistable version" (aka `media/...`)
- and plugin displays
       - which use the "accessible version" aka `https://cdn.openmusic.academy/media...`)
 
and as we migrate all plugins, remove the other such helpers scattered around in `media-utils` / `url-utils` / `storage-utils`.
These new helpers are now smart enough to cover all cases.

migrations/educandu-2022-10-18-01-remove-sourceType-from-audio-plugin.js
```diff
@@ -0,0 +1,37 @@
...
+  constructor(db) {
+    this.db = db;
+  }
+
+  async collectionUp(collectionName) {
+    const result = await this.db.collection(collectionName).updateMany(
+      {},
+      {
+        $unset: {
+          'sections.$[sectionElement].content.sourceType': null
```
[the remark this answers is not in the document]
adrianaluchian: I see. Here:
"In the former version, the "accessible version" is just a path and not a full URL"
you meant to say "In the former version, the "persistable version" is just a path and not a full URL" though, right?

So you would suggest that we persist the "portable version" from now on? Currently we persist the "persistable version" aka URL (if external) or CDN path.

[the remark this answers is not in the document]
ahelmberger: Yes, sorry, you're right, I meant the "persistable version" is just a path. And yes, I would suggest to also use the "portable version" for persistence from now on, sorry for the confusion. Watcha think?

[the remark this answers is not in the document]
adrianaluchian: I think it will simplify things even more and we should do it! 😁 
Would it be fine for you to do it in a subsequent ticket, for all plugins (as it needs a thorough migration script), **after** we first migrate the plugins to use the new input and drop `sourceType`? I feel it can get convoluted to do both things at the same time. So keeping the way of persisting as it used to be until the first round of changes is applied.

src/utils/source-utils.js
ahelmberger: Yeah, that will clean up the distributed "mess" we have right now :+1:

src/domain/constants.js
```diff
@@ -175,6 +175,11 @@ export const DOCUMENT_ALLOWED_OPEN_CONTRIBUTION = {
   none: 'none'
 };
 
+export const FORM_ITEM_LAYOUT = {
+  labelCol: { span: 4 },
+  wrapperCol: { span: 14 }
+};
+
```
ahelmberger, line 182: Great idea!

src/plugins/plugin-registry.js
```diff
@@ -20,6 +20,7 @@ import MusicXmlViewerInfo from './music-xml-viewer/music-xml-viewer-info.js';
 import IntervalTrainerInfo from './interval-trainer/interval-trainer-info.js';
 import MultitrackMediaInfo from './multitrack-media/multitrack-media-info.js';
 import InteractiveMediaInfo from './interactive-media/interactive-media-info.js';
+import { isBrowser } from '../ui/browser-helper.js';
```
ahelmberger, line 23: ahem *(clearing throat)*

ahelmberger:
It's thrilling to see the amount of code we can get rid of, so nice!

src/plugins/plugin-registry.js
adrianaluchian: 😱 sacré bleu!

ahelmberger (approved):

PR #685, opened 2022-10-21
EDU-849 Allow playback by clicking on a part

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-849

### DIFF
--- a/src/plugins/media-analysis/media-analysis-display.js
+++ b/src/plugins/media-analysis/media-analysis-display.js
@@ -82,12 +82,14 @@ function MediaAnalysisDisplay({ content }) {
     const title = (
       <div
         key={chapter.key}
-        className="MediaAnalysisDisplay-chapterTitle"
+        className={classNames({
+          'MediaAnalysisDisplay-chapterTitle': true,
+          'MediaAnalysisDisplay-chapterTitle--interactive': canRenderMediaPlayer
+        })}
         style={{
           width: `${widthInPercentage}%`,
           backgroundColor: `${chapter.color}`,
-          color: `${getContrastColor(chapter.color)}`,
-          cursor: canRenderMediaPlayer ? 'pointer' : 'default'
+          color: `${getContrastColor(chapter.color)}`
         }}
         onClick={canRenderMediaPlayer ? () => handleChapterClick(index) : null}
         >
--- a/src/plugins/media-analysis/media-analysis-display.js
+++ b/src/plugins/media-analysis/media-analysis-display.js
@@ -1,20 +1,22 @@
+import { Tooltip } from 'antd';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
 import Markdown from '../../components/markdown.js';
 import ClientConfig from '../../bootstrap/client-config.js';
 import { getContrastColor } from '../../ui/color-helper.js';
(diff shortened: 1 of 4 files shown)


src/plugins/media-analysis/media-analysis-display.js
```diff
@@ -54,6 +62,15 @@ function MediaAnalysisDisplay({ content }) {
     setAreTextsExpanded(prevValue => !prevValue);
   };
 
+  const handleChapterClick = chapterIndex => {
+    if (playState !== MEDIA_PLAY_STATE.playing) {
+      playerRef.current.play();
+      setTimeout(() => playerRef.current.seekToPart(chapterIndex), 0);
+    } else {
+      playerRef.current.seekToPart(chapterIndex);
+    }
```
adrianaluchian, line 71: Tiny thing: would be easier/nicer to follow the code if evaluating `playState === MEDIA_PLAY_STATE.playing` than the "not" in this case.

src/plugins/media-analysis/media-analysis-display.js
```diff
@@ -62,20 +79,26 @@ function MediaAnalysisDisplay({ content }) {
...
+    const title = (
       <div
         key={chapter.key}
         className="MediaAnalysisDisplay-chapterTitle"
         style={{
           width: `${widthInPercentage}%`,
           backgroundColor: `${chapter.color}`,
-          color: `${getContrastColor(chapter.color)}`
+          color: `${getContrastColor(chapter.color)}`,
+          cursor: canRenderMediaPlayer ? 'pointer' : 'default'
```
adrianaluchian, line 90: I think I'd rather set `cursor` through a modifier class, `MediaAnalysisDisplay-chapterTitle--interactive` or so. We normally only set the styles inline when they cannot be set through a CSS class.

src/plugins/media-analysis/media-analysis.yml
```diff
@@ -25,3 +25,6 @@ importFromCsv:
 segmentsDropzoneInfo:
   en: You can drop a CSV file in this box in order to import segments
   de: Sie können eine CSV-Datei in diese Box ziehen, um Segmente zu importieren
+startPlaybackFromHere:
+  en: Click here to start the playback beginning with this part
```
adrianaluchian, line 29: How about "Click here to play from this part"? I find it might be clearer, at least in English 🤔 
For whoever does read it, I mean 😅

src/plugins/media-analysis/media-analysis-display.js
ahelmberger: True. Switched it around!

src/plugins/media-analysis/media-analysis-display.js
ahelmberger: Agree. I created a `MediaAnalysisDisplay-chapterTitle--interactive` class.

src/plugins/media-analysis/media-analysis.yml
ahelmberger: Maybe someone will read it :sweat_smile:

I had something along these lines initially but it's hard to translate that to German nicely. I changed it to your suggestion (but only the English text).

adrianaluchian (approved):

PR #688, opened 2022-10-25
EDU-857 TOC plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-857

### DIFF
--- a/src/plugins/table-of-contents/table-of-contents-display.js
+++ b/src/plugins/table-of-contents/table-of-contents-display.js
@@ -1,0 +1,41 @@
+import React, { useEffect, useState } from 'react';
+import { sectionDisplayProps } from '../../ui/default-prop-types.js';
+
+export default function TableOfContentsDisplay({ content }) {
+  const [nodes, setNodes] = useState([]);
+
+  const { minLevel, maxLevel } = content;
+
+  useEffect(() => {
+    const nodeList = [...window.document.querySelectorAll('[data-header="true"]')]
+      .map(element => ({
+        id: element.id,
+        text: element.textContent,
+        level: Number.parseInt(element.tagName.replace(/\D/g, ''), 10) || null
+      }))
+      .filter(node => typeof node.level === 'number' && node.level >= minLevel && node.level <= maxLevel);
+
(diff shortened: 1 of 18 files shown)


src/plugins/table-of-contents/table-of-contents-info.js
```diff
@@ -0,0 +1,63 @@
...
+  cloneContent(content) {
+    return cloneDeep(content);
+  }
+
+  // eslint-disable-next-line no-unused-vars
+  redactContent(content, _targetRoomId) {
+    return cloneDeep(content);
+  }
+
+  // eslint-disable-next-line no-unused-vars
```
adrianaluchian, line 57: As much as I dislike comments, I think in these cases it's a good idea to always have the arguments consistently present 👍

src/plugins/table-of-contents/table-of-contents-display.js
```diff
@@ -0,0 +1,41 @@
...
+        level: Number.parseInt(element.tagName.replace(/\D/g, ''), 10) || null
+      }))
+      .filter(node => typeof node.level === 'number' && node.level >= minLevel && node.level <= maxLevel);
+
+    setNodes(nodeList);
+  }, [minLevel, maxLevel]);
+
+  if (!nodes.length) {
+    return null;
+  }
```
adrianaluchian, line 23: I guess we could not early return and just let the display component render whatever it found as nodes, even nothing if there's nothing.
The early return made me think that there might be a danger of using `nodes` beyond this point, but there isn't.

adrianaluchian:
Nice and clean implementation 👏 
Just a tiny complication suggestion from me: maybe we could also have the `width` setting here. With that, the user could basically "style" the table of content a bit (an 80% with would basically add some indentation, which might look better with short headline texts).
Also I think we probably have `width` in all other plugins.

src/plugins/table-of-contents/table-of-contents-info.js
ahelmberger: I think so too, also, when adding a header, we will have to re-introduce them anyways again.

PR #689, opened 2022-10-25
EDU-896 Change timeline selection

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-896

Screen Shot 2022-10-25 at 17 04 36


src/components/confirmation-dialogs.yml
```diff
@@ -20,10 +20,10 @@ deleteRoomConfirmation:
   en: Do you really want to delete room "{roomName}"?
   de: Möchtest du den Raum "{roomName}" wirklich löschen?
 deleteRoomMemberConfirmation:
-  en: Do you really want remove member "{memberDisplayName}" from the room?
+  en: Do you really want to remove member "{memberDisplayName}" from the room?
   de: Möchtest du das Mitglied "{memberDisplayName}" wirklich vom Raum entfernen?
 deleteRoomInvitationConfirmation:
-  en: Do you really want revoke room invitation send to "{invitationEmail}"?
+  en: Do you really want to revoke room invitation send to "{invitationEmail}"?
```
ahelmberger, line 26: Actually, not related to the ticket, but there's another type, it should be:

> Do you really want to revoke room invitation sen**t** to "{invitationEmail}"?

ahelmberger:
It's a lot more fun playing with the timeline after these changes, it feels much more intuitive :+1:

src/components/confirmation-dialogs.yml
adrianaluchian: Definitely, I had missed that one. Changed!

adrianaluchian:
> It's a lot more fun playing with the timeline after these changes, it feels much more intuitive 👍

Indeed it is, usage did reveal that the reflex is to click-and-select 😬 I like it more as well!

PR #690, opened 2022-10-25
EDU-902 Fix usage of old field in query lookup

adrianaluchian:
Closes already old ooops bug https://educandu.atlassian.net/browse/EDU-902


ahelmberger (approved):
Oops

PR #691, opened 2022-10-25
EDU-853 Allow tags and search tokens with less than 3 characters

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-853

### DIFF
--- a/src/services/document-service.spec.js
+++ b/src/services/document-service.spec.js
@@ -1055,6 +1055,13 @@ describe('document-service', () => {
       });
     });
 
+    describe('when the query consists of multiple tokens so that both partial and exact matching logic has to be applied', () => {
+      it('should return documents both matching partially as well as exactly', async () => {
+        const results = await sut.getSearchableDocumentsMetadataByTags('structor 1');
+        expect(results).toHaveLength(2);
+      });
+    });
+
     describe('when the query matches a single document', () => {
       it('should project the data correctly', async () => {
         const results = await sut.getSearchableDocumentsMetadataByTags('Wolf   gang \t beat Oven');
--- a/src/resources/resources.json
+++ b/src/resources/resources.json
@@ -1396,6 +1396,7 @@
       "filesView_list": "List",
       "filesView_grid": "Grid",
       "publicStorageWarning": "This directory is publicly accessible",
+      "searchTextTooShort": "The search text has to be at least {minCharCount, plural, =1 {# character} other {# characters}} long",
       "searchOngoing": "Searching …",
       "searchResultInfo": "You are seeing results for the search term <0>{searchTerm}</0>",
       "backToDirectoryView": "Back to directory view",
@@ -1409,6 +1410,7 @@
       "filesView_list": "List",
       "filesView_grid": "Grid",
       "publicStorageWarning": "Dieses Verzeichnis ist öffentlich zugänglich",
+      "searchTextTooShort": "Der Suchbegriff muss mindestens {minCharCount} Zeichen lang sein",
       "searchOngoing": "Suche …",
       "searchResultInfo": "Du siehst Ergebnisse für den Suchbegriff <0>{searchTerm}</0>",
       "backToDirectoryView": "Zurück zur Verzeichnisansicht",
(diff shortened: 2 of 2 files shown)


src/components/search-bar.js
```diff
@@ -18,10 +16,10 @@ export default function SearchBar({ initialValue, autoFocus, onSearch }) {
   };
 
   const handleSearch = value => {
-    if (value.length < MIN_CHAR_COUNT) {
+    if (!value.length) {
       Modal.error({
         title: t('common:error'),
-        content: t('common:searchTextTooShort', { minCharCount: MIN_CHAR_COUNT }),
+        content: t('common:searchTextTooShort', { minCharCount: 1 }),
```
adrianaluchian, line 22: Maybe we should change the validation to say that a value is required. Seems a bit weird and inconsistent this way now.

src/services/document-service.js
```diff
@@ -81,35 +86,52 @@ class DocumentService {
...
+
+    for (const token of tokens) {
+      const isNegative = token.length > 1 && token.startsWith('-');
+      const rawToken = token.substr(isNegative ? 1 : 0).toLowerCase();
+      if (isNegative) {
+        allNegativeTokens.add(rawToken);
+      } else if (rawToken.length < DOCUMENT_PARTIAL_SEARCH_THRESHOLD) {
+        allPositiveTokens.add(rawToken);
+        positiveExactTokens.add(rawToken);
+      } else {
```
adrianaluchian, line 102: Please bear with my little nitpicking here.
This 3rd case is a little hard to follow in one's mind (or this one's mind at least), you have to do a recap "so not negative, then length >= ..." 😅 
I normally like to avoid cascading `else` blocks as much I try to avoid nested `if` blocks, when possible, but here I think it would make it easier to follow as :
```
else {
     allPositiveTokens.add(rawToken);

     if (rawToken.length < DOCUMENT_PARTIAL_SEARCH_THRESHOLD) {
        positiveExactTokens.add(rawToken);
     } else {
        positivePartialTokens.add(rawToken);
    }
}
````
Oooor... we just set `allPositiveTokens` by joining `positiveExactTokens` and `positivePartialTokens` at the end of the iteration.

File: src/services/document-service.js [hunk 3]
```diff
@@ -81,35 +86,52 @@ class DocumentService {
...
+        allPositiveTokens.add(rawToken);
+        positivePartialTokens.add(rawToken);
+      }
+    }
 
-    if (!positiveTokens.size) {
+    if (!positiveExactTokens.size && !positivePartialTokens.size) {
```
adrianaluchian, line 108: Basically `!allPositiveTokens.size`, right?

src/services/document-service.spec.js
```diff
@@ -972,7 +972,7 @@ describe('document-service', () => {
         description: 'Description 1',
         slug: 'doc-1',
         sections: [],
-        tags: ['music', 'instructor', 'Dj.D', 'Cretu'],
+        tags: ['music', 'instructor', 'Dj.D', 'Cretu', '1'],
```
adrianaluchian, line 975: I'm sure of the author of this "Cretu" name in code 😊

src/services/document-service.spec.js
```diff
@@ -1071,7 +1071,7 @@ describe('document-service', () => {
       });
     });
 
-    describe('when I search with a query that returns multiple documents', () => {
+    describe('when the query matches multiple documents', () => {
       it('does not contain archived documents', async () => {
         const results = await sut.getSearchableDocumentsMetadataByTags('music');
         expect(results.map(result => result.title)).not.toContain('Doc 4');
```
adrianaluchian, line 1077: We've got the test for searching "1 or 2" and the one searching "Doc 4", but I think we need one testing something like "Music 2", to cover the case of both tags matching in one single document, one matching exactly and one partially. This is a case not covered in the current tests as far as I could spot.

src/components/search-bar.js
ahelmberger: Agreed, I created a more specific validation message.

src/services/document-service.js
ahelmberger: Good idea. I followed your first suggestion and divided the logic into nested binary decisions.

src/services/document-service.js
ahelmberger: Yes, you're right (I introduced the `allPositiveTokens` at a later point not realizing that it can also simplify this condition), changed!

src/services/document-service.spec.js
ahelmberger: It was **not** Maria Magdalena :rofl:

src/services/document-service.spec.js
ahelmberger: True. I added a test case with two terms, one of them matching one doc partially, the other one matching a different doc exactly.

src/services/document-service.spec.js
adrianaluchian: 😂

PR #692, opened 2022-10-25
EDU-899 Remove interval-trainer plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-899


src/plugins/interval-trainer/interval-trainer-display.js
```diff
@@ -1,279 +0,0 @@
-import { Button } from 'antd';
-import { Piano } from 'react-piano';
-import { RadialChart } from 'react-vis';
```
adrianaluchian, line 3: Can this guy be uninstalled then?

ahelmberger: Good catch, I didn't think about the modules. I could remove both `react-piano` and `react-vis` :man_dancing:

adrianaluchian (approved):

PR #693, opened 2022-10-25
EDU-901 Add caption to table-of-contents plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-901

### DIFF
--- a/src/plugins/table-of-contents/table-of-contents-display.js
+++ b/src/plugins/table-of-contents/table-of-contents-display.js
@@ -5,7 +5,7 @@ import { sectionDisplayProps } from '../../ui/default-prop-types.js';
 export default function TableOfContentsDisplay({ content }) {
   const [nodes, setNodes] = useState([]);
 
-  const { minLevel, maxLevel, caption } = content;
+  const { minLevel, maxLevel, text } = content;
 
   useEffect(() => {
     const nodeList = [...window.document.querySelectorAll('[data-header="true"]')]
@@ -21,10 +21,8 @@ export default function TableOfContentsDisplay({ content }) {
 
   return (
     <div className="TableOfContentsDisplay">
-      {!!caption && (
-        <h1 className="TableOfContentsDisplay-caption">
-          <Markdown inline>{caption}</Markdown>
-        </h1>
+      {!!text && (
+        <Markdown className="TableOfContentsDisplay-text">{text}</Markdown>
       )}
       {!!nodes.length && (
         <ul className="TableOfContentsDisplay-nodeList">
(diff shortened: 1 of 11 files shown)


src/plugins/table-of-contents/table-of-contents-display.js
```diff
@@ -1,10 +1,11 @@
 import React, { useEffect, useState } from 'react';
+import Markdown from '../../components/markdown.js';
 import { sectionDisplayProps } from '../../ui/default-prop-types.js';
 
 export default function TableOfContentsDisplay({ content }) {
   const [nodes, setNodes] = useState([]);
 
-  const { minLevel, maxLevel } = content;
+  const { minLevel, maxLevel, caption } = content;
```
adrianaluchian, line 8: Could we maybe rather call this `title`? 😬 
We always call it so in the other plugins, except for for `music-xml-viewer` and `pdf-viewer`, where it actually does not act as a title but as an explanation text.

src/plugins/table-of-contents/table-of-contents-display.js
```diff
@@ -18,21 +19,26 @@ export default function TableOfContentsDisplay({ content }) {
...
-          >
-          <a href={`#${node.id}`}>{node.text}</a>
-        </li>
-      ))}
-    </ul>
+    <div className="TableOfContentsDisplay">
+      {!!caption && (
+        <h1 className="TableOfContentsDisplay-caption">
+          <Markdown inline>{caption}</Markdown>
+        </h1>
```
adrianaluchian, line 27: I have a feeling that if we format the title ourselves (which we don't do with other titles, but I do get that a TOC title makes no sense otherwise), I have a feeling we might need to offer a choice of headline size 🧐 But should we wait and see if that's necessary? What do you think?

src/plugins/table-of-contents/table-of-contents-display.js
ahelmberger: So, as we agreed, I repurposed the field to be named `text` and to be a block markdown field making things much more flexible for the editor.

src/plugins/table-of-contents/table-of-contents-display.js
ahelmberger: See my comment above.

src/plugins/table-of-contents/table-of-contents-info.js
```diff
@@ -51,14 +54,19 @@ class TableOfContentsInfo {
...
-    return cloneDeep(content);
+  redactContent(content, targetRoomId) {
+    const redactedContent = cloneDeep(content);
+
+    redactedContent.text = this.gfm.redactCdnResources(
+      redactedContent.text,
+      url => couldAccessUrlFromRoom(url, targetRoomId) ? url : ''
+    );
+
+    return redactedContent;
```
adrianaluchian, line 65: Woop woop, now we need tests here 😋

ahelmberger: Added (and also some for the markdown plugin which works exactly the same).

adrianaluchian: 👍

adrianaluchian (approved):

PR #696, opened 2022-10-27
EDU-861 Create markdown with image plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-861

Screen Shot 2022-10-26 at 16 13 58


src/plugins/markdown-with-image/markdown-with-image-display.js
```diff
@@ -0,0 +1,41 @@
...
+  const clientConfig = useService(ClientConfig);
+  const { text, image } = content;
+
+  const imageSrc = getAccessibleUrl({ url: image.sourceUrl, cdnRootUrl: clientConfig.cdnRootUrl });
+  const imageWrapperStyle = { float: image.position };
+  if (image.position === IMAGE_POSITION.left) {
+    imageWrapperStyle.paddingLeft = 0;
+  } else {
+    imageWrapperStyle.paddingRight = 0;
+  }
```
ahelmberger, line 21: It would be nicer to extract these inline styles into two classes:

* `MarkdownWithImageDisplay-imageWrapper--left`
* `MarkdownWithImageDisplay-imageWrapper--right`

src/plugins/markdown-with-image/markdown-with-image-display.js
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <div className="MarkdownWithImageDisplay">
+      <div className="MarkdownWithImageDisplay-verticallyStretchedContent">
```
ahelmberger, line 25: I haven't worked with `float`ing HTML since early childhood, so I was not sure if the "clearfix hack" one needed back then was still a thing. Obviously it's not, but there is something more modern: https://stackoverflow.com/a/31954550. And it seems to be supported by all browsers: https://caniuse.com/?search=flow-root

We could use that just to follow current practices, I guess, in reality it probably won't make any visual difference, I guess.

src/plugins/markdown-with-image/markdown-with-image-display.js
adrianaluchian: Makes sense and cleans up the code, done, thanks

src/plugins/markdown-with-image/markdown-with-image-display.js
adrianaluchian: Nice, works, thanks

ahelmberger (approved):

PR #702, opened 2022-10-31
EDU-907 Create memory editor component

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-907

Screen Shot 2022-10-31 at 19 44 09
Screen Shot 2022-10-31 at 19 45 18

### DIFF
--- a/src/plugins/memory/memory-editor.js
+++ b/src/plugins/memory/memory-editor.js
@@ -1,11 +1,98 @@
-import React from 'react';
+import { SIZE } from './constants.js';
+import React, { useState } from 'react';
+import { Button, Form, Radio } from 'antd';
+import { useTranslation } from 'react-i18next';
+import cloneDeep from '../../utils/clone-deep.js';
+import { resizeTilePairs } from './memory-utils.js';
+import { FORM_ITEM_LAYOUT } from '../../domain/constants.js';
+import MarkdownInput from '../../components/markdown-input.js';
 import { sectionEditorProps } from '../../ui/default-prop-types.js';
 
-// eslint-disable-next-line no-unused-vars
+const FormItem = Form.Item;
+const RadioGroup = Radio.Group;
+const RadioButton = Radio.Button;
+
 function MemoryEditor({ content, onContentChanged }) {
+  const { t } = useTranslation('memory');
+
+  const { title, size, tilePairs } = content;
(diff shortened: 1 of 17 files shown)


src/plugins/memory/memory-editor.js
```diff
@@ -1,11 +1,98 @@
...
 
-// eslint-disable-next-line no-unused-vars
+const FormItem = Form.Item;
+const RadioGroup = Radio.Group;
+const RadioButton = Radio.Button;
+
 function MemoryEditor({ content, onContentChanged }) {
+  const { t } = useTranslation('memory');
+
+  const { title, size, tilePairs } = content;
```
adrianaluchian, line 18: Actually I'm considering dropping the `title`, given the `markdown` plugin can always be used for title and/or description. I've also made a list of where we stand with the common plugin fields and we're a bit mixed with `title` though most plugins actually don't have it:
https://educandu.atlassian.net/wiki/spaces/WIKI/pages/145096705/Plugins+data+models

src/plugins/memory/memory.yml
```diff
@@ -1,3 +1,12 @@
 name:
   en: Memory
   de: Memory
+tilePairs:
+  en: Tile pairs
+  de: Kachelpaare
```
ahelmberger, line 6: Just a thought: Now that I see the German translation I wonder if we shouldn't go with the term "card" instead of "tile" as it would resemble an analog game more closely. Even if we want to stick to "tile" in code, I'd suggest to use the words "card"/"Karte" on the UI here.

src/plugins/memory/memory-editor.js
ahelmberger: I agree, `width` would be more consistent and also bring functionality not available otherwise. If editors want a header, there's the `markdown` plugin for it.

src/plugins/memory/memory.yml
adrianaluchian: Changed the translations as you suggested 👍

src/plugins/memory/memory-editor.js
adrianaluchian: Ok then, I've dropped `title` and added `width`.

ahelmberger: :+1:

ahelmberger (approved):

PR #704, opened 2022-11-02
EDU-904 Wikimedia in resource picker

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-904


src/components/resource-picker/resource-picker-dialog.js
```diff
@@ -2,11 +2,9 @@ import React from 'react';
 import { Modal } from 'antd';
 import PropTypes from 'prop-types';
 import ResourceSelector from './resource-selector.js';
-import { STORAGE_LOCATION_TYPE } from '../../domain/constants.js';
+import { SOURCE_TYPE } from '../../domain/constants.js';
 
-// This can later be made configurable by individual plugins
-// (when we have more locations than just from the public/private CDN).
```
adrianaluchian, line 8: It finally happened..happened..happened...

src/utils/source-utils.js
```diff
@@ -66,6 +66,10 @@ export function getSourceType({ url, cdnRootUrl }) {
     return SOURCE_TYPE.youtube;
   }
 
+  if (url.startsWith('https://upload.wikimedia.org/')) {
+    return SOURCE_TYPE.wikimediaCommons;
+  }
```
adrianaluchian, line 71: Let's please update the tests too.

src/components/resource-picker/wikimedia-commons-search.yml
```diff
@@ -0,0 +1,24 @@
...
+  de: PDF
+initialFilePreviewHeader:
+  en: Currently selected file
+  de: Momentan ausgewählte Datei
+initialFilePreviewFooter:
+  en: In order to select a different file, start a new search using the search field above
+  de: Um eine andere Datei auszuwählen, starte eine neue Suche im Suchfeld oben
+loadMoreSearchResults:
+  en: load more
+  de: mehr laden
```
adrianaluchian, line 24: We've got the translation
```
more:
  en: more ...
  de: mehr ...
```
as well, maybe we could streamline? "load more" is clearer, but I like the shorter version actually. And I think people know through common UX patterns out there that "more ..." is for loading more 🤔 Or I presume they know

src/components/resource-picker/wikimedia-commons-files-grid-viewer.js
```diff
@@ -0,0 +1,96 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { Spin, Tooltip } from 'antd';
+import { useTranslation } from 'react-i18next';
+import PreviewIcon from '../icons/general/preview-icon.js';
+import { wikimediaFileShape } from '../../ui/default-prop-types.js';
+import WikimediaCommonsIcon from '../icons/wikimedia-commons/wikimedia-commons-icon.js';
+import ActionButton, { ActionButtonGroup, ACTION_BUTTON_INTENT } from '../action-button.js';
+
+function WikimediaCommonsFilesGridViewer({
```
adrianaluchian, line 11: Honestly since there is only one type of viewer for wikimedia (and since it's dedicated), we could shorten the component to `WikimediaCommonsFilesViewer`? What do you think.

adrianaluchian:
Wonderful feature 😍 Nicely done and integrated, this feels like yet another huge game changer for the project!

adrianaluchian:
There would be a few suggestions from my side.
1. Regarding opening the resource picker having no initial URL set in the input field, then selecting the Wikimedia Commons tab
- I think we should better indicate to the user that resources **do** exist but they need to be searched for
- maybe show text and icon similar to this example:
Screen Shot 2022-11-03 at 09 29 23

2. Regarding changing the selection for search file type (image/audio/video/PDF)
- when you make a search, then change the search file types, this creates a discrepancy between what you see and what you've actually searched for; for this same reason we introduced the text under the viewer showing the user what results they are viewing at the moment (for which search text)
- when on top of this, you then click "load more", you see the screen loading/searching and you're under the impression that a new search, with the given text + search file types, is carried out, but actually the previously selected file types are used for the search, not what you see selected at the moment
- I would however not suggest to include the file types in the text under the search results ( "Du siehst x Ergebnisse für den Suchbegriff y") because this would be too much text and when it's too much nobody reads. What I would suggest would be to trigger a new search each time a search type checkbox state changes. This would be then also the same behaviour as with the search input itself, which triggers a search also on click

3. Regarding the "load more" link, I can imagine it was tricky to think how to integrate that one. My idea would be something like this (with some opacity to the bar, to see underneath a little as you scroll):
Screen Shot 2022-11-03 at 10 00 47

src/components/resource-picker/resource-picker-dialog.js
ahelmberger: :laughing: although I might be slightly mad for some time now

src/components/resource-picker/wikimedia-commons-search.yml
ahelmberger: I streamlined it (and another place it was used, too).

src/components/resource-picker/wikimedia-commons-files-grid-viewer.js
ahelmberger: Yes, it's still long enough :laughing:, done!

ahelmberger:
> Wonderful feature heart_eyes Nicely done and integrated, this feels like yet another huge game changer for the project!

:bow: danke!

ahelmberger:
Thanks for the valuable input :pray:, I gave the whole tab an overhaul trying to incorporate all suggestions, and I think it is much clearer and more user-friendly now:

> 1. Regarding opening the resource picker having no initial URL set in the input field, then selecting the Wikimedia Commons tab
> 
> * I think we should better indicate to the user that resources **do** exist but they need to be searched for

We show a message now inviting the user to search

> 2. Regarding changing the selection for search file type (image/audio/video/PDF)
> 
> * when you make a search, then change the search file types, this creates a discrepancy between what you see and what you've actually searched for; for this same reason we introduced the text under the viewer showing the user what results they are viewing at the moment (for which search text)
> * when on top of this, you then click "load more", you see the screen loading/searching and you're under the impression that a new search, with the given text + search file types, is carried out, but actually the previously selected file types are used for the search, not what you see selected at the moment
> * I would however not suggest to include the file types in the text under the search results ( "Du siehst x Ergebnisse für den Suchbegriff y") because this would be too much text and when it's too much nobody reads. What I would suggest would be to trigger a new search each time a search type checkbox state changes. This would be then also the same behaviour as with the search input itself, which triggers a search also on click

The elements are now displayed differently on screen and there is a dedicated search button that makes it clearer that the category selection only applies to a new search. For continuing an already started search (i.e. loading more items), I implemented "infinite scrolling", so the user does not have to press another button, it will just load more results as the user scrolls in the grid viewer, making it also clearer that the checkboxes on top only apply to "fresh" searches.

> 3. Regarding the "load more" link, I can imagine it was tricky to think how to integrate that one. My idea would be something like this (with some opacity to the bar, to see underneath a little as you scroll):

See Point 2, there's no "load more" link anymore.

adrianaluchian:
> Thanks for the valuable input 🙏, I gave the whole tab an overhaul trying to incorporate all suggestions, and I think it is much clearer and more user-friendly now:
> 
> > 1. Regarding opening the resource picker having no initial URL set in the input field, then selecting the Wikimedia Commons tab
> > 
> > 
> > * I think we should better indicate to the user that resources **do** exist but they need to be searched for
> 
> We show a message now inviting the user to search
> 
> > 2. Regarding changing the selection for search file type (image/audio/video/PDF)
> > 
> > 
> > * when you make a search, then change the search file types, this creates a discrepancy between what you see and what you've actually searched for; for this same reason we introduced the text under the viewer showing the user what results they are viewing at the moment (for which search text)
> > * when on top of this, you then click "load more", you see the screen loading/searching and you're under the impression that a new search, with the given text + search file types, is carried out, but actually the previously selected file types are used for the search, not what you see selected at the moment
> > * I would however not suggest to include the file types in the text under the search results ( "Du siehst x Ergebnisse für den Suchbegriff y") because this would be too much text and when it's too much nobody reads. What I would suggest would be to trigger a new search each time a search type checkbox state changes. This would be then also the same behaviour as with the search input itself, which triggers a search also on click
> 
> The elements are now displayed differently on screen and there is a dedicated search button that makes it clearer that the category selection only applies to a new search. For continuing an already started search (i.e. loading more items), I implemented "infinite scrolling", so the user does not have to press another button, it will just load more results as the user scrolls in the grid viewer, making it also clearer that the checkboxes on top only apply to "fresh" searches.
> 
> > 3. Regarding the "load more" link, I can imagine it was tricky to think how to integrate that one. My idea would be something like this (with some opacity to the bar, to see underneath a little as you scroll):
> 
> See Point 2, there's no "load more" link anymore.

Nice, looks super good with the initial search message! And infinite scrolling is also so much better than the "load more" option 👏 
One thing I noticed just, when you search for something, then trigger infinite loading a few times, then type in the search field again, the change even seems laggy, takes some time till the text change is visible. Funny because on the initial search the typing is real time 🤔 Don't know if you have any ideas what might be the case there.

ahelmberger:
> Nice, looks super good with the initial search message! And infinite scrolling is also so much better than the "load more" option clap One thing I noticed just, when you search for something, then trigger infinite loading a few times, then type in the search field again, the change even seems laggy, takes some time till the text change is visible. Funny because on the initial search the typing is real time thinking Don't know if you have any ideas what might be the case there.

I switched back to the debounced input in order to avoid the rendering lag on each key stroke.

src/components/debounced-input.js
```diff
@@ -29,13 +29,21 @@ function DebouncedInput({ apiRef, timeLimit, elementType: Component, value, onCh
     };
   }
 
+  if (onPressEnter) {
```
adrianaluchian, line 32: Ok yeah, of course, this was needed now..

adrianaluchian (approved):

src/components/debounced-input.js
ahelmberger: Yeah, I followed the `onSearch` example :-)

src/components/resource-picker/wikimedia-commons-search.js
```diff
@@ -108,7 +105,19 @@ function WikimediaCommonsSearch({
     }
 
     setHasSearchedAtLeastOnce(true);
-    onSearchParamsChange({ searchTerm: typedInSearchTerm, searchFileTypes: selectedSearchFileTypes });
+
+    // This will trigger search in the parent component:
```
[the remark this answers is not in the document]
ahelmberger: OK, dropped!

PR #707, opened 2022-11-09
EDU-916 Expand memory plugin data model

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-916

Screen Shot 2022-11-09 at 21 52 03

---

Screen Shot 2022-11-09 at 21 53 02

### DIFF
--- a/src/plugins/memory/memory-tile.js
+++ b/src/plugins/memory/memory-tile.js
@@ -1,17 +1,27 @@
 import PropTypes from 'prop-types';
-import React, { useRef } from 'react';
+import React, { useEffect, useRef } from 'react';
 import Markdown from '../../components/markdown.js';
 import ClientConfig from '../../bootstrap/client-config.js';
+import { analyzeMediaUrl } from '../../utils/media-utils.js';
 import { getAccessibleUrl } from '../../utils/source-utils.js';
-import { getResourceType } from '../../utils/resource-utils.js';
 import { useService } from '../../components/container-context.js';
 import MediaPlayer from '../../components/media-player/media-player.js';
 import { MEDIA_SCREEN_MODE, RESOURCE_TYPE } from '../../domain/constants.js';
 
-function MemoryTile({ text, sourceUrl }) {
+function MemoryTile({ text, sourceUrl, isFlipped }) {
   const mediaPlayerRef = useRef();
   const clientConfig = useService(ClientConfig);
-  const resourceType = sourceUrl ? getResourceType(sourceUrl) : RESOURCE_TYPE.none;
+  const { resourceType } = sourceUrl ? analyzeMediaUrl(sourceUrl) : { resourceType: RESOURCE_TYPE.none };
+
+  useEffect(() => {
+    if (mediaPlayerRef?.current) {
+      if (isFlipped) {
+        mediaPlayerRef.current.play?.();
+      } else {
+        mediaPlayerRef.current.stop?.();
+      }
+    }
+  }, [isFlipped]);
 
   const renderMedia = () => {
     const url = getAccessibleUrl({ url: sourceUrl, cdnRootUrl: clientConfig.cdnRootUrl });
@@ -52,11 +62,13 @@ function MemoryTile({ text, sourceUrl }) {
 }
 
 MemoryTile.propTypes = {
+  isFlipped: PropTypes.bool,
   sourceUrl: PropTypes.string,
   text: PropTypes.string
 };
 
 MemoryTile.defaultProps = {
+  isFlipped: false,
   sourceUrl: '',
   text: ''
 };
--- a/src/plugins/memory/memory-tile.js
+++ b/src/plugins/memory/memory-tile.js
@@ -1,5 +1,5 @@
-import React from 'react';
 import PropTypes from 'prop-types';
+import React, { useRef } from 'react';
 import Markdown from '../../components/markdown.js';
 import ClientConfig from '../../bootstrap/client-config.js';
 import { getAccessibleUrl } from '../../utils/source-utils.js';
@@ -9,6 +9,7 @@ import MediaPlayer from '../../components/media-player/media-player.js';
 import { MEDIA_SCREEN_MODE, RESOURCE_TYPE } from '../../domain/constants.js';
 
 function MemoryTile({ text, sourceUrl }) {
+  const mediaPlayerRef = useRef();
   const clientConfig = useService(ClientConfig);
   const resourceType = sourceUrl ? getResourceType(sourceUrl) : RESOURCE_TYPE.none;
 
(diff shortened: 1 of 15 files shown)


src/plugins/memory/memory.yml
```diff
@@ -10,3 +10,12 @@ tileA:
 tileB:
   en: Card B
   de: Karte B
+text:
+  en: text
```
ahelmberger, line 14: The other labels in the editor (like "Width", "Card pairs", etc.) are capitalized also in english. Let's also adjust "text" and "media URL" ("Text" already exists in `common.yml`)

src/plugins/memory/memory-tile.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    playerRef.current?.seekTo(0);
+    timeoutToPlayMedia = setTimeout(() => {
+      if (isMounted.current && playMedia) {
+        setIsPlaying(true);
+      }
+    }, 500);
+
+  }, [playerRef, playMedia]);
```
ahelmberger, line 48: I was trying to figure out what this `useEffect` is doing in contrast to just using the prop `playMedia`. It also uses the the global variable `timeoutToPlayMedia` shared between all possible memory tiles on the page. Is this intentional?

src/plugins/memory/memory-tile.js
```diff
@@ -0,0 +1,92 @@
...
+        return <img className="MemoryTile-image" src={accessibleUrl} />;
+      default:
+        return null;
+    }
+  };
+
+  return (
+    <div className="MemoryTile">
+      {!!text && (<div className="MemoryTile-markdown"><Markdown>{text}</Markdown></div>)}
+      {!!sourceUrl && renderMedia()}
```
ahelmberger, line 75: As `renderMedia` uses `accessibleUrl` we could also make it the condition here (although it is derived from `sourceUrl`).

src/plugins/memory/memory.yml
```diff
@@ -10,3 +10,12 @@ tileA:
...
+  de: Text
+url:
+  en: media URL
+  de: Medien-URL
+preview:
+  en: Card pairs preview
+  de: Kartenpaarevorschau
```
ahelmberger, line 21: The preview just shows one pair at a time, so we might wanna go with "Card pair preview" / "Kartenpaarvorschau".

ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-916

This starts to become really nice. We might wanna consider making the markdown text centered by default, as I think this would fit the game better (it will most like often be just 1-2 lines of text, or in the case of vocab training just a single word). Not related to the ticket, though.

:heart_eyes:

src/plugins/memory/memory.yml
adrianaluchian: Indeed, streamlined them, thanks for noticing.

src/plugins/memory/memory-tile.js
adrianaluchian: Uokey, though the dependency on it is not immediately noticeable and `accessibleUrl` cannot be empty when `sourceUrl` has a value, but just in case I've changed it.

src/plugins/memory/memory.yml
adrianaluchian: True, makes sense, corrected.

src/plugins/memory/memory-tile.js
adrianaluchian: Though technically only the last flipped card should play and the previous should be stopped, actually this component should not directly impact the others. So yes, not a good place to put that variable, I've moved it into a `ref`.

adrianaluchian:
> > Closes https://educandu.atlassian.net/browse/EDU-916
> 
> This starts to become really nice. We might wanna consider making the markdown text centered by default, as I think this would fit the game better (it will most like often be just 1-2 lines of text, or in the case of vocab training just a single word). Not related to the ticket, though.
> 
> 😍

I was actually thinking if we should do that. It would only not fit with a full card of text, perhaps, but probably centered would still be the most needed use-case. I'll change it and let's see what feedback comes, if any.

src/plugins/memory/memory-tile.js
ahelmberger: I was just thinking we might not need the whole `useEffect` block. I tried removing it and handing down the `playMedia` prop directly to the react player and it worked in my use case. But I haven't tested all cases, so I was wondering why we need it?

ahelmberger:
> > > Closes https://educandu.atlassian.net/browse/EDU-916
> > 
> > 
> > This starts to become really nice. We might wanna consider making the markdown text centered by default, as I think this would fit the game better (it will most like often be just 1-2 lines of text, or in the case of vocab training just a single word). Not related to the ticket, though.
> > heart_eyes
> 
> I was actually thinking if we should do that. It would only not fit with a full card of text, perhaps, but probably centered would still be the most needed use-case. I'll change it and let's see what feedback comes, if any.

:+1: Cool!

src/plugins/memory/memory-tile.js
adrianaluchian: We need the `useEffect` so that when you flip a card, the media starts playing with a small delay. Otherwise it starts playing immediately, before you even get to see the card completely face-up. I find that not to be a great experience, possibly even startling for the user.

ahelmberger: Ah, ok! Got it, thanks for the explanation. In this case I think it would be clearer to set `isPlaying` initially to `false` in any case in line 23, so that it is only set to `true` in a delayed manner within the `useEffect`. What do you think?

adrianaluchian: Currently the media will play directly if the initial `playState` is `true` as there is no transition of the flip card to wait for. To me the delay makes sense in the case of actual state change, in this context. We actually only have the case of card flip, never of direct play, so I don't feel strongly in any direction. But I feel the intention of the code might be clearer this way? 🤷‍♀️

ahelmberger: The way it is now just has this strange sequence of events when you follow the case of `playMedia` set to `true` initially: `isPlaying` is set to `true` which makes the player play immediately. On the next tick (when `useEffect` is called the first time), we seek back to position 0 (`playerRef.current?.seekTo(0);`) and then 500 ms later we set it `isPlaying` to `true` again (which it already is). That made the code hard to understand for me. So even if I agree on the flip transition argument I would still set `isPlaying` to false initially to keep the code logic easier to follow. In our case it doesn't apply anyways.

adrianaluchian: Ok, if this were ever the use case, the reset to timecode 0 would happen weirdly again mid play, that's true. I've set it to `false` then. Thanks

ahelmberger (approved):

PR #708, opened 2022-11-10
EDU-920 batch room invitations

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-920

### DIFF
--- a/src/components/room-invitation-creation-modal.js
+++ b/src/components/room-invitation-creation-modal.js
@@ -87,7 +87,6 @@ function RoomInvitationCreationModal({ isVisible, onOk, onCancel, roomId }) {
       <Form
         form={form}
         layout="vertical"
-        name="send-room-invitations-form"
         initialValues={initialFormValues}
         onFinish={handleFormFinish}
         >
--- a/src/components/room-invitation-creation-modal.js
+++ b/src/components/room-invitation-creation-modal.js
@@ -98,7 +98,7 @@ function RoomInvitationCreationModal({ isVisible, onOk, onCancel, roomId }) {
           rules={emailsValidationRules}
           normalize={normalizeEmails}
           >
-          <TextArea onChange={handleOnEmailsChange} />
+          <TextArea onChange={handleOnEmailsChange} autoSize={{ minRows: 3, maxRows: 12 }} />
         </FormItem>
       </Form>
     </Modal>
--- a/src/components/room-invitation-creation-modal.js
+++ b/src/components/room-invitation-creation-modal.js
@@ -1,14 +1,20 @@
 import PropTypes from 'prop-types';
-import { Form, Modal } from 'antd';
 import Logger from '../common/logger.js';
+import { Form, Input, Modal } from 'antd';
 import { useTranslation } from 'react-i18next';
 import errorHelper from '../ui/error-helper.js';
-import EmailFormItem from './email-form-item.js';
 import React, { useEffect, useState } from 'react';
+import inputValidators from '../utils/input-validators.js';
 import RoomApiClient from '../api-clients/room-api-client.js';
 import { useSessionAwareApiClient } from '../ui/api-helper.js';
 
+const FormItem = Form.Item;
+const { TextArea } = Input;
+
 const logger = new Logger(import.meta.url);
+
+const normalizeEmails = value => value.toLowerCase();
+const extractEmails = value => value.match(/[^\s;,]+/g) || [];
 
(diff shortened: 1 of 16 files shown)


src/components/room-invitation-creation-modal.js
```diff
@@ -1,15 +1,21 @@
...
+import inputValidators from '../utils/input-validators.js';
 import RoomApiClient from '../api-clients/room-api-client.js';
 import { useSessionAwareApiClient } from '../ui/api-helper.js';
 
+const FormItem = Form.Item;
+const { TextArea } = Input;
+
 const logger = new Logger(import.meta.url);
 
+const normalizeEmails = value => value.toLowerCase();
```
adrianaluchian, line 16: This should be `normalizeEmail`, as it deals with just one at a time.

adrianaluchian: No, sorry, got it, it's applied over the input, not the extracted email 😬

src/components/room-invitation-creation-modal.js
```diff
@@ -1,15 +1,21 @@
...
+const extractEmails = value => value.match(/[^\s;,]+/g) || [];
```
adrianaluchian, line 17: How about `const extractEmails = value => (value.match(/[^\s;,]+/g) || []).filter(item => item);`
just to take out possible altogether missing emails (if ever that is the case)?
I'm thinking that those don't fall under the "invalid email" category 🤔

src/components/room-invitation-creation-modal.js
```diff
@@ -46,16 +77,29 @@ function RoomInvitationCreationModal({ isVisible, onOk, onCancel, roomId }) {
...
       visible={isVisible}
       okButtonProps={{ loading: isLoading }}
       forceRender
       >
-      <Form form={form} name="new-room-invitation-form" initialValues={initialFormValues} onFinish={handleFormFinish} layout="vertical">
-        <EmailFormItem name="email" />
+      <Form
+        form={form}
+        layout="vertical"
+        name="send-room-invitations-form"
```
adrianaluchian, line 90: Tiny thing: does the form need to have `name` set?

src/components/room-invitation-creation-modal.js
ahelmberger: I had something similar first, then changed the regex. This regex prevents any empty string matches, so `filter(item => item)` would never take out any item.

adrianaluchian:
Maybe we should put an info icon next to the input label, telling the user what separators they can use between email addresses.

src/components/room-invitation-creation-modal.js
adrianaluchian: Ah I see, smart regex, smart.

src/components/room-invitation-creation-modal.js
ahelmberger: I assumed as it was there before, but I just tried and I don't think it's necessary, so I removed it.

ahelmberger:
> Maybe we should put an info icon next to the input label, telling the user what separators they can use between email addresses.

Nice idea! Done.

adrianaluchian:
> > Maybe we should put an info icon next to the input label, telling the user what separators they can use between email addresses.
> 
> Nice idea! Done.

👍 Cool
Should we in this case move it to the end of the label? Looks rather weird with the required mark.

Screen Shot 2022-11-10 at 17 25 37

Or maybe we can remove the requirement altogether and just not do anything when the user presses "Send" without any emails written?

ahelmberger:
> > > Maybe we should put an info icon next to the input label, telling the user what separators they can use between email addresses.
> > 
> > 
> > Nice idea! Done.
> 
> +1 Cool Should we in this case move it to the end of the label? Looks rather weird with the required mark.
> Or maybe we can remove the requirement altogether and just not do anything when the user presses "Send" without any emails written?

I actually did both as it looks best. Validation will kick in anyways when no email is specified.

adrianaluchian:
> > > > Maybe we should put an info icon next to the input label, telling the user what separators they can use between email addresses.
> > > 
> > > 
> > > Nice idea! Done.
> > 
> > 
> > +1 Cool Should we in this case move it to the end of the label? Looks rather weird with the required mark.
> > Or maybe we can remove the requirement altogether and just not do anything when the user presses "Send" without any emails written?
> 
> I actually did both as it looks best. Validation will kick in anyways when no email is specified.

Wunderbar!

adrianaluchian (approved):

PR #710, opened 2022-11-10
EDU-921 Restrict document deletion in rooms to owners

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-921


adrianaluchian:
Maybe we should also update the `documentsModeInfo` text in the `room-metadata-form` to specify that rooms can be added or edited. Now the message writes:
"In exclusive room documents can be managed only by the owner, in collaborative rooms by all room members."

ahelmberger:
> documentsModeInfo

Great find! I updated these texts making them more specific.

adrianaluchian (approved):

PR #713, opened 2022-11-11
EDU-924 Add option for image border in markdown-with-image plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-924

Looks good on most images:

looksgood

Looks not so great on music:

looksbad

On whether the copyright should be inside or outside the border, to me there's not much difference, I opted for the solution requiring less HTML/CSS (inside).


ahelmberger:
Closing because we will not do the border, and add text top offset instead

PR #714, opened 2022-11-14
EDU-923 Replace media range selection means

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-923

Screen Shot 2022-11-14 at 13 40 32

### DIFF
--- a/src/components/media-player/media-range-selector.yml
+++ b/src/components/media-player/media-range-selector.yml
@@ -5,8 +5,8 @@ playbackRange:
   en: "Playback range: from {from, select, start {the beginning} other {timecode {from}}} to {to, select, end {the end} other {timecode {to}}}"
   de: "Abspielbereich: von {from, select, start {Anfang} other {Time-Code {from}}} bis {to, select, end {Ende} other {Time-Code {to}}}"
 selectRangeLabel:
-  en: Set current playback timecode
-  de: Setz den aktuellen Wiedergabe Time-Code
+  en: Set current playback position
+  de: Setze die aktuelle Wiedergabeposition
 setAsStart:
   en: as start time
   de: als Anfangszeit
@@ -19,6 +19,6 @@ loadingMessage:
 errorMessage:
   en: An error occurred while retrieving the media file
   de: Beim Laden der Mediendatei ist ein Fehler aufgetreten
-setMaxRange:
-  en: Set maximum range
-  de: Maximaler Abspielbereich setzen
+clearRange:
+  en: Clear range
+  de: Bereich löschen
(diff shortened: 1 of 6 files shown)


adrianaluchian:
@ahelmberger , added the button to (re)set the range to the max values, thanks for suggesting it. Plus/Minus the translations/texts of course, not sure about those.

src/components/media-player/media-range-selector.yml
```diff
@@ -4,9 +4,21 @@ modalTitle:
...
+  de: als Endezeit
 loadingMessage:
   en: Determining media information ...
   de: Medieninformationen werden ermittelt ...
 errorMessage:
   en: An error occurred while retrieving the media file
   de: Beim Laden der Mediendatei ist ein Fehler aufgetreten
+setMaxRange:
+  en: Set maximum range
+  de: Maximaler Abspielbereich setzen
```
ahelmberger, line 24: I'd suggest:

"Reset to maximum range" / "Auf maximalen Bereich zurücksetzen"

Watcha think?

src/components/media-player/media-range-selector.yml
```diff
@@ -4,9 +4,21 @@ modalTitle:
 playbackRange:
   en: "Playback range: from {from, select, start {the beginning} other {timecode {from}}} to {to, select, end {the end} other {timecode {to}}}"
   de: "Abspielbereich: von {from, select, start {Anfang} other {Time-Code {from}}} bis {to, select, end {Ende} other {Time-Code {to}}}"
+selectRangeLabel:
+  en: Set current playback timecode
+  de: Setz den aktuellen Wiedergabe Time-Code
```
ahelmberger, line 9: "Setz**e**" and "Wiedergab**e-T**ime-Code"

Maybe an even less technical alternative: "Set current playback position" / "Setze die aktuelle Wiedergabeposition" ???

src/components/media-player/media-range-selector.yml
adrianaluchian: I was initially thinking the same, but thought it's just very long. Can do it though, it will wrap on mobile.

src/components/media-player/media-range-selector.yml
adrianaluchian: Sounds good, I will change both texts 👍

adrianaluchian: Done!

src/components/media-player/media-range-selector.yml
adrianaluchian: Set as discussed.

ahelmberger: :+1:

ahelmberger (approved):

PR #718, opened 2022-11-18
EDU-941 Extract publicContext within document model

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-941


ahelmberger:
The "new" dialog is much cleaner, great work! :heart_eyes:

I found one deviation from the former behavior though, that we have to correct:

When cloning a document, the old implementation resets the four properties that make up the `publicContext` to their default values, while the reworked solution keeps the original values from the source document (at least as long as the cloning strategy dropdown stays unchanged). This is problematic as it:

* prevents non-maintainers from cloning+editing any document with restricted `allowedOpenContributions`
* enables "stealing" of the `verified` status by simply cloning and modifying an already verified document

In other words: in the `resetStates` function we have to disregard the `initialDocumentMetadata.publicContext` values when the dialog was opened in cloning mode.

src/components/document-metadata-modal.js
```diff
@@ -56,47 +67,65 @@ function DocumentMetadataModal({
...
+  );
+
+  const resetStates = useCallback(() => {
+    setTitle(initialDocumentMetadata.title || t('newDocument'));
+    setDescription(initialDocumentMetadata.description || '');
+    setSlug(initialDocumentMetadata.slug || '');
+    setTags(initialDocumentMetadata.tags || []);
+    setTagOptions(composeTagOptions(initialDocumentMetadata.tags));
+    setLanguage(initialDocumentMetadata.language || getDefaultLanguageFromUiLanguage(uiLanguage));
+    setPublicContext(initialDocumentMetadata.publicContext || getDefaultPublicContext());
```
ahelmberger, line 121: It would feel cleaner if we'd `deepClone`d this object (although technically irrelevant). Also, please see my comment at the end of this PR on the cloning case.

src/components/document-metadata-modal.yml
```diff
@@ -7,9 +7,6 @@ cloneDocument:
 editDocument:
   en: Edit document metadata
   de: Dokument-Metadaten bearbeiten
-roomRequired:
-  en: Please specify a room
-  de: Bitte gib einen Raum an
```
ahelmberger, line 12: Did you delete this validation text on purpose (because the label of the field is more or less the same)?

src/domain/schemas/document-schemas.js
```diff
@@ -15,17 +15,31 @@ export const patchDocSectionsBodySchema = joi.object({
...
+  allowedOpenContribution: joi.string().valid(...Object.values(DOCUMENT_ALLOWED_OPEN_CONTRIBUTION)).required()
+});
+
+export const publicContextConditionalSchema = joi.alternatives().conditional(
+  'roomId',
+  {
+    is: idOrKeySchema,
+    then: null,
+    otherwise: publicContextSchema.required()
+  }
```
ahelmberger, line 31: For some reason I would have expected the inverted test:

~~~
{
  is: null,
  then: publicContextSchema.required(),
  otherwise: null
}
~~~

But it's probably a matter of taste. Consider this comment purely informational and non-passionate :smile:

src/components/document-metadata-modal.yml
adrianaluchian: Initially I even thought that the label was wrong because it was phrased like a validation message 😅 So I reduced them.

src/domain/schemas/document-schemas.js
adrianaluchian: This is just temporary, it will make sense when we add the schema for `roomContext`, in which case `is: idOrKeySchema` is the setup that makes more sense.

adrianaluchian:
> The "new" dialog is much cleaner, great work! 😍
> 
> I found one deviation from the former behavior though, that we have to correct:
> 
> When cloning a document, the old implementation resets the four properties that make up the `publicContext` to their default values, while the reworked solution keeps the original values from the source document (at least as long as the cloning strategy dropdown stays unchanged). This is problematic as it:
> 
> * prevents non-maintainers from cloning+editing any document with restricted `allowedOpenContributions`
> * enables "stealing" of the `verified` status by simply cloning and modifying an already verified document
> 
> In other words: in the `resetStates` function we have to disregard the `initialDocumentMetadata.publicContext` values when the dialog was opened in cloning mode.

Very good catch, thank you! I've fixed it.

src/components/document-metadata-modal.js
adrianaluchian: Cleaner maybe no, but definitely safer! Cloned it, thanks.

src/domain/schemas/document-schemas.js
adrianaluchian: Actually it's been long enough that I forgot the code. This is a dedicated schema for public context only.. In which case what you suggested is more logical. I am changing this already in the follow-up ticket.

ahelmberger: Poifect!

src/components/document-metadata-modal.yml
ahelmberger: So, as we agreed, we'll bring back this validation message and rephrase the label to

* eng: "Room"
* ger: "Raum"

adrianaluchian: Aye aye, done!

ahelmberger: Cool, thanks!

ahelmberger (approved):

PR #719, opened 2022-11-18
EDU-943 Fix list styling in markdown

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-943

I also decreased the left padding for each list level in general from around 3em to 2.5em which I find looks better.

Before:

before

After:

after

### DIFF
--- a/src/components/markdown.less
+++ b/src/components/markdown.less
@@ -65,6 +65,12 @@
     font-weight: normal;
   }
 
+  ol, dl, ul {
+    overflow: hidden;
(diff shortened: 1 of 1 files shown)


adrianaluchian:
> Closes https://educandu.atlassian.net/browse/EDU-943
> 
> I also decreased the left padding for each list level in general from around 3em to 2.5em which I find looks better.
> ...

Looks much better!

adrianaluchian (approved):

PR #720, opened 2022-11-25
EDU-944 Make the image span the available width

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-944


adrianaluchian:
Can you please update the ticket with explaining the different decision on how to fix the problem? The current text is describing a different solution.

ahelmberger:
> Can you please update the ticket with explaining the different decision on how to fix the problem? The current text is describing a different solution.

Thanks, good point. Updated!

adrianaluchian (approved):

PR #725, opened 2022-12-07
EDU-939 Only re-render sections when the content has changed

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-939

### DIFF
--- a/src/ui/hooks.js
+++ b/src/ui/hooks.js
@@ -99,3 +99,9 @@ export function useDedupedCallback(callback) {
   });
   return obj.current.wrapper;
 }
+
+export function useStableCallback(callback) {
+  const callbackRef = useRef();
+  callbackRef.current = callback;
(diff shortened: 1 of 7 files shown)


src/ui/hooks.js
```diff
@@ -99,3 +99,9 @@ export function useDedupedCallback(callback) {
   });
   return obj.current.wrapper;
 }
+
+export function useStableCallback(callback) {
+  const callbackRef = useRef();
+  callbackRef.current = callback;
+  return useCallback((...args) => callbackRef.current?.(...args), [callbackRef]);
+}
```
adrianaluchian, line 107: Wonderful find and solution!! 👏 So the callbacks were the issue, not the actual content change!

adrianaluchian (approved):

src/ui/hooks.js
ahelmberger: Thanks! :bow:

PR #726, opened 2022-12-07
EDU-951 Partial fix for youtube video auto-play scenario

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-951

After spending more than one day on this issue, I have currently drawn the line at transforming the bug into a different bug we could live with until we want to invest more time in this.

The situation is as follows:
- when a media plugin section sets a youtube URL as its `sourceUrl`, then plays and pauses it, then moves another section up, before it, the `react-player` starts disregarding the `false` value we pass to its `playing` param and starts auto-playing the video
- it seems the `react-player` ends up in an incorrect state from this point on
- using a `ref` for the value we pass to the `playing` param of the `react-player` does not solve the issue of its state change being disregarded in the described scenario
- manually pausing the video has the effect of stopping the auto-play; however the `playing` state change continues to not function correctly passed this point
- there is no auto-play actually enabled on the `react-player`, at least not that I could find by drilling into the `playerRef` data

This change transforms the existing bug so that the video no longer auto plays, however once the user clicks to play from our custom controls, the media starts playing and can no longer be paused through the controls (clicking pause has no effect anymore). When a youtube video is loaded in the `video` section (as opposed to an `audio` section), playing and pausing can still be done by clicking the video directly.

### DIFF
--- a/src/components/media-player/media-player-track.js
+++ b/src/components/media-player/media-player-track.js
@@ -206,8 +206,6 @@ function MediaPlayerTrack({
   const isBufferingWhilePlaying = trackPlayState.current === MEDIA_PLAY_STATE.buffering && trackPlayState.beforeBuffering === MEDIA_PLAY_STATE.playing;
   const shouldPlay = trackPlayState.current === MEDIA_PLAY_STATE.playing || isBufferingWhilePlaying;
 
-  const thumbnailForLightMode = loadImmediately ? false : trackPlayState.current === MEDIA_PLAY_STATE.initializing && (posterImageUrl || true);
-
   return (
     <div className={classes}>
       <div className="MediaPlayerTrack-aspectRatioContainer" style={{ paddingTop }}>
@@ -222,7 +220,7 @@ function MediaPlayerTrack({
           muted={volume === 0}
           playbackRate={playbackRate}
           progressInterval={MEDIA_PROGRESS_INTERVAL_IN_MILLISECONDS}
-          light={thumbnailForLightMode}
+          light={loadImmediately ? false : trackPlayState.current === MEDIA_PLAY_STATE.initializing && (posterImageUrl || true)}
           playing={shouldPlay}
           onReady={handleReady}
           onBuffer={handleBuffer}
--- a/src/components/media-player/media-player-track.js
+++ b/src/components/media-player/media-player-track.js
@@ -26,6 +26,7 @@ function MediaPlayerTrack({
   onDuration,
   onProgress,
   onEndReached,
+  onInvalidStateReached,
   onPlayStateChange
 }) {
   const playerRef = useRef();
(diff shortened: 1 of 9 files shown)


adrianaluchian:
> I think I might have found a solution for the initial problem:
> 
> The problem is that sections do not get unmounted and mounted again when their position change (due to being rendered in a list with the section key as the `react` JSX key).
> 
> This might be a problem for all plugins that meddle directly with the DOM (not only media, but maybe also PDF viewer / MusicXML viewer etc.).
> 
> In order to force unmount/mount for those sections, all we have to do is change line 70 in `sections-display.js`:
> 
> * from `key={section.key}`
> * to `` key={`${section.key}-${index}`} ``
> 
> In my tests that eliminated the problem, because the Youtube player gets unmounted and rendered freshly.

On the one hand that's a good thing that they don't get unmounted, because their states are not reset and the page can keep its appearance/setup when moving sections (only for the interactive sections, of course). 

It's a good (and fast) find, and even though it only fixes the problem for sections in display mode and not in edit mode (where we have media preview in some cases and the autoplay still occurs under the specific setup of play+pause), it's a better solution. I will change to it then, thanks!

src/components/media-player/media-player-track.js
```diff
@@ -150,6 +150,16 @@ function MediaPlayerTrack({
     const currentSourceTimestamp = progress.played * sourceDuration;
     const trackStopTimecode = lastPlaybackRange[1] * sourceDuration;
 
+    // This unbexpected case only occurs when the react-player rerenders, having
+    // loaded, played and paused a youtube resource before that.
+    // In this case the react-player automatically plays, disregarding the value passed to the "playing" prop.
+    // This solution stops the auto-play on re-render, however the react-player's internal state is damaged
+    // from this point on, the state change of the "playing" prop not being handled correctly
+    if (trackPlayState.current === MEDIA_PLAY_STATE.pausing && progress.played === 0) {
```
adrianaluchian: Just to put a pin in the conclusion we reached regarding this topic:
- this check is our hook into identifying when the bug occurs, based on which we can act somehow (possibly by re-rendering the react-player).

src/components/pages/doc.js
```diff
@@ -97,6 +97,7 @@ function Doc({ initialState, PageTemplate }) {
   const [doc, setDoc] = useState(initialState.doc);
   const [comments, setComments] = useState([]);
   const [historyRevisions, setHistoryRevisions] = useState([]);
+  const [editedSectionKeys, setEditedSectionKeys] = useState([]);
```
ahelmberger, line 100: I left this in from an earlier commit. The edited state was the only section state (apart from internal plugin editors) that was not pulled up to the page level before. Now it is handled similar to the invalid and pending states of a section.

ahelmberger:
I updated the PR so that the whole media player component re-mounts itself when an invalid state is reached (according to https://github.com/educandu/educandu/pull/726#discussion_r1045503841) and it works in the sense that the player does not transition into an uncontrollable state or starts playing automatically anymore.

There might be the same problem in `matching-cards-tile.js`, where we use `react-player` directly, however as the cards play automatically and cannot be paused, this does not manifest in weird behavior at the moment. In case we change this behavior in the future, we will have to check for this specific scenario.

src/components/pages/revision.js
```diff
@@ -7,22 +7,12 @@ import { documentRevisionShape } from '../../ui/default-prop-types.js';
...
       <div className="RevisionPage">
-        <SectionsDisplay
-          sections={revision.sections}
-          publicStorage={publicStorage}
-          canEdit={false}
-          />
+        <SectionsDisplay sections={revision.sections} />
+        <CreditsFooter revision={revision} />
       </div>
-      <CreditsFooter revision={revision} />
```
ahelmberger, line 25: This file was just cleaned up, no connection to the actual ticket :stuck_out_tongue:

adrianaluchian: Cool 👍

adrianaluchian:
> I updated the PR so that the whole media player component re-mounts itself when an invalid state is reached (according to [#726 (comment)](https://github.com/educandu/educandu/pull/726#discussion_r1045503841)) and it works in the sense that the player does not transition into an uncontrollable state or starts playing automatically anymore.
> 
> There might be the same problem in `matching-cards-tile.js`, where we use `react-player` directly, however as the cards play automatically and cannot be paused, this does not manifest in weird behavior at the moment. In case we change this behavior in the future, we will have to check for this specific scenario.

Nice! 🙌

PR #727, opened 2022-12-08
EDU-949 Use high res Youtube thumbnail if available

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-949

### DIFF
--- a/src/components/media-player/media-duration-cache.js
+++ b/src/components/media-player/media-duration-cache.js
@@ -5,6 +5,8 @@ import { getDisposalInfo, DISPOSAL_PRIORITY } from '../../common/di.js';
 
 const logger = new Logger(import.meta.url);
 
+const FAILED_ENTRIES_LIFETIME = 5000;
+
 class MediaDurationCache {
   static get inject() { return []; }
 
@@ -74,7 +76,8 @@ class MediaDurationCache {
       this._handleDurationDetermined(sourceUrl, duration, null);
     } catch (error) {
       logger.error(error);
-      this._handleDurationDetermined(null, null, error);
+      this._handleDurationDetermined(sourceUrl, null, error);
+      this._registerForDeletion(sourceUrl);
     }
   }
 
@@ -93,6 +96,15 @@ class MediaDurationCache {
         logger.error(error);
       }
     }
+  }
+
+  _registerForDeletion(sourceUrl) {
+    window.setTimeout(() => {
+      if (!this._isDisposed) {
+        this._entries.delete(sourceUrl);
+        this._notifySubscribers();
+      }
+    }, FAILED_ENTRIES_LIFETIME);
   }
 
   _throwIfDisposed() {
--- a/src/components/media-player/media-duration-cache.js
+++ b/src/components/media-player/media-duration-cache.js
@@ -5,7 +5,7 @@ import { getDisposalInfo, DISPOSAL_PRIORITY } from '../../common/di.js';
 
 const logger = new Logger(import.meta.url);
 
-const FAILED_ENTRIES_LIFETIME = 5000;
+const FAILED_ENTRIES_LIFETIME_IN_MS = 5000;
 
 class MediaDurationCache {
   static get inject() { return []; }
@@ -104,7 +104,7 @@ class MediaDurationCache {
         this._entries.delete(sourceUrl);
         this._notifySubscribers();
       }
-    }, FAILED_ENTRIES_LIFETIME);
+    }, FAILED_ENTRIES_LIFETIME_IN_MS);
   }
 
   _throwIfDisposed() {
--- a/src/components/media-player/media-hooks.js
+++ b/src/components/media-player/media-hooks.js
@@ -2,6 +2,7 @@ import deepEqual from 'fast-deep-equal';
 import { useEffect, useState } from 'react';
 import { useService } from '../container-context.js';
 import MediaDurationCache from './media-duration-cache.js';
+import YoutubeThumbnailUrlCache from './youtube-thumbnail-url-cache.js';
 
 export function useMediaDurations(urls) {
   const [, setSemaphore] = useState(0);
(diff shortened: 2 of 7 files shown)


src/components/media-player/media-duration-cache.js
```diff
@@ -5,6 +5,8 @@ import { getDisposalInfo, DISPOSAL_PRIORITY } from '../../common/di.js';
 
 const logger = new Logger(import.meta.url);
 
+const FAILED_ENTRIES_LIFETIME = 5000;
```
adrianaluchian, line 8: In milliseconds, I guess. Maybe good to specify in name (`_IN_MS`) 🤔

src/utils/media-utils.js
```diff
@@ -113,6 +117,18 @@ export const determineMediaDuration = memoizee(async url => {
...
+export const verifyYoutubeThumbnailUrl = memoizee(url => {
+  const imagePromise = new Promise((resolve, reject) => {
+    preloadImage(url).then(success => resolve(success)).catch(error => reject(error));
+  });
+
+  const timeoutPromise = new Promise((_resolve, reject) => {
+    setTimeout(() => reject(new Error(`Timeout verifying thumbnail URL ${url}`)), MEDIA_TIMEOUT_IN_MS);
+  });
+
+  return Promise.race([imagePromise, timeoutPromise]);
```
adrianaluchian, line 129: Nice!

src/utils/media-utils.js
```diff
@@ -113,6 +117,18 @@ export const determineMediaDuration = memoizee(async url => {
   return Promise.race([playerPromise, timeoutPromise]);
 }, { promise: true });
 
+export const verifyYoutubeThumbnailUrl = memoizee(url => {
+  const imagePromise = new Promise((resolve, reject) => {
```
adrianaluchian, line 121: I'm missing why this need to be wrapped in another promise.

src/utils/media-utils.js
```diff
@@ -113,6 +117,18 @@ export const determineMediaDuration = memoizee(async url => {
   return Promise.race([playerPromise, timeoutPromise]);
 }, { promise: true });
 
+export const verifyYoutubeThumbnailUrl = memoizee(url => {
```
adrianaluchian, line 120: Maybe `verifyMediaUrl` (due to the use of `MEDIA_TIMEOUT_IN_MS`) would be better? Since this method doesn't know anything of youtube actually?

src/utils/media-utils.js
```diff
@@ -27,6 +28,7 @@ export function analyzeMediaUrl(url) {
     return {
       sanitizedUrl: `https://www.youtube.com/watch?v=${encodeURIComponent(videoId)}`,
       isYoutube: true,
+      youtubeVideoId: videoId,
```
adrianaluchian, line 31: I'm only partial to this, but what would you think of removing `isYoutube` in this case and having `youtubeVideoId` as the sole indicator?

src/components/media-player/youtube-thumbnail-url-cache.js
```diff
@@ -0,0 +1,118 @@
...
+    };
+  }
+
+  async _tryVerifyHighResThumbnailUrl(entry) {
+    if (!isBrowser()) {
+      return;
+    }
+
+    try {
+      const sucess = await verifyYoutubeThumbnailUrl(entry.highResThumbnailUrl);
```
adrianaluchian, line 72: This guy needs another "c" to be truly successful 😋

src/components/media-player/media-hooks.js
```diff
@@ -23,3 +24,24 @@ export function useMediaDurations(urls) {
...
+export function useYoutubeThumbnailUrl(url) {
+  const [, setSemaphore] = useState(0);
+  const cache = useService(YoutubeThumbnailUrlCache);
+  const [lastReturnedValue, setLastReturnedValue] = useState(null);
+
+  useEffect(() => {
+    const callback = () => setSemaphore(oldValue => oldValue + 1);
+    cache.subscribe(callback);
+    return () => cache.unsubscribe(callback);
+  }, [cache, setSemaphore]);
```
adrianaluchian, line 37: I think `setSemaphore` does not need to be a dependency here, can be removed, right?

src/components/media-player/media-player-track.js
```diff
@@ -30,6 +31,7 @@ function MediaPlayerTrack({
   const playerRef = useRef();
   const playerHandlersRef = useRef({});
   const [sourceDuration, setSourceDuration] = useState(0);
+  const youtubeThumbnailUrl = useYoutubeThumbnailUrl(sourceUrl);
```
adrianaluchian, line 34: Maybe it's better to `analyzeMediaUrl(sourceUrl)` directly inside `useYoutubeThumbnailUrl` and do an early return of `null` in case the source is not Youtube.

I find it confusing to always have a `youtubeThumbnailUrl` value even though the source could be internal.

src/utils/media-utils.js
ahelmberger: I went with `verifyMediaThumbnailUrl` to not suggest it would be video or audio.

src/components/media-player/youtube-thumbnail-url-cache.js
ahelmberger: :grin: I helped the poor guy.

src/components/media-player/media-hooks.js
ahelmberger: I was under the impression that we usually had the setters we use in the dependency list, but after scanning through the code I realized that we don't. Removed!

src/utils/media-utils.js
ahelmberger: Thanks for spotting it, it's a leftover and is now removed.

src/components/media-player/media-duration-cache.js
ahelmberger: Sure, renamed.

src/utils/media-utils.js
ahelmberger: I thought the same and was not sure if it's worth it, but as you have the same opinion, I'll go for it.

ahelmberger: Removed (and found out that it was not used anymore anyways).

adrianaluchian (approved):

src/components/media-player/media-player-track.js
ahelmberger: True. I left the check in the `youtube-thumbnail-cache` though, rather then putting it into the hook, but it returns `null` for media other than Youtube.

src/components/media-player/media-player-track.js
```diff
@@ -213,7 +213,7 @@ function MediaPlayerTrack({
     lightModeValue = false;
   } else if (posterImageUrl) {
     lightModeValue = posterImageUrl;
-  } else if (youtubeThumbnailUrl.lowResThumbnailUrl) {
+  } else if (youtubeThumbnailUrl?.lowResThumbnailUrl) {
```
adrianaluchian, line 216: Tiny thing: makes sense to only check for `youtubeThumbnailUrl` here now, right?

ahelmberger: Ahh so true, after this change, checking for `youtubeThumbnailUrl` is enough of course. Changed! Thanks!

PR #728, opened 2022-12-08
EDU-930 Update nodejs to v18

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-930

### DIFF
--- a/src/bootstrap/server-bootstrapper.js
+++ b/src/bootstrap/server-bootstrapper.js
@@ -5,10 +5,10 @@ import { Container } from '../common/di.js';
 import Database from '../stores/database.js';
 import ServerConfig from './server-config.js';
 import ClientConfig from './client-config.js';
-import resources from '../resources/resources.json';
 import PageResolver from '../domain/page-resolver.js';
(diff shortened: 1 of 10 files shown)


src/bootstrap/server-bootstrapper.js
```diff
@@ -46,6 +47,12 @@ export async function createContainer(configValues = {}) {
 
   logger.info('Loading resources');
   const additionalResources = await Promise.all(serverConfig.resources.map(async modulePath => {
+    if ((modulePath.toLowerCase()).endsWith('.json')) {
+      const resourceFileUrl = new URL(modulePath, import.meta.url);
+      const resourceJSON = JSON.parse(readFileSync(resourceFileUrl));
+      return resourceJSON;
+    }
+
```
ahelmberger, line 55: As we do not have any other resources, we do not need the `if` and also not the code below this comment.

I wonder why we can still import the "main" resources with the import assertion but not the additional resources. What was the error there?

ahelmberger: So maybe vitest cannot parse dynamic imports with assertions (because static imports seem to work fine). Thanks.

PR #729, opened 2022-12-12
EDU-934 Update react to latest version

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-930


src/components/media-player/media-player-track.js
```diff
@@ -209,39 +217,41 @@ function MediaPlayerTrack({
...
-          onBufferEnd={handleBufferEnd}
-          onStart={handlePlay}
-          onPlay={handlePlay}
-          onPause={handlePause}
-          onEnded={handleEnded}
-          onDuration={handleDuration}
-          onProgress={handleProgress}
-          onClickPreview={handleClickPreview}
-          />
+        {!!isMounted.current && (
```
adrianaluchian, line 220: We need to do this because the react player renders into having a difference between server side and client side 😬

ahelmberger: :grimacing: Let's just hope there are not even worse breaking changes

test-app/src/bundles/home-page-illustration.js
```diff
@@ -12,7 +12,7 @@ function HomePageIllustration() {
       width="300"
       height="143"
       xmlns="http://www.w3.org/2000/svg"
-      xmlns:svg="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/2000/svg"
```
ahelmberger, line 15: This would be rendered as : `xmlns:xlink="http://www.w3.org/1999/xlink"` (a namespace which is not used in this file). Line 15 can simply be deleted as `http://www.w3.org/2000/svg` is already declared the default namespace in line 14. It was probably a leftover from some XML to JSX conversion.

adrianaluchian: Of course, a copy+paste error, thanks for spotting. Will remove completely.

PR #736, opened 2022-12-13
EDU-971 Remove JSON imports and obsolete code

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-971


.eslintignore
```diff
@@ -1,11 +1,2 @@
...
-# Following files will be removed from this list in https://educandu.atlassian.net/browse/EDU-936
-src/index.js
-src/test-helper.js
-src/data/language-name-provider.js
-src/bootstrap/server-bootstrapper.js
-src/components/history-control-panel.js
-src/components/localization/language-select.js
-src/components/localization/language-flag-and-name.js
+/dist/**
+/test-app/dist/**
```
ahelmberger, line 2: I recently learnt that in `.gitignore` and similar files the slash at the beginning means: starting from the directory the ignore file is in. Without the slash it matches any part of the path. So I'd propose to be more specific in the future like above.

gulpfile.js
```diff
@@ -133,13 +130,6 @@ export async function buildTranslations() {
 
 export const buildTestApp = gulp.parallel(buildTestAppCss, buildTranslations, buildTestAppJs);
 
-export async function countriesUpdate() {
-  await Promise.all(supportedLanguages.map(lang => downloadJson(
-    `https://raw.githubusercontent.com/umpirsky/country-list/master/data/${encodeURIComponent(lang)}/country.json`,
-    `./src/data/country-names/${lang}.json`
-  )));
-}
-
```
ahelmberger, line 142: The countries were only used in the old user profile that we replaced with the new user page.

src/repositories/cdn.spec.js
```diff
@@ -1,49 +0,0 @@
...
-  afterAll(async () => {
-    await destroyTestEnvironment(container);
-    await deleteTestDir(testDir);
-  });
-
-  afterEach(async () => {
-    await pruneTestEnvironment(container);
-  });
-
-  describe('uploadObject', () => {
```
ahelmberger, line 30: The usefulness of this test is pretty limited as it tests Minio anyways.

src/components/history-control-panel.js
```diff
@@ -89,7 +89,7 @@ function HistoryControlPanel({
         >
         {t('permalink')}
       </Button>
-      {canRestoreRevisions && (
+      {!!canRestoreRevisions && (
```
adrianaluchian, line 92: Funny how these were not already changed with the switch to node18? 🤔 All other such cases were caught by the linter then.

adrianaluchian (approved):

src/components/history-control-panel.js
ahelmberger: It was in the ignored list in `.eslintrc` before :-)

PR #738, opened 2022-12-15
EDU-935 Update antd to latest version

adrianaluchian:
[Hopefully nicely finally] closes https://educandu.atlassian.net/browse/EDU-935.


src/bootstrap/server-bootstrapper.js
```diff
@@ -60,6 +62,20 @@ export async function createContainer(configValues = {}) {
   logger.info('Registering resource manager');
   container.registerInstance(ResourceManager, resourceManager);
 
+  logger.info('Loading theme files');
+  const globalVariablesFileUrl = new URL(path.resolve('./src/styles/global-variables.less'), import.meta.url);
```
ahelmberger, line 66: There's no need for using an URL as we only feed it to `fs.readFile(Sync)`, so I think it would be better to resolve the path to the global variables relative to the current module, basically the same way as we do for the resources above (line 20):

~~~
const globalVariablesFilePath = path.resolve(thisDir, '../styles/global-variables.less');
~~~

src/bootstrap/server-bootstrapper.js
```diff
@@ -60,6 +62,20 @@ export async function createContainer(configValues = {}) {
...
+  let variables = await lessVariablesToJson(readFileSync(globalVariablesFileUrl, 'utf8'));
+
+  if (serverConfig.themeFile) {
+    const themeFileUrl = new URL(serverConfig.themeFile, import.meta.url);
```
ahelmberger, line 70: Also here there's no need for a URL. `serverConfig.theme` (as all file paths coming from server config) is already an absolute file path and can be used by `readFile(Sync)` as is. The URL construction has no effect, or is there something I don't see?

src/bootstrap/server-bootstrapper.js
```diff
@@ -1,13 +1,15 @@
...
 import path from 'node:path';
-import { promises as fs } from 'node:fs';
 import Cdn from '../repositories/cdn.js';
 import Logger from '../common/logger.js';
 import { Container } from '../common/di.js';
 import Database from '../stores/database.js';
 import ServerConfig from './server-config.js';
 import ClientConfig from './client-config.js';
 import PageResolver from '../domain/page-resolver.js';
+import { promises as fs, readFileSync } from 'node:fs';
```
ahelmberger, line 10: For the other file reading matters we use the asynchronous version in this module (`fs.readFile`), I'd propose to not use the blocking `...Sync` versions.

src/bootstrap/client-bootstrapper.js
```diff
@@ -46,6 +46,7 @@ export async function hydrateApp({ bundleConfig }) {
     pageName: window.__pageName__,
     container,
     initialState: window.__initalState__,
+    theme: window.__theme__,
```
ahelmberger, line 49: We should also, somewhere above in this file, register the `ThemeManager` with the container (like with the `ResourceManager` above), so it has the correct values set when it is requested from the container (I know it currently isn't in client-running code yet, but it could):

~~~
const themeManager = new ThemeManager(window.__theme__);
container.registerInstance(ThemeManager, themeManager);
~~~

src/components/pages/users.js
```diff
@@ -80,7 +72,6 @@ function Users({ initialState, PageTemplate }) {
   const [filterText, setFilterText] = useState('');
   const [usersById, setUsersById] = useState(new Map());
   const [internalUsers, setInternalUsers] = useState([]);
-  const [externalUsers, setExternalUsers] = useState([]);
```
ahelmberger, line 83: Good find :-)

src/bootstrap/client-bootstrapper.js
adrianaluchian: Yes of course, registered.

src/bootstrap/server-bootstrapper.js
adrianaluchian: Did it the same, thanks!

src/bootstrap/server-bootstrapper.js
adrianaluchian: Just copied the same file reading setup from loading `additionalResources`, but I updated both now. And of course reading them async 🤦‍♀️

src/bootstrap/server-bootstrapper.js
```diff
@@ -49,17 +52,19 @@ export async function createContainer(configValues = {}) {
   container.registerInstance(Cdn, cdn);
 
   logger.info('Loading resources');
-  const additionalResources = await Promise.all(serverConfig.resources.map(async modulePath => {
-    const resourceFileUrl = new URL(modulePath, import.meta.url);
-    const resourceJSON = await fs.readFile(resourceFileUrl, 'utf8').then(JSON.parse);
-    return resourceJSON;
-  }));
-
+  const additionalResources = await Promise.all(serverConfig.resources.map(modulePath => fs.readFile(modulePath, 'utf8').then(JSON.parse)));
```
ahelmberger, line 55: Yeah :laughing: :+1:

adrianaluchian:
> Great (and huge) job on the update! Looks good and seems to be more performant actually. There's one thing I realized that looks a little bit odd: In the search result table we have a 1px horizontal line between the rows, but the row with the hover on it doesn't have the line (neither on top nor on the bottom), so it looks a little bit confusing when moving with the mouse over the table. Is that also on your browser?

Thanks 😊 🙏 
Yes, that table hover (d)effect the one thing that I could not fix nicely and that AntD introduced. Should I look into overwriting the ant classes to remove this?

src/bootstrap/client-bootstrapper.js
adrianaluchian: Sorry, I hadn't seen the edited version of this comment.

ahelmberger: Maybe I didn't get it right. The `ThemeManager` constructor expects an object with the less variables and constructs a theme object out of them. But on the client, in `window.__theme__` we already have the constructed theme. In the code you don't seem to distinguish :thinking:

If I get the instance on the browser like this:

~~~
  const themeManager = useService(ThemeManager);
  console.log(themeManager.getTheme());
~~~

... I get:

~~~
number-utils.js:2 Uncaught TypeError: Cannot read properties of undefined (reading 'match')
    at getNumberFromString (number-utils.js:2:26)
~~~

ahelmberger:
> > Great (and huge) job on the update! Looks good and seems to be more performant actually. There's one thing I realized that looks a little bit odd: In the search result table we have a 1px horizontal line between the rows, but the row with the hover on it doesn't have the line (neither on top nor on the bottom), so it looks a little bit confusing when moving with the mouse over the table. Is that also on your browser?
> 
> Thanks blush pray Yes, that table hover (d)effect is the one thing that I could not fix nicely and that AntD introduced. Should I look into overwriting the ant classes to remove this?

Ah, I see. I'm not sure if it only confused me because it's different than before or because it looks weird in general. If you find an override and it's not too much effort and hacking, it would be nice to keep those lines also when hovering.

src/bootstrap/client-bootstrapper.js
ahelmberger: Another deadlock :laughing:

adrianaluchian: Oh yes, I see, that was completely wrong what I had done there actually. Corrected it, thanks!

adrianaluchian:
> > > Great (and huge) job on the update! Looks good and seems to be more performant actually. There's one thing I realized that looks a little bit odd: In the search result table we have a 1px horizontal line between the rows, but the row with the hover on it doesn't have the line (neither on top nor on the bottom), so it looks a little bit confusing when moving with the mouse over the table. Is that also on your browser?
> > 
> > 
> > Thanks blush pray Yes, that table hover (d)effect is the one thing that I could not fix nicely and that AntD introduced. Should I look into overwriting the ant classes to remove this?
> 
> Ah, I see. I'm not sure if it only confused me because it's different than before or because it looks weird in general. If you find an override and it's not too much effort and hacking, it would be nice to keep those lines also when hovering.

I shall attempt and see what I find then. It does look weird for me too of course but I wasn't sure if I should try and go so much against the AntD styling.

src/bootstrap/client-bootstrapper.js
ahelmberger: :+1:

ahelmberger:
> > > > > Great (and huge) job on the update! Looks good and seems to be more performant actually. There's one thing I realized that looks a little bit odd: In the search result table we have a 1px horizontal line between the rows, but the row with the hover on it doesn't have the line (neither on top nor on the bottom), so it looks a little bit confusing when moving with the mouse over the table. Is that also on your browser?
> > > > 
> > > > 
> > > > Thanks blush pray Yes, that table hover (d)effect is the one thing that I could not fix nicely and that AntD introduced. Should I look into overwriting the ant classes to remove this?
> > > 
> > > 
> > > Ah, I see. I'm not sure if it only confused me because it's different than before or because it looks weird in general. If you find an override and it's not too much effort and hacking, it would be nice to keep those lines also when hovering.
> > 
> > 
> > I shall attempt and see what I find then. It does look weird for me too of course but I wasn't sure if I should try and go so much against the AntD styling.
> 
> Fixed! sunglasses I also removed some pre-existing overrides to the table header to have them all look similar across out tables.

Cool :heart_eyes:

ahelmberger (approved):

PR #739, opened 2022-12-15
EDU-962 Streamline linting and build

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-962


src/plugins/table/table-utils.spec.js
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable max-lines */
```
adrianaluchian, line 1: Good. I have always disliked this rule. The decision to split a file should be based on logic not on a counter 😬

adrianaluchian:
Nice (and extensive) streamlining! 👏

adrianaluchian (approved):

PR #743, opened 2022-12-20
EDU-981 Rerender media player in specific case

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-981


src/components/media-player/media-player-track.js
```diff
@@ -166,7 +166,18 @@ function MediaPlayerTrack({
     // In this case the react-player automatically plays, disregarding the value passed to the "playing" prop.
     // Therefore we have to inform the parent player about the internal state damage and reload the whole player.
     if (trackPlayState.current === MEDIA_PLAY_STATE.pausing && progress.played === 0) {
-      playerRef.current.getInternalPlayer()?.pauseVideo();
+      playerRef.current.getInternalPlayer()?.pauseVideo?.();
```
adrianaluchian, line 169: Apparently there is no `pauseVideo` method in the internal player when it's not a youtube source 😬

ahelmberger: Ahhh, of course!

ahelmberger (approved):

PR #745, opened 2022-12-20
EDU-977 Support setting X-Frame-Options header

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-977


src/server/page-renderer.js
```diff
@@ -90,13 +90,19 @@ class PageRenderer extends PageRendererBase {
...
+    }
+
     return res
       .type('html')
-      .set({
-        'Cache-Control': 'max-age=0, no-cache, no-store, must-revalidate',
-        'Pragma': 'no-cache',
-        'Expires': 'Wed, 11 Jan 1984 05:00:00 GMT'
-      })
+      .set(headers)
```
ahelmberger, line 105: We should do that probably also on the `error-page-renderer.js`, I guess.

adrianaluchian: Good catch! Done.

ahelmberger (approved):

PR #746, opened 2022-12-20
EDU-985 Fix log out (due to passport changes in latest version)

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-985

### DIFF
--- a/src/server/user-controller.js
+++ b/src/server/user-controller.js
@@ -112,16 +112,10 @@ class UserController {
     return this.pageRenderer.sendPage(req, res, PAGE_NAME.login, {});
   }
 
-  async handleGetLogoutPage(req, res, next) {
+  async handleGetLogoutPage(req, res) {
     if (req.isAuthenticated()) {
-      const logout = util.promisify(req.logout);
-
-      try {
-        await logout();
-        res.clearCookie(this.serverConfig.sessionCookieName);
-      } catch (err) {
-        return next(err);
-      }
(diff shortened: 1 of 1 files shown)


src/server/user-controller.js
```diff
@@ -111,10 +112,16 @@ class UserController {
...
-      req.logout();
-      res.clearCookie(this.serverConfig.sessionCookieName);
+      const logout = util.promisify(req.logout);
+
+      try {
+        await logout();
+        res.clearCookie(this.serverConfig.sessionCookieName);
+      } catch (err) {
+        return next(err);
+      }
```
ahelmberger, line 124: All the handlers in educandu are automatically wrapped into a promise calling `next` in case of an error (done by the `express-async-errors` package.

So it's enough to change it like (we should actually never call `next` in a handler, only in middlewares):

~~~
  async handleGetLogoutPage(req, res) {
    if (req.isAuthenticated()) {
      await util.promisify(req.logout)();
      res.clearCookie(this.serverConfig.sessionCookieName);
    }

    return res.redirect(routes.getDefaultLogoutRedirectUrl());
  }
~~~

Looks like `passport.js` has changed the `logout` function from synchronous to asynchronous and therefore demands a callback now (probably not really documented clearly).

adrianaluchian: Interesting. Will reduce the `next` call, thanks 👍 Yeah, by "documented clearly" you mean "not documented", right? :D

PR #747, opened 2022-12-20
EDU-976 Prevent password brute forcing

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-976

### DIFF
--- a/src/components/confirmation-dialogs.js
+++ b/src/components/confirmation-dialogs.js
@@ -283,8 +283,8 @@ export function reloginAfterSessionExpired(modal, t, onOk, onCancel) {
   const formRef = createRef();
 
   let dialog = null;
-  let isFrozen = false;
   let isLoggingIn = false;
+  let isLoginFrozen = false;
   let createDialogProps = null;
 
   const handleLoginStarted = () => {
@@ -306,7 +306,7 @@ export function reloginAfterSessionExpired(modal, t, onOk, onCancel) {
   };
 
   const handleLoginFailedTooOften = () => {
-    isFrozen = true;
+    isLoginFrozen = true;
     dialog.update(createDialogProps());
   };
 
@@ -340,7 +340,7 @@ export function reloginAfterSessionExpired(modal, t, onOk, onCancel) {
     onCancel,
     okButtonProps: {
       loading: isLoggingIn,
-      disabled: isFrozen
+      disabled: isLoginFrozen
     }
   });
 
@@ -351,8 +351,8 @@ export function confirmWithPassword(modal, t, email, onOk, onCancel = () => {}) 
   const formRef = createRef();
 
   let dialog = null;
-  let isFrozen = false;
   let isLoggingIn = false;
+  let isLoginFrozen = false;
   let createDialogProps = null;
 
   const handleLoginStarted = () => {
@@ -374,7 +374,7 @@ export function confirmWithPassword(modal, t, email, onOk, onCancel = () => {}) 
   };
 
   const handleLoginFailedTooOften = () => {
-    isFrozen = true;
+    isLoginFrozen = true;
     dialog.update(createDialogProps());
   };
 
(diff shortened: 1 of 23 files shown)


src/server/admin-controller.js
```diff
@@ -118,6 +123,12 @@ class AdminController {
       needsPermission(permissions.ADMIN),
       (req, res) => this.handlePostCdnUploadDirectoryCreationRequest(req, res)
     );
+
+    router.get(
+      '/api/v1/admin/request-info',
+      needsPermission(permissions.ADMIN),
+      (req, res) => this.handleGetRequestInfoRequest(req, res)
+    );
```
ahelmberger, line 131: This new endpoint is there to find out the number of proxies the application is running behind on production (see also the  explanation in the ticket: https://educandu.atlassian.net/browse/EDU-976)

src/services/request-limit-record-service.js
```diff
@@ -0,0 +1,50 @@
+import Logger from '../common/logger.js';
+import { createShortHash } from '../utils/string-utils.js';
+import RequestLimitRecordStore from '../stores/request-limit-record-store.js';
+
+const logger = new Logger(import.meta.url);
+
+const FAILED_LOGIN_ATTEMPTS_THRESHOLD = 5;
+const FAILED_LOGIN_ATTEMPTS_KEY = 'failed-login-attempts';
+const FAILED_LOGIN_ATTEMPTS_TIME_WINDOW_IN_MS = 15 * 60 * 1000;
```
adrianaluchian, line 9: I know these are used juts here, but maybe it would be handy to collect all such setup in the `constants` module. We already have similar constants there like `PENDING_USER_REGISTRATION_EXPIRATION_IN_HOURS`

src/services/request-limit-record-service.js
```diff
@@ -0,0 +1,50 @@
...
+    return [RequestLimitRecordStore];
+  }
+
+  constructor(requestLimitRecordStore) {
+    this.requestLimitRecordStore = requestLimitRecordStore;
+  }
+
+  getRoomById(roomId) {
+    return this.roomStore.getRoomById(roomId);
+  }
```
adrianaluchian, line 22: This is probably a leftover.

src/services/request-limit-record-service.js
ahelmberger: Good point. Moved them to the `constants` module.

src/services/request-limit-record-service.js
ahelmberger: :scream: You should have never seen that.

src/utils/string-utils.spec.js
```diff
@@ -49,4 +49,12 @@ describe('string-utils', () => {
     });
   });
 
+  describe('createShortHash', () => {
+    it('returns the same hash code for the same input', () => {
```
adrianaluchian, line 53: I think this test can only be trusted if there's also a test that checks the negative scenario 😋

ahelmberger: The function and the test are gone now anyways :-)

src/components/confirmation-dialogs.js
```diff
@@ -283,6 +283,7 @@ export function reloginAfterSessionExpired(modal, t, onOk, onCancel) {
   const formRef = createRef();
 
   let dialog = null;
+  let isFrozen = false;
```
adrianaluchian, line 286: This variable should maybe have a clearer name since the module is not a "login" dedicated context. Maybe `isLoginFrozen` or `isLoggingInFrozen` to correlate with `isLogginIn`?

src/services/request-limit-record-service.js
adrianaluchian: 😵 🙈

src/utils/string-utils.spec.js
adrianaluchian: 👍

src/components/confirmation-dialogs.js
ahelmberger: Agree. Changed it to `isLoginFrozen`.

adrianaluchian (approved):

PR #750, opened 2022-12-21
EDU-986 Rate limit password reset API endpoints

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-986

### DIFF
--- a/src/domain/constants.js
+++ b/src/domain/constants.js
@@ -197,8 +197,10 @@ export const PENDING_PASSWORD_RESET_REQUEST_EXPIRATION_IN_HOURS = 24;
 export const PENDING_ROOM_INVITATION_EXPIRATION_IN_DAYS = 7;
 
 export const FAILED_LOGIN_ATTEMPTS_LIMIT = 5;
-export const FAILED_LOGIN_ATTEMPTS_KEY = 'failed-login-attempts';
 export const FAILED_LOGIN_ATTEMPTS_TIME_WINDOW_IN_MS = 15 * 60 * 1000;
+
+export const PASSWORD_RESET_REQUEST_LIMIT = 10;
+export const PASSWORD_RESET_REQUEST_TIME_WINDOW_IN_MS = 60 * 60 * 1000;
 
 export const NOT_ROOM_OWNER_ERROR_MESSAGE = 'The user is not the room owner';
 export const NOT_ROOM_OWNER_OR_MEMBER_ERROR_MESSAGE = 'The user is not a room owner or member';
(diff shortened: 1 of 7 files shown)


src/server/error-controller.js
```diff
@@ -36,7 +36,7 @@ class ErrorController {
 
       if (isApiCall) {
         this.sendErrorJson(res, consolidatedErr);
-      } else if (!this.tryRedirectToLogin(req, res, consolidatedErr)) {
```
ahelmberger, line 39: This is already called a few lines above

src/server/error-controller.js
```diff
@@ -70,17 +70,17 @@ class ErrorController {
   }
 
   tryRespondToExpiredSession(req, res, err) {
-    if (!req.isAuthenticated() && req.cookies[this.serverConfig.sessionCookieName]) {
+    if (err.status === HTTP_STATUS.unauthorized && !req.isAuthenticated() && req.cookies[this.serverConfig.sessionCookieName]) {
```
ahelmberger, line 73: It happened to me while testing that I got a "session expired" instead of the real underlying error. It should only assume that the user wants to re-login because of an expired session when there was a 401 before.

src/domain/constants.js
```diff
@@ -223,3 +225,11 @@ export const DEFAULT_COLOR_SWATCHES = [
   ['#fffbe6', '#fff1b8', '#ffe58f', '#ffd666', '#ffc53d', '#faad14', '#d48806', '#ad6800', '#874d00', '#613400'],
   ['#ffffff', '#fafafa', '#f5f5f5', '#e8e8e8', '#d9d9d9', '#bfbfbf', '#8c8c8c', '#595959', '#262626', '#000000']
 ];
+
+export const HTTP_STATUS = {
+  badRequest: 400,
+  unauthorized: 401,
```
ahelmberger, line 231: I took the official HTTP message wording here which is misleading, should have been "unauthenticated" actually.

adrianaluchian: I remember we talked about this before and you pointed this out 👍  I agree, we should go with the official naming though..

src/server/error-controller.js
adrianaluchian: Indeed! 😳

src/domain/rate-limit-middleware.js
```diff
@@ -0,0 +1,16 @@
+import httpErrors from 'http-errors';
+
+const { TooManyRequests } = httpErrors;
+
+function createRateLimitMiddleware({ maxRequests, expiresInMs, store }) {
+  return async function rateLimitMiddleware(req, _res, next) {
+    try {
+      const newCount = await store.incrementCount({ req, expiresInMs });
+      return newCount > maxRequests ? next(new TooManyRequests()) : next();
```
adrianaluchian, line 9: I was about to point out that in the `handlePostUserLogin` method the condition is `>= FAILED_LOGIN_ATTEMPTS_LIMIT` and here the limit is exclusive (`>`) but then I realized this middleware is not used..right? 🤔

ahelmberger: Yeah, it's a little bit tricky, but this middleware is first updating the count for each request and then deciding if it should block or not. The `handlePostUserLogin` just serves the request and then updates the count afterwards (it has to because it only updates for *failed* login attempts, not for each request). So if we set the limit to `5` for example, in both situations the user has 5 shots before it blocks the next time.

ahelmberger: Ah yes, to actually answer your question, the middleware is not used in `handlePostUserLogin`

adrianaluchian: What I meant is I cannot see where it is used at all.

adrianaluchian (approved):

PR #751, opened 2022-12-22
EDU-991 Create and use an account locked out component

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-991


src/components/irrecoverable-login-error.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useService } from './container-context.js';
+import { Trans, useTranslation } from 'react-i18next';
+import ClientConfig from '../bootstrap/client-config.js';
+
+export default function IrrecoverableLoginError({ type }) {
```
adrianaluchian, line 7: How about going with`BlockedLoginMessage` instead of `IrrecoverableLoginError`? To me all these cases are actually recoverable since it's not a system failure.

src/components/login-form.js
```diff
@@ -5,9 +5,10 @@ import { useSetUser } from './user-context.js';
 import { useTranslation } from 'react-i18next';
 import React, { useEffect, useState } from 'react';
 import { useService } from './container-context.js';
-import { HTTP_STATUS } from '../domain/constants.js';
+import { ERROR_CODES, HTTP_STATUS } from '../domain/constants.js';
```
adrianaluchian, line 8: 🤫

src/components/pages/login.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import LoginForm from '../login-form.js';
 import routes from '../../utils/routes.js';
 import { useTranslation } from 'react-i18next';
-import React, { useRef, useState } from 'react';
+import React, { Fragment, useRef, useState } from 'react';
```
adrianaluchian, line 6: 😱

gulpfile.js
```diff
@@ -41,6 +41,7 @@ const testAppEnv = {
   TEST_APP_CONSENT_COOKIE_NAME_PREFIX: 'CONSENT_TEST_APP',
   TEST_APP_UPLOAD_LIABILITY_COOKIE_NAME: 'UPLOAD_LIABILITY_TEST_APP',
   TEST_APP_X_FRAME_OPTIONS: 'SAMEORIGIN',
+  TEST_APP_ADMIN_EMAIL_ADDRESS: 'educandu-test-admin@test.com',
```
adrianaluchian, line 44: We should flag the JIRA ticket so that we remember to set and admin email in OMA 😬

src/components/irrecoverable-login-error.js
ahelmberger: That is a much better name, thanks! Will do!

src/components/login-form.js
ahelmberger: OMG, fixed

src/components/pages/login.js
ahelmberger: You can derictly derive my readiness for holidays by the number of times this happens :mx_claus:  Fixed, thanks!

gulpfile.js
ahelmberger: I will ask Ilka for an address and flag it, thanks (it's optional though, so it doesn't have to happen immediately).

src/components/irrecoverable-login-error.js
ahelmberger: Done!

adrianaluchian (approved):

PR #752, opened 2022-12-22
EDU-989 Update user data model

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-989


migrations/educandu-2022-12-22-01-remove-provider-from-users.js
```diff
@@ -0,0 +1,22 @@
+export default class Educandu_2022_12_22_01_remove_provider_from_users {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async up() {
+    await this.db.collection('users').updateMany({}, { $unset: { provider: null } });
+    await this.db.collection('users').dropIndex('_idx_verificationCode_provider_');
```
ahelmberger, line 8: We should add an index on the single property `verificationCode` again after dropping this, because we're still filtering by it.

adrianaluchian: I wanted to do that and found we already had a separate `_idx_verificationCode_` index.

src/server/user-controller.js
```diff
@@ -239,11 +239,12 @@ class UserController {
         await this.requestLimitRecordService.resetCount({ req });
       }
 
-      return req.login(user, loginError => {
+      return req.login(user, async loginError => {
         if (loginError) {
           return next(loginError);
         }
 
+        await this.userService.recordUserLogIn(user._id);
```
ahelmberger, line 247: This only updates the user in the database while the `user` object used here for the request still is the "old" version. What do you think of returning the updated user from the services end use that one instead of `user` from here on?

adrianaluchian: What do you mean by "directly on the user"?
Technically we don't use the `lastLoggedIn` date in this context at all, but makes sense to return the updated value here.

migrations/educandu-2022-12-22-01-remove-provider-from-users.js
ahelmberger: Oh, sorry, I didn't remember that. :+1:

src/server/user-controller.js
ahelmberger: Sorry, with the edit it got confusing. I try to phrase it clearer: By calling `req.login` we're basically setting `req.user`. But we call it with the not-updated user (it has the non-current `lastLoggedIn` value). I thought about it on the way home and realized that calling `req.login` basically never fails (the error handling code is just there out of formality), so I'd suggest to call `req.login` with an already updated `user` object (that maybe for example could be returned from `recordUserLogIn`. So the `req.user` reflects the current database state. (I know we do not use the property anywhere, nevertheless I think it should be the most up to date object)

src/stores/user-store.js
```diff
@@ -60,6 +60,10 @@ class UserStore {
     return this.collection.replaceOne({ _id: user._id }, user, { session, upsert: true });
   }
 
+  updateUserLastLoggedIn(userId, { session } = {}) {
+    return this.collection.updateOne({ _id: userId }, { $set: { lastLoggedInOn: new Date() } }, { session, upsert: true });
```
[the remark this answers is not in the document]
adrianaluchian: I've moved the setting of `lastLoggedInOn` where it belongs, thanks.

src/server/user-controller.js
adrianaluchian: Fully understood what you meant now, thanks for clarifying. Done it like that.

src/stores/user-store.js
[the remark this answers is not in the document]
ahelmberger: :+1:

src/server/user-controller.js
```diff
@@ -281,11 +283,19 @@ class UserController {
     return res.status(201).send({ roles: newRoles });
   }
 
-  async handlePostUserLockedOut(req, res) {
+  async handlePostUserAccountLockedOnBodySchema(req, res) {
```
ahelmberger, line 286: Oops, that renaming went wrong :thinking:

adrianaluchian: I know it looks wrong, but it actually isn't :( handle_post_userAccountLockedOn_body_schema, cause `accountLockedOn` is the field name on the `user`.

ahelmberger: But this is the handler function's name, it cannot end with `Schema` :thinking:

adrianaluchian: Oh of course it cannot! Merry Christmas 🎅 😅 
Dropped the `BodySchema` postfix, thanks!

ahelmberger: Merry ChristmasSchema :mx_claus: :rofl:

adrianaluchian: And even Christmas**Body**Schema sadly 😂

PR #754, opened 2022-12-29
EDU-1001 Generalize brute force defence over all respective endpoints

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1001


src/server/user-controller.js
```diff
@@ -222,31 +220,20 @@ class UserController {
...
   }
 
-  async handlePostUserLogin(req, res, next) {
-    const failedAttemptsCount = await this.requestLimitRecordService.getCount({ req });
-    if (failedAttemptsCount >= FAILED_LOGIN_ATTEMPTS_LIMIT) {
-      throw new TooManyRequests();
-    }
-
-    passport.authenticate('local', async (err, user) => {
+  handlePostUserLogin(req, res, next) {
```
ahelmberger, line 223: This function is basically reset to the state before https://educandu.atlassian.net/browse/EDU-976

src/server/user-controller.js
```diff
@@ -98,9 +96,9 @@ class UserController {
     this.requestLimitRecordService = requestLimitRecordService;
     this.passwordResetRequestService = passwordResetRequestService;
 
-    this.passwordResetRateLimitOptions = {
-      maxRequests: PASSWORD_RESET_REQUEST_LIMIT,
-      expiresInMs: PASSWORD_RESET_REQUEST_TIME_WINDOW_IN_MS,
+    this.antiBruteForceRequestLimitOptions = {
+      maxRequests: ANTI_BRUTE_FORCE_MAX_REQUESTS,
+      expiresInMs: ANTI_BRUTE_FORCE_EXPIRES_IN_MS,
       service: this.requestLimitRecordService
```
adrianaluchian, line 102: Actually is the `requestLimitRecordService` used anymore in this case? 🤔

ahelmberger: Yes, it is used by the middleware that is configured using these options.

ahelmberger: BUUT, good catch: I can delete 2 of the 3 service methods! Will do.

ahelmberger: Done!

adrianaluchian (approved):

PR #756, opened 2022-12-30
EDU-1002 Use plyr instead of react-player in video and audio plugins

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1002

This library also has a few glitches that I have corrected or that we might need to correct (or live with). So far though it seems to be less of a black box compared to `react-player` and the play/pause toggle seems to be reliable.
Also, all videos can now be played/paused by clicking on the video itself, so we would have consistency between youtube and CDN resources.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "passport": "^0.6.0",
     "passport-local": "^1.0.0",
     "passport-strategy": "^1.0.0",
+    "plyr": "^3.7.3",
     "pretty-bytes": "^6.0.0",
     "prop-types": "^15.8.1",
     "react": "^18.2.0",
--- a/src/components/media-player/plyr/media-playback.js
+++ b/src/components/media-player/plyr/media-playback.js
@@ -1,0 +1,239 @@
+import Plyr from 'plyr';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { useService } from '../../container-context.js';
+import { useStableCallback } from '../../../ui/hooks.js';
+import { useYoutubeThumbnailUrl } from '../media-hooks.js';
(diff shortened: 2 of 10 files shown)


ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-1002
> 
> This library also has a few glitches that I have corrected or that we might need to correct (or live with). So far though it seems to be less of a black box compared to `react-player` and the play/pause toggle seems to be reliable. Also, all videos can now be played/paused by clicking on the video itself, so we would have consistency between youtube and CDN resources.

Looks good so far, and the code is straight-forward. Of course, this is the most basic case, but I'd say we give it a shot also for the more complex media players. The only thing I stumbled upon is that on the cursor on the video (or poster) is not a hand anymore, so it doesn't look like it could be clicked. But I realized that that's also the case on YouTube itself. So maybe we should leave it that way to be consistent.

ahelmberger (approved):

adrianaluchian:
> > Closes https://educandu.atlassian.net/browse/EDU-1002
> > This library also has a few glitches that I have corrected or that we might need to correct (or live with). So far though it seems to be less of a black box compared to `react-player` and the play/pause toggle seems to be reliable. Also, all videos can now be played/paused by clicking on the video itself, so we would have consistency between youtube and CDN resources.
> 
> Looks good so far, and the code is straight-forward. Of course, this is the most basic case, but I'd say we give it a shot also for the more complex media players. The only thing I stumbled upon is that on the cursor on the video (or poster) is not a hand anymore, so it doesn't look like it could be clicked. But I realized that that's also the case on YouTube itself. So maybe we should leave it that way to be consistent.

I actually had added the hand but removed it for exactly this reason of consistency 🤔 Let's see how this works out then.

PR #758, opened 2022-12-30
EDU-999 Reduce font in matching cards on phone

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-999

### DIFF
--- a/src/plugins/matching-cards/matching-cards.less
+++ b/src/plugins/matching-cards/matching-cards.less
@@ -93,7 +93,7 @@
   text-align: center;
 
   @media @edu-media-phone {
-    font-size: 12px;
+    font-size: @edu-font-size-sm;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/plugins/matching-cards/matching-cards.less
```diff
@@ -91,6 +91,10 @@
   display: flex;
   justify-content: center;
   text-align: center;
+
+  @media @edu-media-phone {
+    font-size: 12px;
```
ahelmberger, line 96: Maybe we could use a variable instead of the hard-coded `px` value here - or - use a percentage like `75%`?

adrianaluchian: Ok, I've used`@edu-font-size-sm` (which was actually not used before).

ahelmberger (approved):

PR #759, opened 2022-12-30
EDU-997 Flip card rendering issues

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-997

### DIFF
--- a/src/plugins/matching-cards/matching-cards-display.js
+++ b/src/plugins/matching-cards/matching-cards-display.js
@@ -100,8 +100,8 @@ function MatchingCardsDisplay({ content }) {
           <MatchingCardsTile
             text={tile.text}
             sourceUrl={tile.sourceUrl}
-            playMedia={isFlipped}
             showMatched={wasMatched}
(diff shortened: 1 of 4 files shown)


src/plugins/matching-cards/matching-cards-display.js
```diff
@@ -100,8 +100,8 @@ function MatchingCardsDisplay({ content }) {
           <MatchingCardsTile
             text={tile.text}
             sourceUrl={tile.sourceUrl}
-            playMedia={isFlipped}
             showMatched={wasMatched}
+            playMedia={isFlipped && !wasMatched}
```
adrianaluchian, line 104: When 2 cards are matched we show the grey overlay and green checkmark, but the media still plays underneath. I think it's better that we stop it in this case. What do you think?

ahelmberger: Completely agree.

ahelmberger (approved):

PR #770, opened 2023-01-12
EDU-1024 Fix form in modal always showing the values of the last open state

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1024


src/components/storage-plan-modal.js
```diff
@@ -70,6 +70,7 @@ function StoragePlanModal({ isOpen, storagePlan, storagePlanNamesInUse, onOk, on
       onCancel={onCancel}
       maskClosable={false}
       okButtonProps={{ loading }}
+      forceRender
```
ahelmberger, line 73: This causes a hydration error (only when the page starts on the storage-plans` tab)

PR #776, opened 2023-01-13
EDU-1039 Replace react-player-in-media-utils

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1039

### DIFF
--- a/src/utils/media-utils.js
+++ b/src/utils/media-utils.js
@@ -1,14 +1,12 @@
 import React from 'react';
 import memoizee from 'memoizee';
-import reactPlayerNs from 'react-player';
 import ReactDOMClient from 'react-dom/client';
 import { preloadImage } from './image-utils.js';
 import { getResourceType } from './resource-utils.js';
 import { RESOURCE_TYPE } from '../domain/constants.js';
 import { getAccessibleUrl, isInternalSourceType } from './source-utils.js';
 import { getUrlValidationStatus, URL_VALIDATION_STATUS } from '../ui/validation.js';
-
-const ReactPlayer = reactPlayerNs.default || reactPlayerNs;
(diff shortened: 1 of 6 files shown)


src/utils/media-utils.js
```diff
@@ -89,22 +87,16 @@ export const determineMediaDuration = memoizee(async url => {
...
-        playing: false,
-        onDuration: durationInSeconds => {
-          const durationInMiliseconds = Math.ceil(durationInSeconds * 1000);
+      const player = React.createElement(MediaDurationIdentifier, {
+        sourceUrl: validUrl,
+        onDuration: durationInMiliseconds => {
           resolve(durationInMiliseconds);
           ensureCleanup();
-        },
-        onError: error => {
```
ahelmberger, line 101: If the media for some reasons cannot be loaded, we end up with an eternally unsettled Promise. I saw that for HTML5 players there was an `error` event in `plyr` (https://github.com/sampotts/plyr#standard-media-events). But it might be worth using a `Promise.race` with a timeout of 5 seconds or so to ensure there is some return value, watcha think?

adrianaluchian: I think it'a a good idea, so much so that you have already implemented it 😋 
```
  const timeoutPromise = new Promise((_resolve, reject) => {
    setTimeout(() => {
      reject(new Error(`Timeout determining duration of ${url}`));
      ensureCleanup();
    }, MEDIA_TIMEOUT_IN_MS);
  });
  return Promise.race([playerPromise, timeoutPromise]);
  ```
  Or did you mean something else?

ahelmberger: Oopsie, we were inside the `playerPromise here`. We will still end up with the DOM element hanging in there and not being removed, but the user's browser will tolerate a few of them :-)

ahelmberger (approved):

PR #777, opened 2023-01-13
EDU-995 Extend post-auth logic for external accounts

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-995

### DIFF
--- a/src/services/user-service.spec.js
+++ b/src/services/user-service.spec.js
@@ -73,9 +73,20 @@ describe('user-service', () => {
       });
     });
 
+    describe('when userId matches but the account is not yet confirmed', () => {
+      beforeEach(async () => {
+        user.expiresOn = new Date();
+        await updateTestUser(container, user);
+        result = await sut.findConfirmedActiveUserById({ userId: user._id });
+      });
+      it('should return null', () => {
+        expect(result).toBe(null);
+      });
+    });
+
     describe('when userId matches but the account is closed', () => {
       beforeEach(async () => {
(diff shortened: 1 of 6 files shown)


src/services/user-service.spec.js
```diff
@@ -52,6 +52,78 @@ describe('user-service', () => {
     sandbox.restore();
   });
 
+  describe('findConfirmedActiveUserById', () => {
```
adrianaluchian, line 55: We could still throw the `expires[On]` case in here, in both methods 😅

ahelmberger: I'm not sure what you mean :thinking:

adrianaluchian: The method has this part that created the bug fixed today:
```
   if (!user || user.expiresOn) {
      return null;
    }
```
with `expires` vs `expiresOn` (sorry I made it look like an array item accessor with `expires[On]`, I meant +/- the `On` part)

ahelmberger: I'm not sure if we're already taking about the same thing. But it's a good idea to add test cases for both `findConfirmedActiveUserById` and `findConfirmedActiveUserByEmailAndPassword` that check that no user with `expiresOn` is let through by the query. I'll do that :+1:

ahelmberger: I have added those test cases.

adrianaluchian: Perfect! 👌

adrianaluchian (approved):

PR #779, opened 2023-01-13
EDU-1037 Use plyr-based media player in media-range-selector

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1037

### DIFF
--- a/src/components/media-player/media-range-selector.js
+++ b/src/components/media-player/media-range-selector.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import { Modal, Button, Spin } from 'antd';
-import MediaPlayer from './media-player.js';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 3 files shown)


src/components/media-player/media-range-selector.js
```diff
@@ -13,10 +13,10 @@ function MediaRangeSelector({ sourceUrl, range, onRangeChange }) {
   const clientConfig = useService(ClientConfig);
   const { t } = useTranslation('mediaRangeSelector');
 
+  const [isModalOpen, setIsModalOpen] = useState(false);
```
ahelmberger, line 16: :partying_face:

ahelmberger (approved):

src/components/media-player/media-range-selector.js
adrianaluchian: 😆

PR #784, opened 2023-01-17
EDU-1043 Read and auto-update IDP metadata from XML

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1043


src/bootstrap/server-config.js
```diff
@@ -76,10 +76,12 @@ const configSchema = joi.object({
...
       displayName: joi.string().required(),
-      entryPoint: joi.string().required(),
-      cert: joi.string().required(),
       logoUrl: joi.string().allow(null).default(null),
-      expiryTimeoutInDays: joi.number().integer().min(1).default(6 * 30)
+      expiryTimeoutInDays: joi.number().integer().min(1).default(6 * 30),
+      metadata: joi.object({
+        url: joi.string().required(),
+        entityId: joi.string().required()
+      }).required()
```
adrianaluchian, line 84: From setup perspective (as well) this is so much better/nicer! It's contained and simplified 😍

src/components/login-form.js
```diff
@@ -134,7 +135,8 @@ export default function LoginForm({
   ];
 
   const hasBlockingError = hasLoginFailedTooOften || isUserAccountLocked;
-  const canLoginWithShibboleth = allowExternalLogin && clientConfig.samlAuth?.identityProviders.length;
+  const canLoginWithShibboleth = samlIdentityProviders.length
+    && !clientConfig.disabledFeatures.includes(FEATURE_TOGGLES.shibbolethLoginForm);
```
adrianaluchian, line 140: We need to put this into the ticket as a config update 😬 For now I just flagged the ticket so we don't forget.

src/bootstrap/server-config.js
ahelmberger: Yeah, it makes it a lot easier.

src/services/saml-config-service.js
```diff
@@ -0,0 +1,79 @@
...
+    } catch (error) {
+      if (error.code === 'ECONNABORTED') {
+        throw new Error(`Timeout trying to fetch "${url}"`);
+      } else {
+        throw error;
+      }
+    }
+  }
+
+  resolveMetadata(xmlDoc, entityId) {
```
adrianaluchian, line 47: Maybe we should wrap `resolveMetadata` and `resolveIdentityProviderMetadata` in `try`-`catch` and gracefully handle the mishap.

ahelmberger: I deliberately didn't as we mostly do not handle errors in the services. They "bubble up" (in this case the scheduled task will handle it and not update the config and just log the error).

adrianaluchian: Of course, I see.

adrianaluchian (approved):

PR #794, opened 2023-01-23
EDU-1053 Fix media playing out of range

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1053

### DIFF
--- a/src/components/media-player/html5-player.js
+++ b/src/components/media-player/html5-player.js
@@ -48,9 +48,12 @@ function Htlm5Player({
     if (player) {
       setWasPlayTriggeredOnce(true);
       const currentActualTimeInMs = player.currentTime * 1000;
-      const isEndReached = !!playbackRangeInMs[1] && currentActualTimeInMs >= playbackRangeInMs[1];
 
-      if (isEndReached) {
+      const isCurrentTimeOutsideRange
+        = (!!playbackRangeInMs[0] && currentActualTimeInMs < playbackRangeInMs[0])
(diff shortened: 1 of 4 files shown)


ahelmberger:
With these changes, all Youtube videos that have a playback range with start time code other than 0, start initially muted. Sound starts playing as soon as the first pause/stop/reset has been trigger (not when the volume changes, though). I couldn't find out why that is, because the code that sets the volume on the player is actually called, it just seems to not have any effect.

adrianaluchian:
> With these changes, all Youtube videos that have a playback range with start time code other than 0, start initially muted. Sound starts playing as soon as the first pause/stop/reset has been trigger (not when the volume changes, though). I couldn't find out why that is, because the code that sets the volume on the player is actually called, it just seems to not have any effect.

That's a really good find, thanks! 😳 At some point I stopped testing with audio thinking seeing the video play is enough to test range, but I did not expect this situation!
Seems it's the same case of `wasPlayTriggeredOnce`, where volume is not picked up. I have identified the broken scenarios in which the youtube API no longer behaves as expected and I think I managed to adjust to those cases without side-effects this time 🤞  I'm very glad the html5 player is separate given all this!

ahelmberger:
I tested all scenarios from yesterday and everything works as it should. It seems to be very stable. Thanks for all the investigations into these tricky weird behaviors, great job!!! :+1: :heart_eyes:

ahelmberger (approved):

PR #795, opened 2023-01-23
EDU-1034 Simplify registration workflow

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1034


src/components/registration-form.js
```diff
@@ -1,148 +1,285 @@
...
+  };
 
-  const register = async ({ email, password, displayName }) => {
+  const handleEnterDataFinish = async ({ email, password, displayName }) => {
     try {
-      onRegistrationStarted();
-      const { result, user: registeredUser } = await userApiClient.register({ email, password, displayName });
+      setIsLoading(true);
+      const { result, user } = await userApiClient.requestRegistration({
+        email,
```
adrianaluchian, line 106: Should we trim this guy too? 🤔 We do it at login but not here. Or we could maybe trim both on the server side actually, might be better as we lowercase the email there anyway.

src/components/registration-form.js
```diff
@@ -1,148 +1,285 @@
...
-    formRef.current = form;
-  }
+  useEffect(() => {
+    if (currentStepKey === STEP.enterData) {
+      setCreatedUser(null);
+      setForbiddenEmails([]);
+      setInvalidVerificationCodes([]);
+    }
+    setSteps(previousSteps => {
+      const newSteps = currentStepKey === STEP.enterData ? createDefaultSteps() : previousSteps.map(step => ({ ...step }));
```
adrianaluchian, line 84: For the sake of better clarity I would suggest to create a `setStepProcess` method just like `setStepError` and `setStepSuccess`. That pattern makes it so much easier to follow these state changes.

src/components/registration-form.js
```diff
@@ -1,148 +1,285 @@
...
+  const renderEnterDataForm = () => {
+    const termsPage = settings.termsPage?.[uiLanguage];
+    const agreementValidationRules = [
+      {
+        message: t('confirmTerms'),
+        validator: (_, value) => value ? Promise.resolve() : Promise.reject(new Error(t('confirmTerms')))
+      }
+    ];
+    return (
+      <div className="RegistrationForm-form">
```
adrianaluchian, line 169: Small thing and the code is old, but we could drop the `-form` in the class name, as it's the outer-most class in this component.

src/components/registration-form.js
```diff
@@ -1,148 +1,285 @@
...
+      {
+        validator: (_rule, value) => {
+          return value && invalidVerificationCodes.includes(value)
+            ? Promise.reject(new Error(t('verificationCodeIsInvalid')))
+            : Promise.resolve();
+        }
+      }
+    ];
+    return (
+      <div className="RegistrationForm-form">
```
adrianaluchian, line 221: Here as well.

src/components/registration-form.js
```diff
@@ -1,148 +1,285 @@
...
+  };
+
   return (
-    <div className="RegistrationForm">
-      {user ? registrationConfirmation : registrationForm}
+    <div className="RegistrationForm u-panel">
+      {renderSteps()}
+      {currentStepKey === STEP.enterData && renderEnterDataForm()}
+      {currentStepKey === STEP.enterCode && renderEnterCodeForm()}
+      {currentStepKey === STEP.completed && renderCompletedMessage()}
```
adrianaluchian, line 280: Nice switching mechanism with these steps and states! 👍

src/components/pages/complete-password-reset.js
```diff
@@ -63,7 +62,6 @@ function CompletePasswordReset({ initialState, PageTemplate, SiteLogo }) {
     <div className="CompletePasswordResetPage-form">
       <Form onFinish={handleFinish} scrollToFirstError>
         <PasswordFormItem name="password" {...formItemLayout} />
-        <PasswordConfirmationFormItem name="confirm" passwordFormItemName="password" {...formItemLayout} />
```
adrianaluchian, line 66: 👏 💯

src/server/user-controller.spec.js
```diff
@@ -219,6 +222,119 @@ describe('user-controller', () => {
...
+            headers: { host: 'localhost' },
+            session: { externalAccount: null },
+            body: { userId: 'cnztc5ztm41mx03z', verificationCode: '3xzrxzt43z0xtm1' },
+            login: sandbox.stub().callsArg(1)
+          });
+          res = httpMocks.createResponse({ eventEmitter: events.EventEmitter });
+
+          res.on('end', resolve);
+
+          userService.verifyUser.resolves(dbUser);
```
adrianaluchian, line 245: Just a tiny thing, but for extra safety, we could set this up as `userService.verifyUser.withArgs(req.body.userId, req.body.verificationCode).resolves(dbUser)`

adrianaluchian:
This looks absolutely wonderful now! 😍 Makes up for such a nicer experience and is much more clear for the user! Was a good idea to change and it's nicely implemented.

src/components/registration-form.js
ahelmberger: I trimmed it here as I consider it more a UI concern than the lowercasing, so the API gets already sanitized values.

src/components/registration-form.js
ahelmberger: Good idea, simplified the function accordingly.

src/components/pages/complete-password-reset.js
ahelmberger: Yeah, I liked that too :-)

src/components/registration-form.js
ahelmberger: :+1:

adrianaluchian (approved):

PR #796, opened 2023-01-24
EDU-1073 fix player reloading

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1073

### DIFF
--- a/src/components/media-player/multitrack-media-player.js
+++ b/src/components/media-player/multitrack-media-player.js
@@ -9,6 +9,20 @@ import MediaPlayerControls from './media-player-controls.js';
 import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { MEDIA_ASPECT_RATIO, MEDIA_SCREEN_MODE } from '../../domain/constants.js';
 
+const sourcesCanBeConsideredEqual = (source1, source2) => {
+  if (!source1 || !source2 || source1 === source2) {
+    return source1 === source2;
+  }
+
+  return source1.mainTrack.sourceUrl === source2.mainTrack.sourceUrl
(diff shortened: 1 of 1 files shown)


ahelmberger:
The first two commits only introduce a new util, the fix is in the third commit. Only at the end I realized that I kinda over-engineered this, so I'd like to have your input if I should revert the utils and only do some property comparisons in the multitrack media player module. For me, I don't have a preference.

adrianaluchian:
> The first two commits only introduce a new util, the fix is in the third commit. Only at the end I realized that I kinda over-engineered this, so I'd like to have your input if I should revert the utils and only do some property comparisons in the multitrack media player module. I don't have a preference.

I also think it's a bit overly complicated for this case 🤔
We could just set a  `const sourcePropsToRender = ['sourceUrl', 'aspectRatio', ...]` and just go through the list and compare prop to prop and decide if to rerender? Poor man's solution just in-place..

ahelmberger:
> > The first two commits only introduce a new util, the fix is in the third commit. Only at the end I realized that I kinda over-engineered this, so I'd like to have your input if I should revert the utils and only do some property comparisons in the multitrack media player module. I don't have a preference.
> 
> I also think it's a bit overly complicated for this case thinking We could just set a `const sourcePropsToRender = ['sourceUrl', 'aspectRatio', ...]` and just go through the list and compare prop to prop and decide if to rerender? Poor man's solution just in-place.. Though for `playbackRange` this would not work this way. Or maybe just really simply compare `source.aspectRatio`, `source.playbackRange[0]`, `source.playbackRange[1]` and so on?

I reverted the changes and only did some local comparisons, Now I think that's really enough, too :laughing:

adrianaluchian (approved):

PR #797, opened 2023-01-24
EDU-1061 Dashboard page improvements

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1061


src/components/user-card.yml
```diff
@@ -0,0 +1,3 @@
+profileButton:
+  en: view profile
```
ahelmberger, line 2: Should be uppercased: "**V**iew profile" (like "Enter room", "Accept invitation", etc.)

ahelmberger:
This is **so** good. The whole dashboard went from amateurish to starting-to-professional-looking, I cannot wait to see more improvements like these :heart_eyes::heart_eyes::heart_eyes:

src/components/user-card.yml
adrianaluchian: Oh, corrected, thanks!

adrianaluchian:
> This is **so** good. The whole dashboard went from amateurish to starting-to-professional-looking, I cannot wait to see more improvements like these 😍😍😍

Yes, it's amazing what an artistic eye can see!!

ahelmberger (approved):

PR #799, opened 2023-01-24
EDU-1036 Restore section display overlay button styles after Antd update

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1036

### DIFF
--- a/src/components/section-display.less
+++ b/src/components/section-display.less
@@ -155,27 +155,24 @@
 }
 
 .SectionDisplay-overlayButton {
-  .m-box-shadow;
-  border-radius: 50%;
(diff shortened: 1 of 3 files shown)


ahelmberger:
Unfortuntely, a lot of `!important`s were necessary to override the Antd button styles.

adrianaluchian:
> Unfortuntely, a lot of `!important`s were necessary to override the Antd button styles.

Yes, I tried to keep them to a minimum but sadly all were necessary.. And the default "small" size from Antd is still too big.

adrianaluchian (approved):

PR #803, opened 2023-01-25
EDU-1065 Document page improvements

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1065

**Example of badges look on "Documents" page:**
badges

**Example of plugin form layout, with info icon before label (as is currently):**
info-before

**and example of how this would look like if we'd consider moving it now after the label:**
info-after

**Scrolling through all the plugins in edit mode, before the changes:**
https://cdn.staging.openmusic.academy/media/Dg9HSzkDd3hSTCQEXyRTSK/document-plugin-forms-before-fzWj8btTk25nRzMj77RNXg.mov

**Scrolling through all the plugins in edit mode, after the changes:**
https://cdn.staging.openmusic.academy/media/Dg9HSzkDd3hSTCQEXyRTSK/document-plugin-forms-after-kJKHkcTk4euTW7RtzZU73U.mov

**Notes**
I find the plugins layout changes make info alot easier to read and find, because of the space created 🤩 
From fitting perspective I would suggest we try to find a different text for "Urheberschutzvermerk".
Would "Copyright Hinweis" or "Copyright-Vermerk" work?


ahelmberger:
> **Scrolling through all the plugins in edit mode, after the changes:** https://cdn.staging.openmusic.academy/media/Dg9HSzkDd3hSTCQEXyRTSK/document-plugin-forms-after-kJKHkcTk4euTW7RtzZU73U.mov

Looks like there's so much more space :heart_eyes: (although it's just redistributed). If we'd drop the colon after the labels, we could put the info icon on the right, so all text would start equally left-aligned, just a thought for a future direction (unless we redesign the form layout completely).

> **Notes** I find the plugins layout changes make info alot easier to read and find, because of the space created :star_struck:  From fitting perspective I would suggest we try to find a different text for "Urheberschutzvermerk". Would "Copyright Hinweis" or "Copyright-Vermerk" work?

We would have to ask, but I doubt it, as the label probably has to fit the legal terms exactly, and I don't know if that's the case at least in Germany.

adrianaluchian:
> If we'd drop the colon after the labels, we could put the info icon on the right, so all text would start equally left-aligned, just a thought for a future direction (unless we redesign the form layout completely).

I'm not sure about dropping the colon on the labels, since they are a bit far apart they might look disconnected.
But I had attached a screenshot of the variation with the info icon on the right, with keeping the colon, to see what you think of that.

> We would have to ask, but I doubt it, as the label probably has to fit the legal terms exactly, and I don't know if that's the case at least in Germany.
Would be a shame, as it's by far our longest label and the only problematic one. But we can leave it for now, no new issue was introduced with these changes regarding this label.

ahelmberger:
> > If we'd drop the colon after the labels, we could put the info icon on the right, so all text would start equally left-aligned, just a thought for a future direction (unless we redesign the form layout completely).
> 
> I'm not sure about dropping the colon on the labels, since they are a bit far apart they might look disconnected. But I had attached a screenshot of the variation with the info icon on the right, with keeping the colon, to see what you think of that.
> 

I agree, the colons help closing the gap between the label and the input. On the other hand, a colon directly after an icon looks a little bit strange to me. What about starting all label left-aligned and those labels that have an info icon have their icon left (like on a negative x coordinate)?

> > We would have to ask, but I doubt it, as the label probably has to fit the legal terms exactly, and I don't know if that's the case at least in Germany.
> 
> Would be a shame, as it's by far our longest label and the only problematic one. But we can leave it for now, no new issue was introduced with these changes regarding this label.

src/components/media-player/volume-presets-modal.js
```diff
@@ -89,12 +90,21 @@ function VolumePresetsModal({ volumePresets, isOpen, onOk, onClose }) {
...
+        <Button key="cancel" onClick={handleCancel}>
+          {t('common:cancel')}
+        </Button>,
+        <Button key="ok" type="primary" disabled={isModalOkButtonDisable} onClick={handleOk}>
+          {t('common:ok')}
+        </Button>
+      ]}
       >
       <div className="u-modal-body">
         {managedVolumePresets.map(renderVolumePreset)}
```
ahelmberger, line 110: Tiny inconsistency: could we disable the icon rendered by `renderVolumePreset` when there's only one preset left (and thus this last one cannot be deleted)?

src/components/document-metadata-modal.yml
```diff
@@ -22,9 +22,21 @@ invalidTags:
...
-  de: Standardvorlage benutzen
+tagsInfo:
+  en: Tags make the document discoverable when searching by words that match them
+  de: Tags make the document discoverable when searching by words that match them [but in German]
+content:
+  en: Content
+  de: Inhalt
+contentEmpty:
+  en: empty
+  de: Lehr
```
ahelmberger, line 33: It would be "Leer" - but what about:

Kein Inhalt - No content
Inhalt aus Vorlage - Content from template

?

src/components/document-metadata-modal.yml
```diff
@@ -22,9 +22,21 @@ invalidTags:
 tagsPlaceholder:
   en: Enter keywords
   de: Stichwörter angeben
-useTemplateDocument:
-  en: Use default template
-  de: Standardvorlage benutzen
+tagsInfo:
+  en: Tags make the document discoverable when searching by words that match them
+  de: Tags make the document discoverable when searching by words that match them [but in German]
```
ahelmberger, line 27: This description sounds very technical, maybe we could be more casual like:

Here you can specify search terms with which your document can be found by other users
Hier kannst du Suchbegriffe angeben, mit denen dein Dokument von anderen Nutzern gefunden werden kann

src/components/document-metadata-modal.yml
```diff
@@ -22,9 +22,21 @@ invalidTags:
...
+  de: Inhalt
+contentEmpty:
+  en: empty
+  de: Lehr
+contentFromTemplate:
+  en: from template
+  de: aus Vorlage
+contentInfo:
+  en: Either start with an empty document or with proposed content meant to help you with setting up your document
+  de: Either start with an empty document or with proposed content meant to help you with setting up your document [but in German]
```
ahelmberger, line 39: Starte entweder mit einem leeren Dokument oder mit einer Vorlage mit Inhaltsvorschlägen

ahelmberger:
I realized that, when I add a new comment to a topic, the topic panel still collapses for a millisecond and immediately reopens

src/components/pages/doc.yml
```diff
@@ -11,5 +11,20 @@ commentsPanelTitle:
   en: You are viewing the document page in "comments" mode
   de: Du befindest dich im Kommentar-Modus
 commentsHeader:
-  en: Comment topics
-  de: Kommentar-Themen
+  en: Comments
+  de: Kommentare
+viewHistoryRestrictionTooltip_annonymousUser:
+  en: Viewing document history is only allowed for logged in users
+  de: Ansehen von vorherige Versionen ?? is nur für angemeldet Benutzern elaubt
```
ahelmberger, line 18: Maybe we could incentive the user to login here:

Log in in order to view all versions of this document
Melde dich an, um die alle Versionen dieses Dokuments zu sehen

Then on a second thought, why do we not allow anonymous users to see all versions (they can still go to each version by permalink). The only thing they should not see when not logged in is the restore button :thinking:

src/components/pages/doc.yml
```diff
@@ -11,5 +11,20 @@ commentsPanelTitle:
...
+editRestrictionTooltip_annonymousUser:
+  en: Document editing is only allowed for logged in users
+  de: Bearbeitung von Dokumenten is nur für angemeldet Benutzern elaubt
```
ahelmberger, line 21: Maybe also here:

Log in in order to edit this document
Melde dich an, um dieses Dokument zu bearbeiten

src/components/pages/doc.yml
```diff
@@ -11,5 +11,20 @@ commentsPanelTitle:
...
+editRestrictionTooltip_openContribution:
+  en: Editing this document was disabled by a maintainer
+  de: Die Bearbeitung dieses Dokuments wurde von einem Maintainer deaktiviert
+editRestrictionTooltip_archive:
+  en: This document is archived
+  de: Dieses Dokument ist archiviert
```
ahelmberger, line 27: In order to understand this the user has to have the knowledge what the button would do when the document would not be archived. We could add more info:

This document cannot be edited because it is archived
Dieses Dokument kann nicht bearbeitet werden da es archiviert ist

src/components/media-player/volume-presets-modal.js
adrianaluchian: Done, good that you spotted this, it was a forgotten case in the `delete-button` component.

src/components/document-metadata-modal.yml
adrianaluchian: Since the label is already "Content" / "Inhalt", it would sound repetitive to put the word also in each radio option.
How does this look?
Screen Shot 2023-01-26 at 12 01 54
Screen Shot 2023-01-26 at 12 02 07

Maybe "kein", with small letters?

src/components/pages/docs.js
```diff
@@ -206,16 +203,20 @@ function Docs({ initialState, PageTemplate }) {
...
+        {row.allowedOpenContribution === DOCUMENT_ALLOWED_OPEN_CONTRIBUTION.content && (
+          <Tooltip title={t('allowedOpenContributionBadge_content')}>
+            <div className="u-badge">
+              C
+            </div>
+          </Tooltip>
+        )}
+        {row.allowedOpenContribution === DOCUMENT_ALLOWED_OPEN_CONTRIBUTION.metadataAndContent && (
+          <Tooltip title={t('allowedOpenContributionBadge_metadataAndContent')}>
+            <div className="u-badge">
```
ahelmberger, line 215: I was wondering if we should set `user-select: none;` on the `u-badge` class (or maybe set the cursor to question mark shape), because (at least in this use case) the combination of the caret shaped cursor and the tooltip looks not as nice.

Similarly the language "badge" should maybe also not have the caret (and maybe a tooltip showing the actual language name in full text in the current UI language).

What do you think?

src/components/document-metadata-modal.yml
ahelmberger: I would always capitalize the first word of each control. Unfortunately "Kein(er)" cannot stand alone in German, it just sounds weird, while in English "none" alone does work. I can only think of the two options:

* Repeating the word "Inhalt" as in "Kein Inhalt"
* Reverting back to "Leer" (but "Leerer Inhalt"/"Empty content" is as geeky as "empty string" or "empty set", for "normal" people, "Inhalt"/"Content" already excludes emptiness)

src/components/media-player/volume-presets-modal.js
ahelmberger: :+1:

src/components/document-metadata-modal.yml
adrianaluchian: Looking at other cases, we went for "Kein <Nomen>" so let's then go with:
- [en] "none" / "from template" (here we can afford to drop the repetition)
- [de] "Kein Inhalt" / "Inhalt aus Vorlage" (here if we repeat in one it sounds off to not repeat in the other)

src/components/pages/doc.yml
ahelmberger: > Maybe we could incentive the user to login here:
> 
> Log in in order to view all versions of this document Melde dich an, um die alle Versionen dieses Dokuments zu sehen
> 
> Then on a second thought, why do we not allow anonymous users to see all versions (they can still go to each version by permalink). The only thing they should not see when not logged in is the restore button thinking

Sorry, wrong German:
Melde dich an, um **~die~** alle Versionen dieses Dokuments zu sehen

src/components/document-metadata-modal.yml
adrianaluchian: Sounds better, changed!

src/components/document-metadata-modal.yml
adrianaluchian: Changed the German version and went for an English phrasing that sound similar.

src/components/pages/doc.yml
adrianaluchian: Much better, changed!

src/components/pages/doc.yml
adrianaluchian: That's true. And it's a tooltip, we can afford to be clearer (with more text) in tooltips. Changed.

src/components/pages/doc.yml
adrianaluchian: True, why not allow history navigation? Let's do that! I will adjust.

src/components/pages/docs.js
adrianaluchian: True, I hadn't notice the caret cursor. Super good idea also with the language 👏 
Changed all as suggested.

adrianaluchian:
> I realized that, when I add a new comment to a topic, the topic panel still collapses for a millisecond and immediately reopens

I think we can't avoid that, because the key of that panel changes. Initially it is "newTopic", then it changes to the actual topic. Which means the "newTopic" panel has to be closed, and the newly introduced panel needs to be open.

src/components/pages/doc.yml
adrianaluchian: Change, please have a look 🙏

ahelmberger:
> > I realized that, when I add a new comment to a topic, the topic panel still collapses for a millisecond and immediately reopens
> 
> I think we can't avoid that, because the key of that panel changes. Initially it is "newTopic", then it changes to the actual topic. Which means the "newTopic" panel has to be closed, and the newly introduced panel needs to be open.

Ah, ok, I see.

src/components/pages/doc.yml
ahelmberger: Gorgeous :heart_eyes:

ahelmberger (approved):

ahelmberger:
Great work :heart:

adrianaluchian:
> I agree, the colons help closing the gap between the label and the input. On the other hand, a colon directly after an icon looks a little bit strange to me. What about starting all label left-aligned and those labels that have an info icon have their icon left (like on a negative x coordinate)?

I managed to do this with a tiny tweak, but it looks really good now!

Screen Shot 2023-01-26 at 15 46 57

adrianaluchian:
> Great work ❤️

Many thanks! 🙇‍♀️  ☺️  And thank you for all the improvement suggestions and corrections!

PR #804, opened 2023-01-25
EDU-1063 Room page improvements

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1063

**Room members tab without pending invitations**
Screen Shot 2023-01-25 at 18 44 54

**Room members tab with pending invitations**
Screen Shot 2023-01-25 at 18 44 37

**Room membership confirmation page with single step**
Screen Shot 2023-01-25 at 20 18 40

**Room documents tab**
Screen Shot 2023-01-25 at 20 20 25

**Room settings tab**
Screen Shot 2023-01-25 at 20 29 43


src/components/pages/room-membership-confirmation.yml
```diff
@@ -1,12 +1,12 @@
+confirmationInProgress:
+  en: Confirmation in progress...
+  de: Confirmation in progress... [in German]
```
ahelmberger, line 3: Would be "Bestätigung in Bearbeitung..." (sounds like lawyer's term though)

As the users already know what process they are currently in, probably a simple "processing..." or "please wait (while we process your request)..." would be enough, I guess?

src/components/pages/room.js
```diff
@@ -433,25 +439,25 @@ export default function Room({ PageTemplate, initialState }) {
                 label: t('settingsTabTitle'),
                 children: (
                   <div className="Tabs-tabPane" >
-                    <Card className="RoomPage-card" title={t('updateRoomCardTitle')}>
+                    <div className="RoomPage-sectionHeadline">{t('roomMetadataHeadline')}</div>
+                    <section className="RoomPage-metadataSection">
```
ahelmberger, line 443: As this is the only section in the tab I would suggest that we do not need a title and additional indentation for the section content. Watcha think?

ahelmberger: Edit: I realized this is because it makes the effect of the update button clearer. So please ignore the comment.

adrianaluchian: I actually had removed the title cause I had the same thought, but then I realized we might need to adjust all labels, because it seemed a bit vague to have just a "Name" input, given we don't (yet 🤞 ) have a way to clearly showing the user the context they are in.. But let's see how this evolves.

src/components/pages/room-membership-confirmation.yml
adrianaluchian: I change it to "Please wait..." / "Bitte warte..." (though not sure if it sounds ok in German)

ahelmberger: Sounds good

ahelmberger (approved):

PR #805, opened 2023-01-26
EDU-1067 Improve resource selector

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1067


src/components/resource-picker/storage-location.js
```diff
@@ -123,29 +79,38 @@ function StorageLocation({
...
-            elementType={Search}
-            placeholder={t('common:search')}
-            value={typedInSearchTerm}
-            onSearch={handleSearchClick}
-            onChange={handleSearchTermChange}
+          <Input
+            allowClear
+            value={filterText}
+            placeholder={t('filterPlaceholder')}
+            onChange={handleFilterTextChange}
```
ahelmberger, line 98: One thing I did in my current ticket (which has a real-time table filter behaving in the exact same way) was adding `prefix={<SearchOutlined />}` which looks nice (I know it's technically no search, but people will use it to search for something). Maybe we could align on it (or something similar)?

src/components/resource-picker/storage-location.yml
```diff
@@ -1,18 +1,18 @@
...
 filesView_list:
-  en: List
-  de: Liste
+  en: List view
+  de: Listenansicht
 filesView_grid:
-  en: Grid
-  de: Grid
+  en: Grid view
+  de: Grid-Ansicht
```
ahelmberger, line 9: We could go with "Kachelansicht" in German here

ahelmberger:
Works very nicely :+1:

src/components/resource-picker/storage-location.js
adrianaluchian: Good idea, added it too 👍 
I tried using the AntD `FilterOutlined` icon but it does not look good enough to make it intuitive:
Screen Shot 2023-01-26 at 16 06 22

src/components/resource-picker/storage-location.yml
adrianaluchian: Went with, thanks!

src/components/resource-picker/storage-location.js
ahelmberger: That's exactly what I tried too but realized I perceive it as a speaker box that some rotated :stuck_out_tongue_closed_eyes:

ahelmberger (approved):

PR #806, opened 2023-01-26
EDU-1035 Add external account admin tooling

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1035


src/components/admin/license-settings.js
```diff
@@ -5,18 +5,20 @@ import { useTranslation } from 'react-i18next';
...
+
 function LicenseSettings({ license, onChange }) {
   const { t } = useTranslation();
 
   const name = license?.name || '';
   const url = license?.url || '';
 
   const isUrlValid = newUrl => {
-    return validateUrl(newUrl, t, { allowEmpty: true, allowHttp: true });
+    return validateUrl(newUrl, t, urlValidationOptions).validateStatus !== 'error';
```
adrianaluchian, line 17: This actually gets changed with this cleanup here https://github.com/educandu/educandu/pull/807/files 😋

src/server/batch-controller.js
```diff
@@ -51,10 +74,23 @@ class BatchController {
   registerApi(router) {
     router.get(
       '/api/v1/batches/:batchId',
-      needsPermission(permissions.VIEW_BATCHES),
+      needsPermission(permissions.MANAGE_BATCHES),
```
adrianaluchian, line 77: Very nice split of the admin controller into the actually responsible controllers with dedicated clearly named permission!
Makes things much more clear.

src/components/admin/settings-tab.js
```diff
@@ -78,120 +91,114 @@ function SettingsTab({
   const handleSaveButtonClick = async () => {
     const changedSettings = dirtyKeys.reduce((map, key) => ({ ...map, [key]: settings[key] }), {});
     try {
-      await settingsApiClient.saveSettings({ settings: changedSettings });
-      const updatedSettings = { ...lastSavedSettings, ...changedSettings };
-      setLastSavedSettings(updatedSettings);
+      setIsSaving(true);
+      const res = await settingsApiClient.saveSettings({ settings: changedSettings });
```
adrianaluchian, line 95: We need to pass `settingsApiClient.saveSettings({ settings: { ...settings, ...changedSettings } })` here, otherwise we lose/clear all other settings but the changed ones 😬

src/resources/resources.json
```diff
@@ -171,18 +171,68 @@
...
+      "accountsWithStorage": "Accounts with storage",
+      "closedAccounts": "Closed accounts",
+      "selectPlan": "Select plan",
+      "addReminder": "Add reminder",
+      "removeAllReminders": "Remove reminders",
+      "deleteAllOwnedRooms": "Delete rooms",
+      "storageSpace": "Storage space",
+      "reminders": "Reminders",
+      "accountClosedOn": "Account closing date",
+      "filterPlaceholder": "Filter by name, email, etc."
```
adrianaluchian, line 208: We only can filter by name or email actually. I think we might get a wrong expectation due to the ".etc" part.

adrianaluchian:
The code and the content are so much better organized this way! 😍 Wonderful job! 👏 
Also helps in the direction of better website navigation, if everything that is "admin-only" is on a clearly separated single place.

src/components/admin/license-settings.js
ahelmberger: Removed :-)

src/server/batch-controller.js
ahelmberger: Yeah, we did not think through permissions a lot, so we went very mixed, this consolidates it a little bit more.

src/components/admin/license-settings.js
adrianaluchian: 👌

src/components/admin/settings-tab.js
ahelmberger: Good catch, thanks :grimacing:

src/resources/resources.json
ahelmberger: That was the state before external accounts, now we can also filter by identity provider key and external user ID. I first wanted to change the placeholder depending on the current table the user sees, but then I realized this is very confusing and magical, so I decided to write it like that to indicate we're searching in all "relevant" fields. If the "etc." is too confusing we could also just simply say "Filter results" or something. What do you think?

adrianaluchian: Made a ticket for unifying filters, as discussed https://educandu.atlassian.net/browse/EDU-1080

src/components/admin/settings-tab.js
```diff
@@ -46,55 +44,52 @@ function SettingsTab({ onDirtyStateChange }) {
...
 
   const handleTemplateDocumentChange = useCallback(documentId => {
-    handleChange('templateDocument', { documentId }, true);
+    handleChange('templateDocument', { documentId });
   }, [handleChange]);
 
-  const handleHelpPageChange = useCallback((value, { isValid }) => {
-    handleChange('helpPage', value, isValid);
+  const handleHelpPageChange = useCallback(value => {
+    handleChange('helpPage', value);
```
adrianaluchian, line 68: Great that you simplified and cleaned up all this useless validation here!

ahelmberger: It's cleanup week, so I thought ...

src/components/admin/settings-tab.js
```diff
@@ -188,7 +182,7 @@ function SettingsTab({ onDirtyStateChange }) {
           loading={isSaving}
           onClick={handleSaveButtonClick}
           className="SettingsTab-saveButton"
-          disabled={!!invalidKeys.length || !dirtyKeys.length}
+          disabled={isLoading}
```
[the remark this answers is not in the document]
ahelmberger: Changed. And also added user feedback on save operations as discussed.

adrianaluchian (approved):

src/components/admin/settings-tab.js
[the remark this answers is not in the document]
adrianaluchian: Wonderful 👏 Let's merge this baby!

PR #811, opened 2023-01-27
EDU-1062 User profile page improvements

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1062


src/api-clients/user-api-client.js
```diff
@@ -200,15 +200,6 @@ class UserApiClient {
...
 
-  getCreatedDocuments({ userId }) {
-    return this.httpClient
-      .get(
-        `/api/v1/users/${encodeURIComponent(userId)}/created-documents`,
-        { responseType: 'json' }
-      )
-      .then(res => res.data);
-  }
-
```
ahelmberger, line 211: Nice that you put the replacement in to a better fitting place :+1:

ahelmberger:
Also: nice consolidation of the small font-size!

src/components/pages/user-profile.yml
```diff
@@ -0,0 +1,9 @@
+accountClosed:
+  en: This user account was closed.
+  de: Dieses Benutzerkonto wurde geschlossen.
+documentsHeadline:
+  en: Recent contributions
+  de: Letzte Beiträge
+moreButton:
+  en: Show {count} more
+  de: Zeig {count} weitere an / {count} weitere anzeige[n] ?
```
ahelmberger, line 9: {count} weitere anzeigen

ahelmberger: But, if `count` is 1 then "{count} weitere**n** anzeigen" (meaning: "1 weiteren Beitrag")

ahelmberger:
As the page name has changed we should also flag the ticket and add the necessary config/code updates

src/components/pages/user-profile.yml
adrianaluchian: Thanks, changed it so.

ahelmberger (approved):

adrianaluchian:
> As the page name has changed we should also mark the ticket and add the necessary config/code updates

Not the case (yet).

PR #813, opened 2023-01-27
EDU-1080 Extract filter component

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1080

**Before**
Screen Shot 2023-01-27 at 16 21 00

**Current**
Screen Shot 2023-01-27 at 16 28 00

### DIFF
--- a/src/components/pages/docs.yml
+++ b/src/components/pages/docs.yml
@@ -7,9 +7,9 @@ newDocument:
 initialAuthor:
   en: Initial author
   de: Erstautor
-filterPlaceholder:
-  en: Filter by title or initial author
-  de: Nach Titel oder Erstautor filtern
+placeholder:
+  en: Search by title or initial author
+  de: Nach Titel oder Erstautor suchen
 badges:
   en: Badges
   de: Kennzeichen
--- a/src/components/pages/docs.yml
+++ b/src/components/pages/docs.yml
@@ -7,9 +7,9 @@ newDocument:
 initialAuthor:
   en: Initial author
   de: Erstautor
-placeholder:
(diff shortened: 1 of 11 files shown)


src/components/pages/docs.yml
```diff
@@ -7,9 +7,9 @@ newDocument:
 initialAuthor:
   en: Initial author
   de: Erstautor
-filterPlaceholder:
-  en: Filter by title or initial author
-  de: Nach Titel oder Erstautor filtern
+placeholder:
+  en: Search by title or initial author
+  de: Nach Titel oder Erstautor suchen
```
adrianaluchian, line 12: Even though we use the input as a filter, we display to the user a search input, so I thought better to phrase this according to the input.

ahelmberger: Is there a reason why you did not use the new component here as `<FilterInput size="large" ... />`?

ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-1080
> 
> **Before** Screen Shot 2023-01-27 at 16 21 00
> 
> **Current** Screen Shot 2023-01-27 at 16 28 00

Looks actually a lot better than I anticipated :relaxed:

src/components/pages/docs.yml
adrianaluchian: The only reason is an invalid one actually 😆 I confused the "Documents" page with the "Search" page (not for the first time) and considered that here it should be search.
I shall change it!

adrianaluchian: Changed, thanks!

ahelmberger (approved):

PR #814, opened 2023-01-27
EDU-1079 Scroll to the last seen section after switching views on doc page

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1079

### DIFF
--- a/src/components/pages/doc.js
+++ b/src/components/pages/doc.js
@@ -1,4 +1,3 @@
-import by from 'thenby';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { ALERT_TYPE } from '../alert.js';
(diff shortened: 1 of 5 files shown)


src/components/pages/doc.js
```diff
@@ -125,6 +127,62 @@ function Doc({ initialState, PageTemplate }) {
...
+  };
+
+  const switchView = newView => {
+    setLastViewInfo({ view, sectionKeyToScrollTo: findSectionKeyToScrollTo() });
+    setView(newView);
+  };
+
+  useEffect(() => {
+    if (view !== VIEW.comments && lastViewInfo?.view !== VIEW.comments && lastViewInfo?.sectionKeyToScrollTo) {
+      setTimeout(() => tryBringSectionIntoView(lastViewInfo.sectionKeyToScrollTo), 500);
```
adrianaluchian, line 178: 🤯 Magical but convenient!

src/components/pages/doc.js
```diff
@@ -125,6 +127,62 @@ function Doc({ initialState, PageTemplate }) {
...
+    const element = window.document.querySelector(`[data-section-key="${sectionKey}"]`);
+    if (!element) {
+      return;
+    }
+
+    const measurement = getViewportMeasurementsForElement(element);
+    if (!measurement.elementTopIsInViewport) {
+      element.scrollIntoView({ behavior: 'smooth' });
+    }
+  };
```
adrianaluchian, line 169: Suggestion from me would be to move `findSectionKeyToScrollTo` and `tryBringSectionIntoView` into `doc-utils.js` to reduce the logic on this huge page a big.

adrianaluchian (approved):

src/components/pages/doc.js
ahelmberger: Good idea. Done! It also helped to keep the custom data attribute names in one place :+1:

PR #815, opened 2023-01-27
EDU-1060 Search page improvements

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1060


src/components/table.js
```diff
@@ -1,43 +0,0 @@
...
-      default:
-        return originalElement;
-    }
-  };
-
-  return (
-    <AntdTable
-      className={['Table', className].filter(x => x).join(' ')}
-      rowClassName={() => 'Table-row'}
-      Pagination={!!pagination && { position: ['bottomCenter'], showSizeChanger: false, itemRender }}
```
adrianaluchian, line 26: This no longer had any effect since the AntD version update. We cannot affect the table pagination from the outside anymore, except for a single option, `position`.

To streamline table usages, I've turned this into a simple `DocumentsTable` component (having a custom component also called `Table` just like the AntD one made it confusing). I've also dropped the customizations of the paging, because with it we were now straying too much from AntD styling and we should probably keep it simple and use the styles that it has. This also keeps the website look and feel more consistent "ant" style, I would say.

src/components/table.yml
```diff
@@ -1,6 +0,0 @@
-prev:
-  en: ‹ back
-  de: ‹ zurück
-next:
-  en: › forward
-  de: › weiter
```
adrianaluchian, line 6: These texts were also no longer the applied ones, since the AntD update, as the whole `itemRender` was not supported anymore via the table.

src/components/table.js
ahelmberger: Ha! I actually completely forgot that we have our own `Table` (only stumbled upon it yesterday). Yeah, let's use the provided pagination and not meddle too much in it. Given the screenshots it is not really worth the difference.

src/components/table.yml
ahelmberger: :sweat_smile:

src/components/tag-selector.yml
```diff
@@ -1,7 +1,4 @@
-linkTextSelectMore:
-  en: + More
-  de: + Mehr
-linkTextSelectFirst:
+linkText:
   en: + Add tag filter
   de: + Tag-Filter hinzufügen
```
ahelmberger, line 3: As this label is directly next to the "x"-Button/Link, we should also adjust the translation `deselectTags` in `search.yml` to "Remove all tag filters" instead of "Remove all tags". What do you think? Otherwise we call these things by two different names.

adrianaluchian: True, it's inconsistent this way.
Dropped the "filters" from ""Remove all tags" as discussed.

ahelmberger (approved):

PR #821, opened 2023-01-30
EDU-1056 Improve copywriting

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1056


src/components/login-form.yml
```diff
@@ -3,11 +3,11 @@ loginForm:
     en: Please enter your password
     de: Bitte gib dein Kennwort an
   loginFailed:
-    en: Login failed. Please check your input.
-    de: Die Anmeldung ist fehlgeschlagen. Bitte überprüfe deine Eingabe.
+    en: The credentials you have entered are not correct.
+    de: Die Zugangsdaten sind nicht richtig.
```
adrianaluchian, line 7: There is probably a better way to write this in German..

src/components/sorting-selector.yml
```diff
@@ -2,6 +2,9 @@ sortingSelector:
   changeSortingValue:
     en: Change sorting
     de: Sortierung ändern
-  changeSortingDirection:
-    en: Change sorting direction
-    de: Sortierungsrichtung ändern
+  changeSortingDirectionAscending:
+    en: Change to ascending order
+    de: In aufsteigende Reihenfolge sortieren
```
adrianaluchian, line 7: For the German version, I think a similar "Change to" phrasing might make it clearer for the user that it is not the current sorting direction, but the one that will be applied. Please help 😊

src/components/blocked-login-error.yml
```diff
@@ -1,9 +1,9 @@
 blockedLoginError:
   loginFailedTooOften:
-    en: Too many failed login attempts within a short time period. Please try again later.
+    en: Too many failed log in attempts within a short time period. Please try again later.
```
ahelmberger, line 3: Actually (I also checked this on LEO), "login" was correct here, as it is a noun here (noun: "the login", verb: "to log in").

src/components/login-form.yml
ahelmberger: Maybe we can also use "korrekt" in German: "Die Zugangsdaten sind nicht korrekt.", or to match the English translation completely: "Die Zugangsdaten, die du eingegeben hast, sind nicht korrekt." (might be too long, and it's kinda clear that they were entered by the user, though).

src/components/sorting-selector.yml
ahelmberger: I agree. The best I can think of would be: "Zu (auf|ab)steigender Sortierung wechseln"

src/resources/common.yml
```diff
@@ -188,8 +179,16 @@ common:
...
   slug:
-    en: URL path
-    de: URL-Pfad
+    en: Slug
+    de: Slug
+  slugInfo:
+    en: |
+      The slug is added to the end of the URL and can contain one or more parts separated by slashed. Each part and can only contain lower case letters (a-z), digits and hyphens.
+    de: |
+      Der Slug wird am ende des URLs hinzugefügt und darf einen order mehrere durch Schrägstriche getrennte Teile enthalten. Jeder Teil darf nur Kleinbuchstaben (a-z), Ziffern und Bindestriche enthalten.
```
ahelmberger, line 188: Some typos: "**E**nde" / "o**~~r~~**der"

src/components/sorting-selector.yml
adrianaluchian: Changed, thanks you!

src/components/login-form.yml
adrianaluchian: I agree, "Die Zugangsdaten sind nicht korrekt." it is then.

src/components/blocked-login-error.yml
adrianaluchian: Indeed, it would have needed to be "Too many failed logging in attempts" 🤔 Corrected!

src/resources/common.yml
adrianaluchian: Amazed those were the only problems! Corrected, thanks 😊

ahelmberger (approved):

PR #824, opened 2023-01-31
EDU-1078 Rework comments panel

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1078


src/components/comments-panel.yml
```diff
@@ -1,16 +1,22 @@
...
+    de: Neues Thema hinzufügen
   topicHeaderPrefix:
     en: Topic
     de: Thema
-  newTopicPlaceholder:
+  newTopicHeaderPrefix:
     en: New topic
     de: Neues Thema
+  newTopicPlaceholder:
+    en: e.g. orthography, citation, etc.
```
adrianaluchian, line 18: We could write these with capital first letter too, as they are examples of topic names 🤔

ahelmberger: True. It looked a little bit weird (in both languages actually), but let's stay consistent. Changed it.

adrianaluchian:
Looks and behaves alot better! 😍 
It's a tiny bit magical when you have the "new topic" panel expanded, and you try to collapse it, then it vanishes and the button appears again. But having this button makes rendering the sections easier and cleaner after posting to the API.

adrianaluchian (approved):

PR #826, opened 2023-01-31
EDU-1070 Make documents page admin only

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1070


src/components/pages/doc.js
```diff
@@ -73,6 +74,24 @@ function createPageAlerts({ doc, docRevision, view, hasPendingTemplateSectionKey
   return alerts;
 }
 
+function getDocumentMetadataModalState({ t, doc, room, isCloning, isOpen = false }) {
```
ahelmberger, line 77: Nice idea!

src/components/pages/redaction.js
```diff
@@ -1,32 +1,38 @@
...
 import ActionButton, { ActionButtonGroup, ACTION_BUTTON_INTENT } from '../action-button.js';
 import { DOCUMENT_ALLOWED_OPEN_CONTRIBUTION, DOC_VIEW_QUERY_PARAM } from '../../domain/constants.js';
 
 const logger = new Logger(import.meta.url);
 
+const TABS = {
+  documents: 'documents'
+};
+
+const determineTab = query => Object.values(TABS).find(val => val === query) || TABS.settings;
```
ahelmberger, line 34: `TABS.settings` does not exist :laughing:

adrianaluchian: Corrected (and improved in copy origin too 😋 )

src/server/controller-factory.js
```diff
@@ -16,27 +16,29 @@ import DocumentController from './document-controller.js';
...
+  SettingsController,
+  DocumentController,
   RevisionController,
   PdfJsApiController,
-  AdminController,
-  AmbController,
-  CommentController
+  DashboardController,
+  RedactionController,
+  UserAgentController
```
ahelmberger, line 41: Even though I strongly appreciate (and admire) the sentiment behind this reordering, the order, in which the middlewares are registered (and then finally called) matters a lot. E.g. the `I18nController` sets the correct language on the `req` object which other controllers further down the pipeline might then use, the `SettingsController` sets `req.settings` which other middleware later might use. The static controller was first so that static file requests e.g. for a font or a script file do not go through the whole application pipeline. The `ErrorController` should be last (I know it somehow has moved into the middle during the last years :thinking:), as it contains the final final middleware that ever should be called. So even though the app might seem to work just fine with the different order, it might cause some subtle bugs that we don't instantly see.

src/components/pages/redaction.js
ahelmberger: :rofl:

src/server/controller-factory.js
adrianaluchian: 😂 Oh I got carried away here!
Thanks for clarifying all the positions, I will try to also change how they are registered so that it's intuitive in the future.

adrianaluchian: Changed!

ahelmberger: That is some nice organization :heart: thanks!

ahelmberger (approved):

PR #827, opened 2023-02-01
EDU-1017 Add media library items collection

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1017


src/utils/resource-utils.js
```diff
@@ -14,31 +15,26 @@ import FileUnknownFilledIcon from '../components/icons/files/file-unknown-filled
 
 const extensionsGroups = [
   { type: RESOURCE_TYPE.pdf, extensions: ['pdf'] },
-  { type: RESOURCE_TYPE.text, extensions: ['txt', 'doc', 'rtf', 'odt'] },
+  { type: RESOURCE_TYPE.text, extensions: ['txt'] },
```
ahelmberger, line 18: There's actually no reason to have a separate `document` category for file types that no browser can do anything with (and why wouldn't we then not also have a category for Powerpoint presentations or Excel sheets?). So everything that's not simply plain text should be `unknown` (can be displayed as "other" on the UI). I will update the ticket accordingly.

src/services/storage-service.js
```diff
@@ -298,7 +299,8 @@ export default class StorageService {
 
   async _uploadFiles(files, parentPath) {
     const cdnPathByOriginalName = files.reduce((map, file) => {
-      map[file.originalname] = componseUniqueFileName(file.originalname, parentPath);
+      const uniqueFileName = createUniqueStorageFileName(file.originalname);
+      map[file.originalname] = parentPath ? urlUtils.concatParts(parentPath, uniqueFileName) : uniqueFileName;
```
adrianaluchian, line 303: You could remove the ternary operator here, `urlUtils.concatParts` is smart enough to take care of empty parts.

src/stores/media-library-item-store.js
```diff
@@ -0,0 +1,35 @@
...
+
+  constructor(db) {
+    this.collection = db.mediaLibraryItems;
+  }
+
+  getMediaLibraryItemsByConditions(conditions, { session } = {}) {
+    return this.collection.find({ $and: conditions }, { session }).toArray();
+  }
+
+  getMediaLibraryItemTagsMatchingText(text) {
```
adrianaluchian, line 17: getMediaLibraryItem**s**TagsMatchingText

src/utils/resource-utils.js
adrianaluchian: Indeed, sounds good! 👍

src/utils/storage-utils.js
```diff
@@ -82,6 +81,10 @@ export function getStorageLocationTypeForPath(path) {
   return STORAGE_LOCATION_TYPE.unknown;
 }
 
+export function getMediaLibraryItemPath() {
```
adrianaluchian, line 84: This would be `getMediaLibraryPath` actually 🤔  Or 
```
getMediaLibraryItemPath(itemId) {
     return `media-library/${itemId}`;
}
```

src/utils/tag-utils.spec.js
```diff
@@ -0,0 +1,135 @@
+import { beforeEach, describe, expect, it } from 'vitest';
+import { createTagSearchQuery, createTagsPipelineQuery } from './tag-utils.js';
+
+describe('tag-utils', () => {
```
adrianaluchian, line 4: Wonderful that you test this!

src/utils/url-utils.js
```diff
@@ -38,6 +38,14 @@ function concatParts(...parts) {
     : '';
 }
 
+function splitAtExtension(pathOrUrl) {
```
adrianaluchian, line 41: Also, nice extraction!

src/services/storage-service.js
ahelmberger: Indeed, good point, changed!

src/stores/media-library-item-store.js
ahelmberger: It's debatable. I named this after `getDocumentTagsMatchingText` that does the same for document tags (actually not: documents tags). The function does not return a list of MediaLibraryItems but a list of MediaLibraryItemTags (tags that are set on a MediaLibraryItems). I'd personally prefer to keep it like that, but I'm also not very passionate about it.

src/utils/storage-utils.js
ahelmberger: Absolutely right, should be `getMediaLibraryPath` (there is actually no `itemId` in the path, just the filename, I think). Changed it.

src/utils/url-utils.js
ahelmberger: :bow:

src/utils/tag-utils.spec.js
ahelmberger: :+1:

PR #828, opened 2023-02-02
EDU-1014 Change CDN path structure

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1014

### DIFF
--- a/migrations/educandu-2023-02-01-02-change-cdn-urls.js
+++ b/migrations/educandu-2023-02-01-02-change-cdn-urls.js
@@ -1,0 +1,197 @@
+import GithubFlavoredMarkdown from '../src/common/github-flavored-markdown.js';
+
+const OLD_ROOMS_CDN_PREFIX = 'cdn://rooms/';
+const OLD_MEDIA_CDN_PREFIX = 'cdn://media/';
+
+export default class Educandu_2023_02_01_02_change_cdn_urls {
+  constructor(db) {
+    this.db = db;
+    this.gfm = new GithubFlavoredMarkdown();
+  }
+
+  redactUrl(url = '') {
+    if (url.startsWith(OLD_ROOMS_CDN_PREFIX)) {
+      return url
+        .replace(OLD_ROOMS_CDN_PREFIX, 'cdn://room-media/')
+        .replace('/media/', '/');
+    }
+    if (url.startsWith(OLD_MEDIA_CDN_PREFIX)) {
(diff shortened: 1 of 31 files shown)


migrations/educandu-2023-02-01-02-change-cdn-urls.js
```diff
@@ -0,0 +1,197 @@
...
+    }
+    return url;
+  }
+
+  redactMarkdown(text = '') {
+    return this.gfm.redactCdnResources(text, url => this.redactUrl(url));
+  }
+
+  updateDoc(doc) {
+    doc.cdnResources = doc.cdnResources.forEach(this.redactUrl);
```
ahelmberger, line 29: I think this should be `map` instead of `forEach` :thinking:

migrations/educandu-2023-02-01-02-change-cdn-urls.js
```diff
@@ -0,0 +1,197 @@
...
+
+    for (const section of doc.sections) {
+      if (!section.content) {
+        return;
```
ahelmberger, line 33: `return` would leave the whole update function, this should be `continue`. Or you could use mapping:

`doc.sections = doc.sections.map(section => { ... });

src/utils/storage-utils.js
```diff
@@ -13,6 +13,18 @@ import {
 
 const rasterImageFileTypes = ['image/png', 'image/jpg', 'image/jpeg', 'image/webp'];
 
+export function getMediaLibraryPath() {
+  return 'media-library/';
```
ahelmberger, line 17: I know we use `urlUtils.concatParts` in the place where this is called to append the file name, but as the other two helpers `getDocumentMediaDocumentPath` and `getRoomMediaRoomPath` do not append a slash at the end, I wonder why you added a trailing slash in this one?

migrations/educandu-2023-02-01-02-change-cdn-urls.js
adrianaluchian: 😳 Thank you! Corrected.

migrations/educandu-2023-02-01-02-change-cdn-urls.js
adrianaluchian: Of course, corrected this and added a basic test covering all cases with having a plugin type as an example. Thanks! 🙏

src/utils/storage-utils.js
adrianaluchian: No need to keep wondering, removed! 😅

ahelmberger (approved):

PR #842, opened 2023-02-08
EDU-1107 Fix broken control panel state

adrianaluchian:
Gladly fixes https://educandu.atlassian.net/browse/EDU-1107

### DIFF
--- a/src/components/pages/doc.js
+++ b/src/components/pages/doc.js
@@ -123,6 +123,20 @@ function Doc({ initialState, PageTemplate }) {
     window.removeEventListener('resize', ensureControlPanelPosition);
   });
 
+  const determineInitialViewState = () => {
+    const requestView = Object.values(VIEW).find(v => v === request.query.view);
+    if (!requestView) {
+      return { preSetView: null, view: VIEW.display };
+    }
+
+    if (requestView === VIEW.edit) {
+      const allowedView = user ? VIEW.edit : VIEW.display;
+      return { preSetView: allowedView, view: allowedView };
+    }
+
+    return { preSetView: requestView, view: requestView };
+  };
+
   const userCanHardDelete = hasUserPermission(user, permissions.HARD_DELETE_SECTION);
   const userCanEdit = hasUserPermission(user, permissions.EDIT_DOC);
   const userCanEditDocContent = canEditDocContent({ user, doc: initialState.doc, room });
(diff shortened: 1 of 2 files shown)


src/components/pages/doc.js
```diff
@@ -133,13 +131,14 @@ function Doc({ initialState, PageTemplate }) {
     ? getEditDocContentRestrictionTooltip({ t, user, doc: initialState.doc, room })
     : t('editRestrictionTooltip_annonymousUser');
 
-  const [isDirty, setIsDirty] = useState(false);
+  const [view, setView] = useState(null);
```
ahelmberger, line 134: This makes the first render (and on the server side, the only render) have `view === null`, which fixes the control panel by not rendering it at all on the server, as we have this logic further down:

~~~
  const showHistoryPanel = view === VIEW.display || view === VIEW.history;
  const showCommentsPanel = view === VIEW.display || view === VIEW.comments;
  const showEditPanel = view === VIEW.display || view === VIEW.edit;
~~~

They are all `false` on the first render, of course.

I find it a little bit problematic though to have `view` initially basically in an undefined state. Could we maybe change it so that it picks up the initial view from the get go (not in a `useEffect`) and at the same time have a different logic (e.g. by means of `useEffect` to defer the control panel rendering until hydration has completed on the client?

adrianaluchian: Although setting an initial state and then re-setting on re-render is a bit of double-bookkeeping with risk of inconsistency to me, I do see your point here. I will change it to have the default value, "by default".

ahelmberger: Why would we reset differently on re-render, we already know the correct value on the very first render: `Object.values(VIEW).find(v => v === request.query.view) || VIEW.display`? :thinking:

adrianaluchian: Only as a left-over, because I had used `window` to get the query param and then switched back 😊 So now there's no need to re-set in `useEffect`. I have changed the logic to only set by default.

src/components/pages/doc.js
```diff
@@ -123,7 +123,19 @@ function Doc({ initialState, PageTemplate }) {
...
   });
 
-  const initialView = Object.values(VIEW).find(v => v === request.query.view) || VIEW.display;
+  const determineInitialViewState = () => {
+    const requestView = Object.values(VIEW).find(v => v === request.query.view);
+    if (!requestView) {
+      return { preSetView: null, view: VIEW.display };
+    }
+
+    if (requestView === VIEW.edit) {
```
ahelmberger, line 132: I think this `if` block is not necessary as the controller already prevents unprivileged users from requesting the page in edit mode:

~~~
    if (view === DOC_VIEW_QUERY_PARAM.edit && !canEditDocContent({ user, doc, room })) {
      return res.redirect(routes.getDocUrl({ id: doc._id, slug: doc.slug }));
    }
~~~
src/server/document-controller.js, lines 105-107

src/components/pages/doc.js
ahelmberger: Cool, great :+1:

src/components/pages/doc.js
adrianaluchian: Nicely spotted, simplified, thanks!

PR #846, opened 2023-02-09
EDU-1095 Extend catalog plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1095

### DIFF
--- a/src/components/drag-and-drop-container.js
+++ b/src/components/drag-and-drop-container.js
@@ -16,8 +16,10 @@ function DragAndDropContainer({
 
   const handleDragEnd = ({ source, destination }) => {
     setIsDragging(false);
-    if (source !== destination && destination >= 0 && destination <= (items.length - 1)) {
-      onItemMove(source.index, destination.index);
+    const fromIndex = source.index;
+    const toIndex = destination.index;
+    if (fromIndex !== toIndex && toIndex >= 0 && toIndex <= (items.length - 1)) {
+      onItemMove(fromIndex, toIndex);
     }
   };
 
--- a/src/components/drag-and-drop-container.js
+++ b/src/components/drag-and-drop-container.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { memo, useState } from 'react';
+import React, { useState } from 'react';
 import { DragDropContext, Droppable, Draggable } from 'react-beautiful-dnd';
 
 function DragAndDropContainer({
@@ -72,4 +72,4 @@ DragAndDropContainer.defaultProps = {
   onItemMove: () => {}
 };
 
-export default memo(DragAndDropContainer);
+export default DragAndDropContainer;
--- a/src/components/drag-and-drop-container.js
+++ b/src/components/drag-and-drop-container.js
@@ -40,13 +40,11 @@ function DragAndDropContainer({
                       ...draggableProvided.draggableProps.style
                     }}
                     >
-                    <section key={item.key}>
-                      {item.render({
-                        dragHandleProps: draggableProvided.dragHandleProps,
-                        isDragged: draggableState.isDragging,
-                        isOtherDragged: !!isDragging && !draggableState.isDragging
-                      })}
-                    </section>
+                    {item.render({
+                      dragHandleProps: draggableProvided.dragHandleProps,
+                      isDragged: draggableState.isDragging,
+                      isOtherDragged: !!isDragging && !draggableState.isDragging
+                    })}
                   </div>
                 )}
               </Draggable>
--- a/src/components/drag-and-drop-container.js
+++ b/src/components/drag-and-drop-container.js
@@ -16,9 +16,9 @@ function DragAndDropContainer({
 
   const handleDragEnd = ({ source, destination }) => {
     setIsDragging(false);
-    const fromIndex = source.index;
-    const toIndex = destination.index;
-    if (fromIndex !== toIndex && toIndex >= 0 && toIndex <= (items.length - 1)) {
+    const fromIndex = source?.index;
+    const toIndex = destination?.index;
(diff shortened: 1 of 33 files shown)


ahelmberger:
The new panel headers look and feel so much nicer to work with. The cherry on top would be, if the panel "title" could reflect the content (especially helpful when reordering them by dragging or when some of them are collapsed while working on others), for examples in the catalog plugin:

Instead of:

localhost_3000_docs_z51Fmh1Mp8SiZ43cFUZe7X_view=edit

... it could look like:

localhost_3000_docs_z51Fmh1Mp8SiZ43cFUZe7X_view=edit_new
*(of course not in red :laughing:)*

Watcha think?

src/plugins/catalog/catalog.less
```diff
@@ -10,6 +10,18 @@
...
 }
 
+.CatalogDisplay-linkListItem {
+  display: flex;
+  flex-direction: column;
+  margin: 5px 0;
+}
+
+.CatalogDisplay-itemLinkDescription {
+  text-align: left;
```
ahelmberger, line 20: In case of image tiles, `center` would look nicer here (as the title directly above is also centered)

src/components/drag-and-drop-container.js
```diff
@@ -0,0 +1,75 @@
...
+                  <div
+                    key={item.key}
+                    ref={draggableProvided.innerRef}
+                    {...draggableProvided.draggableProps}
+                    style={{
+                      userSelect: draggableState.isDragging ? 'none' : null,
+                      ...draggableProvided.draggableProps.style
+                    }}
+                    >
+                    <section key={item.key}>
```
ahelmberger, line 43: Just out of curiosity, is there a reason to nest another `section` inside the `div`, it seems to not have any purpose?

src/components/sections-display.js
```diff
@@ -27,26 +27,11 @@ function SectionsDisplay({
   onSectionPasteFromClipboard,
   onSectionHardDelete
 }) {
-  const [isDragging, setIsDragging] = useState(false);
+  const droppableIdRef = useRef(useId());
```
ahelmberger, line 30: :partying_face: first usage, yay!

src/domain/validation-constants.js
```diff
@@ -13,6 +13,8 @@ export const minUserPasswordLength = 8;
 
 export const maxMediaLibraryItemDescriptionLength = 1000;
 
+export const maxCatalogPluginItemLinkDescriptionLength = 200;
```
ahelmberger, line 16: As discussed we could lift that restriction.

src/resources/common.yml
```diff
@@ -503,3 +503,6 @@ common:
   posterImageUrl:
     en: Preview image URL
     de: Vorschau-Bild-URL
+  dragToReorder:
+    en: Drag to reorder
+    de: Durch Bewegen neu anordnen
```
ahelmberger, line 508: Maybe "Mit Drag&Drop verschieben" would sound clearer in the average user's ear.

src/components/drag-and-drop-container.js
adrianaluchian: I've done that so that the `key` is set in a clear way on the iterated elements. Otherwise that needed to go inside the rendered component and it's easier to miss to set it. Found that it gets a problem out of the way and contained in just in one place.

src/components/sections-display.js
adrianaluchian: Yeees! 😁 I'm happy I could remember about it 😅 Pretty handy!

src/plugins/catalog/catalog.less
adrianaluchian: Sure thing, wasn't sure which was better. Now it is centered.

src/domain/validation-constants.js
adrianaluchian: 🏋️‍♀️ lifted

src/resources/common.yml
adrianaluchian: Changed, thanks!

On that note, I am wondering if now, for the sake of consistency and having the user get used to a pattern, should we maybe re-introduce the drag icon also for sections (with tooltip as well and in the same position as for these inner panels) ?
It might even be now that some users don't know that they can drag and drop sections. That would not surprise me.

adrianaluchian:
> The new panel headers look and feel so much nicer to work with. The cherry on top would be, if the panel "title" could reflect the content (especially helpful when reordering them by dragging or when some of them are collapsed while working on others), for examples in the catalog plugin [...]
> Watcha think?

Good idea! 👏 I've done it so.

src/plugins/catalog/catalog.less
ahelmberger: :+1:

src/resources/common.yml
ahelmberger: I'm torn between the two ways: on the one hand I like that I can just drag anywhere on the header, on the other hand I'm as concerned as you that users might not spot the possibility in the first place in lieu of the missing visual indication the icon would give. I wouldn't change it right away but discuss about it. Maybe we could even ask Radu for advice.

src/components/drag-and-drop-container.js
ahelmberger: Just nitpicking here, but I think we could reduce the nesting here as the key is set already on the parent `div`.

ahelmberger (approved):

ahelmberger:
> > The new panel headers look and feel so much nicer to work with. The cherry on top would be, if the panel "title" could reflect the content (especially helpful when reordering them by dragging or when some of them are collapsed while working on others), for examples in the catalog plugin [...]
> > Watcha think?
> 
> Good idea! clap I've done it so.

Nice :+1:

PR #847, opened 2023-02-09
EDU-1114 Make room documents draggable

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1114

**Owner view:**
Screen Shot 2023-02-09 at 17 24 14

**Collaborator view:**
Screen Shot 2023-02-09 at 17 26 52

**Member view (aka exclusive room member):**
Screen Shot 2023-02-09 at 17 27 16

### DIFF
--- a/src/components/pages/room.js
+++ b/src/components/pages/room.js
@@ -229,9 +229,14 @@ export default function Room({ PageTemplate, initialState }) {
 
   const handleDocumentMove = async (fromIndex, toIndex) => {
     const reorderedDocumentIds = moveItem(room.documents, fromIndex, toIndex);
+    const clientSideDocumentsReordered = reorderedDocumentIds.map(_id => documents.find(doc => doc._id === _id));
+    setDocuments(clientSideDocumentsReordered);
+
     const response = await roomApiClient.updateRoomDocumentsOrder({ roomId: room._id, documentIds: reorderedDocumentIds });
+
+    const serverSideDocumentsReordered = getSortedDocuments(response.room, documents);
     setRoom(response.room);
-    setDocuments(getSortedDocuments(response.room, documents));
+    setDocuments(serverSideDocumentsReordered);
     message.success(t('common:changesSavedSuccessfully'));
   };
 
--- a/src/components/pages/room.js
+++ b/src/components/pages/room.js
@@ -500,6 +500,7 @@ export default function Room({ PageTemplate, initialState }) {
                 label: t('documentsTabTitle', { count: documents.length }),
                 children: (
                   <div className="Tabs-tabPane">
+                    {renderRoomDescription()}
                     {renderCreateDocumentButton()}
                     {!documents.length && t('documentsPlaceholder')}
                     {renderNonDraftDocumentsAsDraggable()}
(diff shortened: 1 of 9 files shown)


ahelmberger:
Looks nice, but I think I found an issue:

Dragging works perfectly fine for documents that were initially loaded with the page. But when I create a new document sequence in the room and then try to drag one of those new documents, they seem to fall back to their old place.

Another (but small) glitch is, that after dropping, the order reverts to the original order for a fraction of a second before turning into the new order (probably due to server communication because the glitch length is proportional to the network throttling :grimacing:)

src/components/pages/room.js
```diff
@@ -9,33 +9,39 @@ import { useUser } from '../user-context.js';
...
+import React, { Fragment, useEffect, useId, useMemo, useRef, useState } from 'react';
 import { roomShape, invitationShape, documentExtendedMetadataShape } from '../../ui/default-prop-types.js';
 import { confirmDocumentDelete, confirmRoomDelete, confirmRoomMemberDelete, confirmRoomInvitationDelete, confirmLeaveRoom } from '../confirmation-dialogs.js';
 
 const logger = new Logger(import.meta.url);
 
+const VIEW_MODE = {
+  owner: 'owner',
+  collaborator: 'collaborator',
+  member: 'member'
```
ahelmberger, line 42: Just a thought: what about `nonCollaboratingMember`?

adrianaluchian: Makes sense, changed to `collaboratingMember` and `nonCollaboratingMember`.

adrianaluchian:
> Looks really nice 😍 , but I think I found an issue:
> 
> Dragging works perfectly fine for documents that were initially loaded with the page. But when I create a new document sequence in the room and then try to drag one of those new documents, they seem to fall back to their old place.
> 
> Another (but small) glitch is, that after dropping, the order reverts to the original order for a fraction of a second before turning into the new order (probably due to server communication because the glitch length is proportional to the network throttling 😬)

Nice finding! Did not think of document sequence resorting, and obviously nobody else did for some time on Production either, because the bug is already there.

I'll:
- make another ticket for this bug you uncovered
- will look into the glitch part within the current ticket/PR

adrianaluchian:
> > Looks really nice 😍 , but I think I found an issue:
> > Dragging works perfectly fine for documents that were initially loaded with the page. But when I create a new document sequence in the room and then try to drag one of those new documents, they seem to fall back to their old place.
> > Another (but small) glitch is, that after dropping, the order reverts to the original order for a fraction of a second before turning into the new order (probably due to server communication because the glitch length is proportional to the network throttling 😬)
> 
> Nice finding! Did not think of document sequence resorting, and obviously nobody else did for some time on Production either, because the bug is already there.
> 
> I'll:
> 
> * make another ticket for this bug you uncovered
> * will look into the glitch part within the current ticket/PR

Created https://educandu.atlassian.net/browse/EDU-1120 and fixed the flickering. Nice catch!

ahelmberger:
> Created https://educandu.atlassian.net/browse/EDU-1120 and fixed the flickering. Nice catch!

Perfect :+1:

ahelmberger (approved):

PR #848, opened 2023-02-10
Submit piano plugin

benewagner:
Hi Adriana and Andreas,

this is my solution for the piano plugin.
It has been renamed 'piano' since the midi functionality does not make up the major part.

I mistakenly made some commits on my local main branch. I thought listed below there would only be "submit piano plugin".

Thank you very much for your effort!

Best regards
Benni


ahelmberger:
We will integrate the plugin via npm, so we can close this PR now.

PR #850, opened 2023-02-10
EDU-1090 Enable upload to media library

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1090


src/components/resource-picker/media-library/media-library-search-screen.js
```diff
@@ -0,0 +1,193 @@
...
+          i18nKey="common:searchResultInfo"
+          values={{ resultCount: files.length, searchTerm: searchParams.searchTerm }}
+          components={[<i key="0" />]}
+          />
+      );
+    } else {
+      searchMessage = null;
+    }
+
+    return searchMessage ? <Alert type="info" message={searchMessage} showIcon /> : null;
```
adrianaluchian, line 97: Tiny tiny thing, but I think direct returns of `<Alert ..>` elements or `null` at the end would be nicer and simpler than `if-else-else` here. It's just one repetition of a straight forward element.

adrianaluchian:
There's some tweaking needed for mobile in the default screen:

Screen Shot 2023-02-13 at 11 49 33

adrianaluchian:
It looks very consistent between locations and streamlined with the wikimedia location 👏 And works like a charm! 👍 Nice work!

I would have only some small suggestions here and there:
1. When moving the mouse into the drop zone, the border effect is very nice, but might be too subtle in some cases (depending on screen size). Could we also show the drop mouse cursor maybe?

2. We could reduce length and improve readability of this message for search results based on the UX Audit suggestion to:
“Found 27 results for:violin” / "No results found for: violin" (with dropping the italic formatting)

3. Set `allowClear` on the search field, like we do in `filter-input`, for easier and more consistent clearing of the search term.

ahelmberger:
> My suggestion regarding the files grouping in the `resource-picker` folder would be to move all that is in the `shared` folder into the parent. I feel like yet another folder is redundant, and it's also the same way we keep the `components` grouped. I think it's intuitive to say that whatever is not grouped is shared.

As the resource selector folder also contains "top-level" components like the dialog that use those shared components, I have left them for now in "shared". We can think of re-organizing in a second step, I guess.

ahelmberger:
> There's some tweaking needed for mobile in the default screen:

Tweaked for mobile it is! :grin:

ahelmberger:
> 1. When moving the mouse into the drop zone, the border effect is very nice, but might be too subtle in some cases (depending on screen size). Could we also show the drop mouse cursor maybe?

I could not find a good cursor (there aren't that many), so I decided to restyle the drop area altogether and always show the border. I think that is closer to conventions. As soon as you drop a file it changes significantly.

> 2. We could reduce length and improve readability of this message for search results based on the UX Audit suggestion to:
>    “Found 27 results for:violin” / "No results found for: violin" (with dropping the italic formatting)

Great idea, done! (also for Wikimedia of course)

> 3. Set `allowClear` on the search field, like we do in `filter-input`, for easier and more consistent clearing of the search term.

Also done!

adrianaluchian:
> > 1. When moving the mouse into the drop zone, the border effect is very nice, but might be too subtle in some cases (depending on screen size). Could we also show the drop mouse cursor maybe?
> 
> I could not find a good cursor (there aren't that many), so I decided to restyle the drop area altogether and always show the border. I think that is closer to conventions. As soon as you drop a file it changes significantly.
> 
Nice, I wouldn't have thought but that actually makes it clearer, because it's a style we use nowhere else. Nice! 👍

adrianaluchian:
Another mobile case to tackle:
Screen Shot 2023-02-14 at 11 26 46

ahelmberger:
> Another mobile case to tackle

It's tackled, I checked all screens on mobile and desktop and hopefully haven't overlooked a case.

adrianaluchian:
> > Another mobile case to tackle
> 
> It's tackled, I checked all screens on mobile and desktop and hopefully haven't overlooked a case.

Nice, looks good except for this one other case we found. I would suggest that we draw the line at this phase and not drag the PR for much longer. We can fix the remaining case in the follow-up ticket. It's not a big UI issue to be a show stopper.

adrianaluchian (approved):

ahelmberger:
> > > Another mobile case to tackle
> > 
> > 
> > It's tackled, I checked all screens on mobile and desktop and hopefully haven't overlooked a case.
> 
> Nice, looks good except for this one other case we found. I would suggest that we draw the line at this phase and not drag the PR for much longer. We can fix the remaining case in the follow-up ticket. It's not a big UI issue to be a show stopper.

OK, sounds good :+1:

PR #851, opened 2023-02-13
EDU-1111 Sanitize CDN URLs in all Markdown inputs

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1111

### DIFF
--- a/src/common/github-flavored-markdown.js
+++ b/src/common/github-flavored-markdown.js
@@ -4,7 +4,7 @@ import markdownItAnchor from 'markdown-it-anchor';
 import { escapeHtml } from '../utils/string-utils.js';
 import { RESOURCE_TYPE } from '../domain/constants.js';
 import { getResourceType } from '../utils/resource-utils.js';
-import { getAccessibleUrl, isPortableCdnUrl, getCdnPath } from '../utils/source-utils.js';
+import { getAccessibleUrl, isPortableCdnUrl, getCdnPath, getPortableUrl } from '../utils/source-utils.js';
 
 // Matches both URLs in e.g.: [![alt](cdn://image.png "image title")](cdn://some-target)
 const imageInsideOfLinkPattern = /(\[!\[[^\]]*\]\()(\S*?)((?:\s+[^)]*)?\s*\)]\()(\S*?)((?:\s+[^)]*)?\s*\))(?!\])/g;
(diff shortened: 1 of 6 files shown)


src/common/github-flavored-markdown.js
```diff
@@ -104,24 +104,27 @@ class GithubFlavoredMarkdown {
...
+  }
 
-    return markdown
-      .replace(imageInsideOfLinkPattern, (_match, g1, g2, g3, g4, g5) => `${g1}${redact(g2)}${g3}${redact(g4)}${g5}`)
-      .replace(simpleLinkOrImagePattern, (_match, g1, g2, g3) => `${g1}${redact(g2)}${g3}`)
-      .replace(autoLinkPattern, (_match, g1, g2, g3) => `${g1}${redact(g2)}${g3}`);
+  makeCdnResourcesPortable(markdown, cdnRootUrl) {
+    return this._redactUrls(markdown, url => {
+      const isInternalResourceUrl = isCdnUrl({ url, cdnRootUrl });
+      return isInternalResourceUrl ? getPortableUrl({ url, cdnRootUrl }) : url;
```
adrianaluchian, line 126: Here we could even simplify to `return this._redactUrls(markdown, url => getPortableUrl({ url, cdnRootUrl }));` because `getPortableUrl` will simply return the url as is if there's no need to alter it.

src/components/markdown-input.js
```diff
@@ -67,6 +71,14 @@ function MarkdownInput({ minRows, disabled, inline, debounced, renderAnchors, va
     onChange(event);
   };
 
+  const handleBlur = event => {
+    const sanitizedValue = gfm.makeCdnResourcesPortable(value, clientConfig.cdnRootUrl);
+    if (value !== sanitizedValue) {
+      onChange({ target: { value: sanitizedValue } });
```
adrianaluchian, line 77: If the user of this even will expect an event, they might try to use functionality like `event.preventDefault()` and this would throw.
I think we should return here the real event, with only setting
```
event.target.value = sanitizedValue;
```

ahelmberger: The real event unfortunately is the wrong one (the `blur` event, while I'm trying to squeeze in a fake `change` event before the blur finishes. I think the "correct" solution would be (what also Antd does for non-DOM-conformant components) to change the argument of the `onChange` callback directly to the value instead of a DOM-like event structure with `target` etc. What do you think (it would required changing quite some usages in all plugins though)?

src/common/github-flavored-markdown.js
ahelmberger: Cool, nice simplification! Done.

src/components/markdown-input.js
adrianaluchian: I see, true, it's not the dedicated event handler there but the `onChange` one.

To be honest I don't really like having a component called `Input` using the direct value in a `onChange` callback. I noticed it could lead to problems in the `NeverScrollingTextArea` when I introduced the `debounce` toggle and had to return either the real `TextArea` or the wrapper, and had then a diverging signature for `onChange`.

In this case I would either leave it like this or try to create a [synthetic event](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event) so that we don't end up with unexpected exceptions thrown with the use of the event in the `onChange` callback. Not sure if this option works though 😬

adrianaluchian: That's even better, I like the use of the native event! 😁 👏 Good solution! Hope that fixes..2 birds with one commit 😆

adrianaluchian (approved):

PR #853, opened 2023-02-13
EDU-1120 Reload room after adding or deleting documents

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1120

### DIFF
--- a/src/server/room-controller.js
+++ b/src/server/room-controller.js
@@ -25,9 +25,11 @@ import {
 import {
   postRoomBodySchema,
   getRoomParamsSchema,
+  getRoomsQuerySchema,
   patchRoomParamsSchema,
   deleteRoomParamsSchema,
   deleteRoomsQuerySchema,
+  getRoomWithSlugParamsSchema,
   patchRoomMetadataBodySchema,
   deleteRoomMemberParamsSchema,
   postRoomInvitationsBodySchema,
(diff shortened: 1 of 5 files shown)


src/server/room-controller.js
```diff
@@ -89,6 +90,25 @@ export default class RoomController {
...
+    const { user } = req;
+    const { roomId } = req.params;
+
+    const room = await this.roomService.getRoomById(roomId);
+
+    if (!room) {
+      throw new NotFound();
+    }
+
+    if (!isRoomOwnerOrInvitedCollaborator({ room, userId: user._id })) {
```
adrianaluchian, line 103: I would change this to `!isRoomOwnerOrInvitedMember`. I did see that this is intentional after checking the tests, but neither the endpoint path nor the `handleGetRoom` handler suggest why this endpoint is dedicated to collaborators only. I think we don't have any reason to exclude members of `exclusive` rooms here actually.

We should also call `const mappedRoom = await this.clientDataMappingService.mapRoom(room, isRoomOwner ? req.user : null);` like we do when we get the page, so that the users still have access to the same members data after this API call is made.

ahelmberger: Good catch! Changed both. It took me quite a long time what the ternary operator is for in `isRoomOwner ? req.user : null`, so I moved this decision into the mapper where I think it's easier to grasp.

adrianaluchian: Cool 👍

adrianaluchian (approved):

PR #856, opened 2023-02-13
EDU-1119 Change external link source input to url-input in catalog plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1119

Attaching picture with (German) translations to check if they make sense this way.

**Case 1: Link list**

Screen Shot 2023-02-13 at 16 44 41

----

**Case 2: Image tiles**

Screen Shot 2023-02-13 at 16 44 56

### DIFF
--- a/src/plugins/catalog/catalog.yml
+++ b/src/plugins/catalog/catalog.yml
@@ -18,10 +18,10 @@ catalog:
     en: Hover effect
     de: Hover-Effekt
   hoverEffect_none:
-    en: No effect
+    en: no effect
     de: Kein Effekt
   hoverEffect_colorizeZoom:
-    en: Color and zoom
+    en: color and zoom
     de: Färben und Zoomen
   itemNumber:
     en: Item {number}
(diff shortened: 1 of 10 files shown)


src/plugins/catalog/catalog.yml
```diff
@@ -18,29 +18,29 @@ catalog:
     en: Hover effect
     de: Hover-Effekt
   hoverEffect_none:
-    en: No effect
+    en: no effect
     de: Kein Effekt
   hoverEffect_colorizeZoom:
-    en: Color and zoom
+    en: color and zoom
```
ahelmberger, line 24: I thought we were going in the direction to have all control texts in sentence case (also in English), and I remember Radu criticizing some that weren't, so I'm not sure about these two changes :thinking:

ahelmberger:
> Closes https://educandu.atlassian.net/browse/EDU-1119
> 
> Attaching picture with (German) translations to check if they make sense this way.

They totally do (and it looks very nice) :+1:

src/plugins/catalog/catalog.yml
adrianaluchian: Now we're consistent in English with all radios (control content) as lowercased and all CTAs and labels capitalized.

ahelmberger: I see, thanks!

ahelmberger (approved):

PR #860, opened 2023-02-16
EDU-1127 Fix cut screens in media library and wikimedia

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1127

I have uncovered a few more styling issues with the screens, please see the ticket for all fixed cases.


src/styles/mixins.less
```diff
@@ -245,25 +245,33 @@
...
 .m-dropzone-active {
-  .m-dropzone-transition;
-  background-color: @edu-background-color-base;
   border-style: solid;
+  border-radius: @edu-border-radius-base;
   .m-border-glow-success;
+
+  &::after {
+    content: '';
+    .m-mask;
```
ahelmberger, line 263: There is a tiny glitch in case the element scrolls (as it does for example in `MediaLibrarySearchScreen-noSearch.is-dropping` on mobile screens): The mask scrolls together with the element content (making the mask disappear at some point when scrolling down). It is not very visible since we use it on a light gray background, and - according to my research - there is no way to fix this using `::before`/`::after`. I could live with this, just weanted to point out. After all, the mask is just an additional visual indication, there's also the the outer glow.

ahelmberger:
Nice consolidations (and fixes) :heart_eyes:

ahelmberger (approved):

PR #861, opened 2023-02-16
EDU-1132 create document tag overview

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1132


adrianaluchian (approved):
Wonderful solution! 😍 
Contained and with saving the code from having to go through iterations of data curation, if the data were to be displayed in a cloud form.

PR #862, opened 2023-02-16
EDU-1140 Remove new document menu entry

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1140

### DIFF
--- a/src/components/default-page-header.js
+++ b/src/components/default-page-header.js
@@ -1,6 +1,6 @@
+import React from 'react';
 import Login from './login.js';
 import PropTypes from 'prop-types';
-import React, { useState } from 'react';
 import { Button, Dropdown } from 'antd';
 import routes from '../utils/routes.js';
 import { useUser } from './user-context.js';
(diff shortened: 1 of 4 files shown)


src/components/default-page-header.js
```diff
@@ -124,13 +98,6 @@ function DefaultPageHeader({ onUiLanguageClick }) {
           </Dropdown>
         </div>
       </div>
-      <DocumentMetadataModal
-        initialDocumentMetadata={{}}
-        isOpen={isDocumentMetadataModalOpen}
-        mode={DOCUMENT_METADATA_MODAL_MODE.create}
-        onSave={handleDocumentMetadataModalSave}
-        onClose={handleDocumentMetadataModalClose}
-        />
```
adrianaluchian, line 133: And life becomes simple again 😎

adrianaluchian (approved):

src/components/default-page-header.js
ahelmberger: So true :-)

PR #864, opened 2023-02-16
EDU-1123 Extend resource selector preview screen

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1123

### DIFF
--- a/src/components/resource-selector/media-library/media-library-edit-screen.yml
+++ b/src/components/resource-selector/media-library/media-library-edit-screen.yml
@@ -1,4 +1,4 @@
 mediaLibraryEditScreen:
   editHeadline:
-    en: Edit data
-    de: Datei bearbeiten
+    en: File editor
+    de: Datei-Editor
--- a/src/components/resource-selector/media-library/media-library-edit-screen.yml
+++ b/src/components/resource-selector/media-library/media-library-edit-screen.yml
@@ -1,0 +1,4 @@
+mediaLibraryEditScreen:
+  editHeadline:
+    en: Edit data
+    de: Datei bearbeiten
--- a/src/components/resource-selector/shared/files-grid-viewer.js
+++ b/src/components/resource-selector/shared/files-grid-viewer.js
@@ -3,6 +3,7 @@ import { Tooltip } from 'antd';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
+import EditIcon from '../../icons/general/edit-icon.js';
 import DeleteIcon from '../../icons/general/delete-icon.js';
 import { RESOURCE_TYPE } from '../../../domain/constants.js';
 import PreviewIcon from '../../icons/general/preview-icon.js';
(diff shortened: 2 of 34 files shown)


src/components/resource-selector/media-library/media-library-edit-screen.yml
```diff
@@ -0,0 +1,4 @@
+mediaLibraryEditScreen:
+  editHeadline:
+    en: Edit data
+    de: Datei bearbeiten
```
ahelmberger, line 4: The header for the upload screen is "Create media"/"Medien erstellen" and here we have "data" (=German "Daten") and "Datei" (=English "file"). Maybe we could unify the wording here. Even though we do not edit the actual media in the edit screen, I could imagine "Edit media"/"Medien bearbeiten" here. Alternatively we could use "File"/"Datei" in both?

src/components/resource-selector/shared/files-grid-viewer.js
```diff
@@ -56,10 +65,17 @@ function FilesGridViewer({
             <ActionButton
               title={t('common:preview')}
               icon={<PreviewIcon />}
-              intent={ACTION_BUTTON_INTENT.success}
```
ahelmberger, line 59: We could do the same thing in `wikimedia-files-viewer.js` (the last green preview button left).

src/components/resource-selector/shared/resource-preview-screen.yml
```diff
@@ -2,3 +2,6 @@ resourcePreviewScreen:
   headline:
     en: File preview
     de: Datei-Vorschau
+  missing:
+    en: missing
+    de: fehtl
```
ahelmberger, line 7: Should be "fehlt". But unfortunately also that doesn't work together with both the singular label "Beschreibung" and plural label "Sprachen" (because there it should be "fehlen"). Alternatives:

* "not provided/specified"/"nicht angegeben"
* "none"/"keine (Beschreibung/Sprachen)" (although it might have a language that was just not provided)

To me, the first bullet seems to fit best, because "missing" sounds rather deficient, as if something were wrong with file.

src/components/resource-selector/media-library/media-library-edit-screen.yml
adrianaluchian: Ah good point and good catch 🤔  I think "Edit media" / "Medien bearbeiten" is a good working for both cases and I'm happy if we can avoid passing down a custom title to the component.
I shall change to this.

src/components/resource-selector/shared/resource-preview-screen.yml
adrianaluchian: Agreed (also, I noticed that "fehlt" does not work for plural but didn't know what else to use). Changed, thanks!

src/components/resource-selector/shared/files-grid-viewer.js
adrianaluchian: Oh, thanks for spotting this, removed!

src/components/resource-selector/media-library/media-library-edit-screen.yml
adrianaluchian: Done (as you already know 😋 )

ahelmberger (approved):

PR #868, opened 2023-02-20
EDU-1153 Fix progress bar tooltip flickering due to overlap of tooltip with the bar

adrianaluchian:
Easily closes https://educandu.atlassian.net/browse/EDU-1153 with keeping the nice hover effect of the bar height changing.

### DIFF
--- a/src/components/media-player/media-player-progress-bar.less
+++ b/src/components/media-player/media-player-progress-bar.less
@@ -1,8 +1,6 @@
 @edu-progress-bar-height: 8px;
 @edu-progress-bar-vertical-padding: 1px;
-@edu-progress-bar-tooltip-text-height: 20px;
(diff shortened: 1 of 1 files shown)


src/components/media-player/media-player-progress-bar.less
```diff
@@ -58,7 +56,7 @@
   justify-content: center;
   flex-direction: column;
   z-index: @edu-z-index-overlay-layer;
-  top: -@edu-progress-bar-tooltip-height - 10px;
+  top: -40px;
```
adrianaluchian, line 59: There just needed to be more space between the tooltip and the tooltip trigger (aka the bar) and that space cannot be calculated in CSS because it has to consider the height of the tooltip itself.

ahelmberger (approved):

PR #870, opened 2023-02-20
EDU-1145 Replace homepageInfo with announcement

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1145


src/components/admin/admin-settings-tab.yml
```diff
@@ -1,7 +1,7 @@
 adminSettingsTab:
-  homepageInfoHeader:
-    en: Homepage info
-    de: Homepage-Info
+  announcementHeader:
+    en: Announcement
+    de: Benachrichtigung
```
ahelmberger, line 4: Maybe "Ankündigung" is better here. I also was thinking if we maybe should put the word "global" or "website" in front to make it clearer what it does?

adrianaluchian: Makes sense, I changed to:
- [en] "Global announcement"
- [de] "Globale Ankündigung"

ahelmberger: Poifect :+1:

ahelmberger (approved):

PR #871, opened 2023-02-20
EDU-1141 Add accredited editors field to public context

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1141

### DIFF
--- a/migrations/educandu-2023-02-17-02-add-accredited-editors-to-documents.js
+++ b/migrations/educandu-2023-02-17-02-add-accredited-editors-to-documents.js
@@ -1,0 +1,23 @@
+export default class Educandu_2023_02_17_02_add_accredited_editors_to_documents {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async collectionUp(collection) {
+    await collection.updateMany({ roomId: { $eq: null } }, { $set: { 'publicContext.accreditedEditors': [] } });
+  }
+
+  async collectionDown(collection) {
+    await collection.updateMany({ roomId: { $eq: null } }, { $unset: { 'publicContext.accreditedEditors': null } });
+  }
+
+  async up() {
+    await this.collectionUp(this.db.collection('documentRevisions'));
+    await this.collectionUp(this.db.collection('documents'));
+  }
+
+  async down() {
+    await this.collectionDown(this.db.collection('documentRevisions'));
+    await this.collectionDown(this.db.collection('documents'));
+  }
+}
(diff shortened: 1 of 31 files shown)


migrations/educandu-2023-02-17-02-add-accredited-editors-to-documents.js
```diff
@@ -0,0 +1,23 @@
+export default class Educandu_2023_02_17_02_add_accredited_editors_to_documents {
```
adrianaluchian, line 1: To clarify, there is a
`educandu-2023-02-17-01-add-protected-flag-to-documents.js`
migration before this and a 
`educandu-2023-02-20-01-repurpose-homepage-info-field-in-settings.js`
migration after this, on the main branch.

So this can stay named as is, it won't have a bad impact. Up to you.

ahelmberger: Perfect, I'll leave it then, it doesn't interfere with the latter one.

src/services/user-service.js
```diff
@@ -47,6 +47,39 @@ class UserService {
...
+        const sanitizedDisplayName = user.displayName.toLowerCase();
+
+        let relevance;
+        if (sanitizedEmail === sanitizedQuery || sanitizedDisplayName === sanitizedQuery) {
+          relevance = 1;
+        } else if (sanitizedEmail.startsWith(sanitizedQuery) || sanitizedDisplayName.startsWith(sanitizedQuery)) {
+          relevance = 2;
+        } else {
+          relevance = 3;
+        }
```
adrianaluchian, line 69: Very nice, considering matching relevance 👍

src/components/user-select.js
```diff
@@ -0,0 +1,76 @@
...
+import { otherUserShape } from '../ui/default-prop-types.js';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+
+const MIN_SEARCH_TEXT_LENGTH = 3;
+
+const filterOption = (inputValue, option) => option.label.toLowerCase().includes(inputValue.toLowerCase());
+
+const createUserLabel = user => user.email ? `${user.displayName} (${user.email})` : user.displayName;
+
+function UserSelect({ value, onChange, onSuggestionsNeeded, ...rest }) {
```
adrianaluchian, line 14: Super useful selector!!
We might easily include it in the room members invitation dialog, if we still have time on the project that is..

adrianaluchian (approved):

src/services/user-service.js
ahelmberger: :bow:

src/components/user-select.js
ahelmberger: I thought so too, but unfortunately this exposes the user collection which we didn't want to do for non-maintainers, so maybe we have to discuss security issues there.

PR #873, opened 2023-02-20
EDU-1148 Redesign homepage above fold

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1148

**But** [this PR ](https://github.com/educandu/educandu/pull/872) needs to be merged first 😊


src/services/document-service.js
```diff
@@ -66,6 +66,28 @@ class DocumentService {
     return documentsMetadata.sort(by(doc => doc.updatedBy, 'desc'));
   }
 
+  async getTopDocumentTags({ maxCount = 0 } = { maxCount: 0 }) {
+    const conditions = [{ roomId: null }];
```
ahelmberger, line 70: I think we also need `archived: false` here, right?

adrianaluchian: Check! ✅

ahelmberger (approved):

PR #881, opened 2023-02-23
EDU-1162 Enable batch execution of changes to user accounts

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1162


adrianaluchian:
Nice that you introduced also roles batch setup 👏 
One small thing I'd like to suggest would be to disable the submit button for the case of roles allocation if there are no roles selected, to avoid getting the backend error.
For plans allocation it makes sense to not be disabled of course since plan removal is possible, but users should have at least one role.

ahelmberger:
> Nice that you introduced also roles batch setup clap One small thing I'd like to suggest would be to disable the submit button for the case of roles allocation if there are no roles selected, to avoid getting the backend error. For plans allocation it makes sense to not be disabled of course since plan removal is possible, but users should have at least one role.

Ha, I was never aware of this, we had that also before in the table. I changed the `roles-select` to prevent deselecting all roles and always start with `user` preselected, so now it's impossible to change roles into an empty set on the UI. Thanks for pointing that out :+1:

adrianaluchian:
> > Nice that you introduced also roles batch setup clap One small thing I'd like to suggest would be to disable the submit button for the case of roles allocation if there are no roles selected, to avoid getting the backend error. For plans allocation it makes sense to not be disabled of course since plan removal is possible, but users should have at least one role.
> 
> Ha, I was never aware of this, we had that also before in the table. I changed the `roles-select` to prevent deselecting all roles and always start with `user` preselected, so now it's impossible to change roles into an empty set on the UI. Thanks for pointing that out 👍

True, was a problem before as well 😳 Great that you fixed all cases then.

adrianaluchian (approved):

PR #887, opened 2023-02-28
EDU-1159 Make navigation bar sticky

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1159

### DIFF
--- a/src/components/default-page-header.less
+++ b/src/components/default-page-header.less
@@ -1,10 +1,18 @@
-.DefaultPageHeader {
-  .m-page-content;
-}
+.DefaultPageHeader {}
 
-.DefaultPageHeader-logo {
-  width: 50px;
-  aspect-ratio: 1;
+.DefaultPageHeader-container {
+  .m-page-content;
+  background-color: @edu-white;
+  z-index: @edu-z-index-sticky-content-layer;
+
+  &.is-sticky {
+    .m-box-shadow;
+    border-radius: 0;
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+  }
 }
 
 .DefaultPageHeader-content {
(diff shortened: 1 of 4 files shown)


src/ui/hooks.js
```diff
@@ -121,3 +121,21 @@ export function useStableCallback(callback) {
   callbackRef.current = callback;
   return useCallback((...args) => callbackRef.current?.(...args), [callbackRef]);
 }
+
+export function useScrollTopOffset() {
+  const [topOffset, setTopOffset] = useState(0);
```
ahelmberger, line 126: Maybe we should add something like:
~~~
useEffect(() => {
  setTopOffset(document.documentElement.scrollTop);
}, []);
~~~
to set the initial position on the client side in case it's not 0?

src/components/default-page-header.less
```diff
@@ -1,10 +1,18 @@
...
   .m-page-content;
-}
+  background-color: @edu-white;
+  z-index: @edu-z-index-sticky-content-layer;
 
-.DefaultPageHeader-logo {
-  width: 50px;
-  aspect-ratio: 1;
+  &.is-sticky {
+    .m-box-shadow;
```
ahelmberger, line 9: I added `transition: all 300ms ease;` and it made the switch a little bit smoother I feel, so maybe you wanna give it a try. It's not a huge difference though.

ahelmberger:
Nice and clean :-)

src/ui/hooks.js
adrianaluchian: Good point, added!

src/components/default-page-header.less
adrianaluchian: That's a good idea, and it feels noticeable to me 👍 Done.

ahelmberger: Cool :+1:

ahelmberger (approved):

PR #888, opened 2023-02-28
EDU-1169 create notifications from events

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1169


src/stores/database.js
```diff
@@ -125,16 +125,8 @@ class Database {
...
+    const umzug = await this._generateUmzug();
     await umzug.up();
-    this.getSchemaHash.clear();
-  }
-
-  async _generateSchemaHash() {
-    const migrations = await this._db.collection('migrations').find({}).toArray();
-    const migrationNames = migrations.map(migration => migration.migrationName).sort().join();
-
-    return md5(migrationNames);
```
adrianaluchian, line 137: Indeed, no longer needed 👍

src/stores/user-store.js
```diff
@@ -24,6 +24,10 @@ class UserStore {
     return this.collection.find({ $and: [{ accountClosedOn: null, $or: searchCriteria }] }, { session }).toArray();
   }
 
+  iterateAllActiveUsers({ session } = {}) {
```
adrianaluchian, line 27: Small thing, how about `getActiveUsersIterator`?

src/services/event-service.js
```diff
@@ -0,0 +1,202 @@
...
+    } finally {
+      await userIterator.close();
+    }
+  }
+
+  async _processCommentCreatedEvent(event, session, context) {
+    const userIterator = this.userStore.iterateAllActiveUsers({ session });
+
+    try {
+      const { documentId, roomId } = event.params;
```
adrianaluchian, line 93: Actually we don't even need the `roomId` in this case, because we have it on the `document`. It's even maybe room for discrepancy if we pass it separately? 🤔

src/utils/notification-utils.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    const testCases = [
+      {
+        description: 'when the notified user is the owner of the room',
+        event: { createdOn: new Date('2023-01-01'), type: EVENT_TYPE.revisionCreated, params: { userId: 'event-user-id' } },
+        revision: { _id: 'revision-id', documentId: 'document-id', roomId: 'room-id', roomContext: { draft: false } },
+        document: { _id: 'document-id', roomId: 'room-id', roomContext: { draft: false } },
+        room: { _id: 'room-id', owner: 'owner-user-id', members: [{ userId: 'notified-user-id' }] },
+        notifiedUser: { _id: 'owner-user-id', createdOn: new Date('2022-12-31'), favorites: [] },
+        expectedReasons: [NOTIFICATION_REASON.roomMembership]
+      },
```
adrianaluchian, line 17: The negative cases of "notified user is the same as event user", "room/document/revision deleted" are not covered from what I can see, would be nice to have those too if it's not a hassle to add them.

src/services/event-service.js
ahelmberger: Hm, that's true :thinking:

On the other hand, then we could also argue that we do not need the `documentId` as it is already on the revision (at least for the `revision-created` type of event). It would make things harder though when assembling the all the data needed to show user notifications (as the notifications use the same event params). We would have to do multiple round of fetches to the db. So I can see both a case for removing and keeping the redundant data. Tending more towards keeping. What do you think?

src/utils/notification-utils.spec.js
ahelmberger: For the first scenario we do have both positive and negative cases:

* `when the notified user has marked the document as favorite`
* `when the user has marked the document as favorite but the notified user is the same user as the event user` (I forgot one `notified there`)

We do not have a test for deleted documents though, I'll add them (and correct the wording above).

src/stores/user-store.js
ahelmberger: Sure, sounds better, renamed!

src/utils/notification-utils.spec.js
ahelmberger: Added!

adrianaluchian: 👍

src/services/event-service.js
adrianaluchian: Ok then, let's keep them.

adrianaluchian (approved):

PR #893, opened 2023-03-01
EDU-1177 Add cron job to delete old notifications

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1177


src/services/notification-service.js
```diff
@@ -1,6 +1,9 @@
+import moment from 'moment';
 import NotificationStore from '../stores/notification-store.js';
 import { groupNotifications } from '../utils/notification-utils.js';
 
+const MAX_NOTIFICATION_LIFETIME_IN_MONTHS = 6;
```
adrianaluchian, line 5: We could put this in `constants.js` to make it easier to find, alongside the other such setup.

ahelmberger: Yep, moved!

ahelmberger:
I reworked this to just simply use a MongoDB TTL index (but kept the CronJob scheduler alive)

adrianaluchian:
> I reworked this to just simply use a MongoDB TTL index (but kept the CronJob scheduler alive)

Perfect! 👌

adrianaluchian (approved):

PR #895, opened 2023-03-02
EDU-1183 Make notifications dismissable

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1183

### DIFF
--- a/src/components/dashboard/notifications-tab.yml
+++ b/src/components/dashboard/notifications-tab.yml
@@ -1,7 +1,7 @@
 notificationsTab:
   info:
-    en: Here you can find your notifications.
-    de: Hier findest du deine Benachrichtigungen.
+    en: Here you can find your notifications about contibutions to your favorite documents, by your favorite users or to documents in rooms that you are a member of.
+    de: Hier findest du deine Benachrichtigungen über Beiträge an deine Lieblingsdokumente, von deine Lieblingsnutzern oder an Dokumente in Räume, in denen du Mitglied bist.
   noNotifications:
     en: There are no new notifications for you.
     de: Es gibt keine neuen Benachrichtigungen für dich.
--- a/src/components/dashboard/notifications-tab.yml
+++ b/src/components/dashboard/notifications-tab.yml
@@ -4,7 +4,7 @@ notificationsTab:
     de: Hier findest du deine Benachrichtigungen.
   noNotifications:
     en: There are no new notifications for you.
-    de: Es gibt keine neue Benachrichtigungen für dich.
+    de: Es gibt keine neuen Benachrichtigungen für dich.
   publicRevisionCreatedNotification:
     en: Public document has been edited
     de: Öffentliches Dokument wurde bearbeitet
@@ -20,9 +20,9 @@ notificationsTab:
   documentNotAvailable:
     en: Document no longer available
     de: Document no longer available
-  dismiss:
-    en: Dismiss
-    de: Verwerfen
-  dismissAll:
-    en: Dismiss all
-    de: Alle verwerfen
+  remove:
+    en: Remove
+    de: Entfernen
+  removeAll:
+    en: Remove all
+    de: Alle entfernen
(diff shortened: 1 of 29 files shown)


src/components/dashboard/notifications-tab.yml
```diff
@@ -1,13 +1,28 @@
 notificationsTab:
   info:
-    en: Here you can find your notifications.
-    de: Hier findest du deine Benachrichtigungen.
+    en: Here you can find your notifications about contibutions to your favorite documents, by your favorite users or to documents in rooms that you are a member of.
+    de: Hier findest du deine Benachrichtigungen über Beiträge an deine Lieblingsdokumente, von deine Lieblingsnutzern oder an Dokumente in Räume, in denen du Mitglied bist.
```
ahelmberger, line 4: Hier findest du Benachrichtigungen über Beiträge zu deinen Lieblingsdokumenten, von deine Lieblingsnutzern, oder zu Dokumenten in Räumen, in denen du Mitglied bist.

Maybe we should also scratch the first "your" in the english version?

src/components/dashboard/notifications-tab.yml
```diff
@@ -1,13 +1,28 @@
...
+    de: Raum-Dokument wurde bearbeitet
+  publicCommentCreatedNotification:
+    en: New comment on public document
+    de: Neuer Kommentar am öffentliches Dokument
+  roomCommentCreatedNotification:
+    en: New comment on room document
+    de: Neuer Kommentar am Raum-Dokument
+  documentNotAvailable:
+    en: Document no longer available
+    de: Document no longer available
```
ahelmberger, line 22: Dokument nicht mehr verfügbar

ahelmberger:
Looks super nice, I love it (also the activities) :heart: 

There's one logical thing that I would improve: in my opinion, the "remove all" should only remove the notifications that the user currently sees on the UI, not what's currently in the DB, meaning that we should only use `deleteNotificationsByNotifiedUserIdAndNotificationIds` and would not need `deleteNotificationsByNotifiedUserId`. Otherwise users would miss notifications that were created during their page load and the time they clicked the "remove all" button. Watcha think?

src/components/dashboard/notifications-tab.yml
adrianaluchian: Actually I knew this one 🤦‍♀️ Thanks!

adrianaluchian:
> Looks super nice, I love it (also the activities) ❤️
> 
> There's one logical thing that I would improve: in my opinion, the "remove all" should only remove the notifications that the user currently sees on the UI, not what's currently in the DB, meaning that we should only use `deleteNotificationsByNotifiedUserIdAndNotificationIds` and would not need `deleteNotificationsByNotifiedUserId`. Otherwise users would miss notifications that were created during their page load and the time they clicked the "remove all" button. Watcha think?

Thank you!! ☺️ 

And super good point about removing currently visible notifications! I was thinking we could use the timestamp of the "remove all" request and remove the notifications created before this date, but that's not enough either. So I will change to actually use the same endpoint as for notification group removal and just pass all the ids from the UI.

ahelmberger:
> > Looks super nice, I love it (also the activities) heart
> > There's one logical thing that I would improve: in my opinion, the "remove all" should only remove the notifications that the user currently sees on the UI, not what's currently in the DB, meaning that we should only use `deleteNotificationsByNotifiedUserIdAndNotificationIds` and would not need `deleteNotificationsByNotifiedUserId`. Otherwise users would miss notifications that were created during their page load and the time they clicked the "remove all" button. Watcha think?
> 
> Thank you!! relaxed
> 
> And super good point about removing currently visible notifications! I was thinking we could use the timestamp of the "remove all" request and remove the notifications created before this date, but that's not enough either. So I will change to actually use the same endpoint as for notification group removal and just pass all the ids from the UI.
> 
> **Update:** Fixed (along with another new issue for the case of deleted documents). Thanks again for pointing out this use case, good one! +1

Poifect :+1:

ahelmberger (approved):

PR #897, opened 2023-03-02
EDU-1185 add email notification frequency to users

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1185

### DIFF
--- a/src/server/user-controller.js
+++ b/src/server/user-controller.js
@@ -38,7 +38,8 @@ import {
   favoriteBodySchema,
   loginBodySchema,
   externalAccountIdParamsSchema,
-  getUsersBySearchQuerySchema
+  getUsersBySearchQuerySchema,
+  postUserNotificationSettingsBodySchema
 } from '../domain/schemas/user-schemas.js';
 
 const jsonParser = express.json();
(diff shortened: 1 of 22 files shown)


src/services/storage-service.spec.js
```diff
@@ -568,11 +568,11 @@ describe('storage-service', () => {
 
       describe('and the user is room collaborator and the room owner does not have a storage plan', () => {
         beforeEach(async () => {
-          const collaboratorUser = await setupTestUser(container, {
+          const collaboratorUser = await createTestUser(container, {
```
adrianaluchian, line 571: Thanks for this, I noticed the naming discrepancy a few times but never streamlined!

src/server/user-controller.js
```diff
@@ -310,6 +311,18 @@ class UserController {
...
+  async handlePostUserNotificationSettings(req, res) {
+    const userId = req.user._id;
+    const { emailNotificationFrequency } = req.body;
+    const updatedUser = await this.userService.updateUserNotificationSettings({ userId, emailNotificationFrequency });
+
+    if (!updatedUser) {
+      throw new NotFound();
+    }
+
+    res.status(201).send({ user: updatedUser });
```
adrianaluchian, line 323: For consistency and to avoid possible future issues (though right now I don't think there are issues), let's do a `this.clientDataMappingService.mapWebsiteUser(updatedUser)` mapping, like we do in all other cases where we return the data of the logged in user (equivalently for a viewed user we use `mapWebsitePublicUser`)

ahelmberger: Very good point, thanks! Changed it (here and also in the case for the profile update).

src/services/storage-service.spec.js
ahelmberger: I had to change the function code anyways, so I thought, it's time :smile:

adrianaluchian (approved):

src/server/user-controller.js
adrianaluchian: Wunderbar 👍

PR #903, opened 2023-03-06
EDU-1194 We give to you: The Separator

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1194

Screen Shot 2023-03-06 at 13 15 35


src/plugins/divider/divider.yml
```diff
@@ -0,0 +1,28 @@
+divider:
+  name:
+    en: Divider
+    de: Verteiler
```
ahelmberger, line 4: Hm, that would be "distributor". Divider would be "Teiler", though it seems to me to not be very clear what the plugin does. Clearer would be "Trennline" (separator / separator line). Maybe "separator" would be also clearer in English, what do you think?

adrianaluchian: I wasn't sure about divider or separator. Then I went with divider because AntD also calls it that 🤷‍♀️ 
We can change to separator, sure.

adrianaluchian: Changed!

ahelmberger: :+1:

ahelmberger (approved):

PR #915, opened 2023-03-07
EDU-1203 Fix document edit permissions for maintainers

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1203

### DIFF
--- a/src/server/document-controller.js
+++ b/src/server/document-controller.js
@@ -9,12 +9,12 @@ import { canEditDoc } from '../utils/doc-utils.js';
 import RoomService from '../services/room-service.js';
 import { shuffleItems } from '../utils/array-utils.js';
 import SettingService from '../services/setting-service.js';
+import { DOC_VIEW_QUERY_PARAM } from '../domain/constants.js';
 import DocumentService from '../services/document-service.js';
 import needsPermission from '../domain/needs-permission-middleware.js';
+import { isRoomOwner, isRoomOwnerOrInvitedMember } from '../utils/room-utils.js';
 import ClientDataMappingService from '../services/client-data-mapping-service.js';
-import { DOC_VIEW_QUERY_PARAM, NOT_ROOM_OWNER_ERROR_MESSAGE } from '../domain/constants.js';
 import { validateBody, validateParams, validateQuery } from '../domain/validation-middleware.js';
-import { isRoomOwner, isRoomOwnerOrInvitedCollaborator, isRoomOwnerOrInvitedMember } from '../utils/room-utils.js';
 import {
   documentIdParamsOrQuerySchema,
   updateDocumentMetadataBodySchema,
@@ -29,7 +29,7 @@ import {
   getSearchableDocumentsTitlesQuerySchema
 } from '../domain/schemas/document-schemas.js';
 
-const { NotFound, BadRequest, Forbidden, Unauthorized } = httpErrors;
+const { NotFound, Forbidden, Unauthorized } = httpErrors;
 
 const jsonParser = express.json();
 const jsonParserLargePayload = express.json({ limit: '2MB' });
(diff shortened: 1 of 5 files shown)


src/server/document-controller.js
```diff
@@ -421,18 +331,6 @@ class DocumentController {
...
-      [needsPermission(permissions.ARCHIVE_DOC), validateParams(documentIdParamsOrQuerySchema)],
-      (req, res) => this.handlePatchDocArchive(req, res)
-    );
-
-    router.patch(
-      '/api/v1/docs/:documentId/unarchive',
-      [needsPermission(permissions.ARCHIVE_DOC), validateParams(documentIdParamsOrQuerySchema)],
-      (req, res) => this.handlePatchDocUnarchive(req, res)
-    );
-
```
adrianaluchian, line 435: Glad that you could ditch these 2 endpoints now.

src/services/document-service.js
```diff
@@ -23,6 +23,8 @@ import { DOCUMENT_VERIFIED_RELEVANCE_POINTS } from '../domain/constants.js';
 import { documentDBSchema, documentRevisionDBSchema } from '../domain/schemas/document-schemas.js';
 import { checkRevisionOnDocumentCreation, checkRevisionOnDocumentUpdate } from '../utils/revision-utils.js';
 import { createSectionRevision, extractCdnResources, validateSection, validateSections } from './section-helper.js';
+import { isRoomOwner } from '../utils/room-utils.js';
+import permissions, { hasUserPermission } from '../domain/permissions.js';
```
adrianaluchian, line 27: Some crazy 💩 happened here 😅

src/services/document-service.js
```diff
@@ -234,6 +236,10 @@ class DocumentService {
           throw new Forbidden(error.message);
         }
 
+        const actualNewOrder = await this.documentOrderStore.getNextOrder();
```
adrianaluchian, line 239: I presume here you thought that moving this closer to the moment of saving the data makes it less likely that another order is generated in the meantime, right? 🤔 
Can we maybe completely remove the preliminary value and only set the order directly once?

src/services/document-service.js
```diff
@@ -300,6 +301,10 @@ class DocumentService {
...
+        newRevision.order = actualNewOrder;
```
adrianaluchian, line 305: Same here?

src/services/document-service.js
```diff
@@ -431,29 +449,41 @@ class DocumentService {
...
+
+        try {
+          checkRevisionOnDocumentUpdate({ previousRevision, newRevision, room, user });
+        } catch (error) {
+          logger.error(error);
+          throw new Forbidden(error.message);
+        }
+
+        const actualNewOrder = await this.documentOrderStore.getNextOrder();
+        newRevision.order = actualNewOrder;
```
adrianaluchian, line 486: And here?

src/services/document-service.js
ahelmberger: :scream: Let's pretend that never happened

adrianaluchian: 💊 ... What never happened?

src/services/document-service.js
ahelmberger: Actually no. The reasoning was that in case the validation above fails we would have consumed a next order number from the db and throw it away. So I set a preliminary one for validation, and only if it passes, get a "real" one from the DB. I was not sure myself if this optimization is really necessary, given that we could just throw away some numbers, but if for some reason we had a lot of repeated failed update attempts, we would skip a lot of order numbers. I don't have a strong opinion on it though. If you think the code simplification would out-benefit my point, I'll gladly change it.

adrianaluchian: Got it, makes sense, good thinking 👍 
Then let's just drop the preliminary value, I think the code will be more straight forwards.

ahelmberger: It's dropped!

src/server/document-controller.js
ahelmberger: :+1:

src/services/document-service.js
ahelmberger: :sunglasses:

adrianaluchian (approved):

PR #917, opened 2023-03-08
EDU-1206 Enable configuring email receiver ignore pattern

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1206


adrianaluchian (approved):
Nice, let's also mark the ticket with `configUpdate` and maybe remove the `nontestable` label and ask Theresa to test that other emails that `@test.com` ones are sent (maybe describe a setup for daily notification email and ask her to check that the email is sent).

ahelmberger:
> Nice, let's also mark the ticket with `configUpdate` and maybe remove the `nontestable` label and ask Theresa to test that other emails that `@test.com` ones are sent (maybe describe a setup for daily notification email and ask her to check that the email is sent).

It actually doesn't require any config update as the default value covers our case. Good idea to let Theresa test it, because she can actually use the test@test.com user and another non-"@test.com" user to check for negative and positive case. I'll update the ticket accordingly.

PR #920, opened 2023-03-08
EDU-1211 Bundle user permissions

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1211

### DIFF
--- a/src/api-clients/document-api-client.js
+++ b/src/api-clients/document-api-client.js
@@ -122,7 +122,7 @@ class DocumentApiClient {
       .then(res => res.data);
   }
 
-  hardDeleteDocument(documentId) {
+  hardDeletePrivateDocument(documentId) {
     return this.httpClient
       .delete(
         '/api/v1/docs',
--- a/src/domain/permissions.js
+++ b/src/domain/permissions.js
@@ -1,70 +1,35 @@
 import { ROLE } from './constants.js';
 
-const ADMIN = 'admin';
-const EDIT_DOC = 'edit-doc';
-const VIEW_DOCS = 'view-docs';
-const EDIT_FILE = 'edit-file';
-const OWN_ROOMS = 'own-rooms';
-const REVIEW_DOC = 'review-doc';
-const VERIFY_DOC = 'verify-doc';
-const VIEW_FILES = 'view-files';
-const JOIN_ROOMS = 'join-rooms';
-const CREATE_FILE = 'create-file';
-const ARCHIVE_DOC = 'archive-doc';
-const MIGRATE_DATA = 'migrate-data';
-const SEARCH_USERS = 'search-users';
+const VIEW_USERS = 'view-users';
+const VIEW_CONTENT = 'view-content';
 const MANAGE_USERS = 'manage-users';
-const MANAGE_BATCHES = 'manage-batches';
-const SEE_USER_EMAIL = 'see-user-email';
-const MANAGE_CONTENT = 'manage-content';
-const PROTECT_ANY_DOC = 'protect-any-doc';
-const PROTECT_OWN_DOC = 'protect-own-doc';
-const MANAGE_SETTINGS = 'manage-settings';
-const DELETE_OWN_FILES = 'delete-own-files';
-const HARD_DELETE_SECTION = 'hard-delete-section';
-const DELETE_FOREIGN_ROOMS = 'delete-foreign-rooms';
-const MANAGE_STORAGE_PLANS = 'manage-storage-plans';
-const DELETE_ANY_STORAGE_FILE = 'delete-storage-file';
-const RESTORE_DOC_REVISIONS = 'restore-doc-revisions';
-const CREATE_DOCUMENT_COMMENTS = 'create-document-comments';
-const MANAGE_DOCUMENT_COMMENTS = 'manage-document-comments';
-const AUTORIZE_ROOMS_RESOURCES = 'authorize-room-resources';
-const MANAGE_ACCREDITED_EDITORS = 'manage-accredited-editors';
+const MANAGE_SETUP = 'manage-setup';
+const CREATE_CONTENT = 'create-content';
+const BATCH_PROCESS_DATA = 'batch-process-data';
+const MANAGE_PUBLIC_CONTENT = 'manage-public-content';
+const PROTECT_OWN_PUBLIC_CONTENT = 'protect-own-public-content';
+const DELETE_OWN_PRIVATE_CONTENT = 'delete-own-private-content';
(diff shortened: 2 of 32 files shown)


src/api-clients/document-api-client.js
```diff
@@ -122,7 +122,7 @@ class DocumentApiClient {
       .then(res => res.data);
   }
 
-  hardDeleteDocument(documentId) {
+  hardDeletePrivateDocument(documentId) {
```
adrianaluchian, line 125: I wanted to make this clear because I kept thinking it's a leftover from when public external documents could be deleted.
Also cause of the restriction of the document being private, inside the service, of course.

ahelmberger: Definitely :+1:

src/domain/permissions.js
```diff
@@ -1,70 +1,35 @@
...
-  EDIT_FILE,
-  VIEW_FILES,
-  DELETE_OWN_FILES,
-  CREATE_FILE,
-  OWN_ROOMS,
-  AUTORIZE_ROOMS_RESOURCES,
-  JOIN_ROOMS,
-  CREATE_DOCUMENT_COMMENTS
+  CREATE_CONTENT,
+  ACCESS_STORAGE,
```
ahelmberger, line 17: Given the very nice logic this PR introduces (namely that the permission names now always clearly describe some action that goes beyond any action that any anonymous user could perform anyways), the name of this permission could be `BROWSE_STORAGE` or `SEARCH_STORAGE`, because anyone can access it anyways (except for room resources). And if you search for usages of this permission, you find it is always used in the scenario of actively searching a file by some means. Except for this one instance: in the `/api/v1/rooms/:roomId/authorize-resources-access` endpoint. But honestly, we could remove it from there anyways, because the actually deciding factor in this endpoint is the user's room membership and therefore access to specific a room resource, not storage access in general, so logically it is a little misleading anyways. What do you think?

src/domain/permissions.js
```diff
@@ -1,70 +1,35 @@
...
-    ARCHIVE_DOC,
-    REVIEW_DOC,
-    VERIFY_DOC,
-    PROTECT_ANY_DOC,
-    MANAGE_DOCUMENT_COMMENTS,
-    MANAGE_CONTENT,
-    MANAGE_ACCREDITED_EDITORS
+    VIEW_USERS,
+    MANAGE_PUBLIC_CONTENT,
+    MANAGE_PROTECTORS_OF_OWN_PUBLIC_CONTENT
```
ahelmberger, line 32: Honestly, curating the list of accredited authors on a document seems to me very much in the same category as setting the protected flag or writing review text etc. In other words, I would also get rid of this fine-grained permission and just let users with `MANAGE_PUBLIC_CONTENT` (un)assign the accreited authors. That would also solve the conundrum that the users in this list do not necessarily have to have `PROTECT_OWN_PUBLIC_CONTENT` permissions. A maintainer can add any old user in there.

src/domain/permissions.js
adrianaluchian: I completely agree and the authorization endpoint was the only reason why I tried to come up with a permission name to fit this case as well.
So then we can go with `BROWSE_STORAGE` on all endpoints, except for the authorization one, where we can switch to `needsAuthentication()`, right?

src/utils/revision-utils.js
```diff
@@ -23,47 +23,43 @@ export function checkRevisionOnDocumentCreation({ newRevision, room, user }) {
...
-    const userCanProtectOwnNewDoc = hasUserPermission(user, permissions.PROTECT_OWN_DOC);
-    const userCanProtectAnyDoc = hasUserPermission(user, permissions.PROTECT_ANY_DOC);
-    const userCanArchiveDoc = hasUserPermission(user, permissions.ARCHIVE_DOC);
-    const userCanVerifyDoc = hasUserPermission(user, permissions.VERIFY_DOC);
-    const userCanReviewDoc = hasUserPermission(user, permissions.REVIEW_DOC);
-
     const { accreditedEditors } = publicContext;
+    const userCanManagePublicContext = hasUserPermission(user, permissions.MANAGE_PUBLIC_CONTENT);
+    const userCanProtectOwnDocWhenCreating = hasUserPermission(user, permissions.PROTECT_OWN_PUBLIC_CONTENT);
+    const userCanDecideWhoCanProtectOwnDocWhenCreating = hasUserPermission(user, permissions.MANAGE_PROTECTORS_OF_OWN_PUBLIC_CONTENT);
```
ahelmberger, line 29: Technically that's not correct (like the permission name, see also my other comment): it should be rather `userCanFreelyModifyTheListOfAccreditedAuthorsForThisDocument`, but as stated in my other comment above, the simplifications in this PR would actually lend to just bundle this ability under `MANAGE_PUBLIC_CONTENT` (and therefore not needing this third boolean flag at all).

adrianaluchian: Agreed 🤝

src/domain/permissions.js
```diff
@@ -100,34 +62,15 @@ export function getAllUserPermissions(user) {
...
-  AUTORIZE_ROOMS_RESOURCES,
-  JOIN_ROOMS,
-  CREATE_DOCUMENT_COMMENTS,
-  MANAGE_DOCUMENT_COMMENTS,
-  MANAGE_CONTENT,
-  MANAGE_ACCREDITED_EDITORS
+  MANAGE_SETUP,
+  BATCH_PROCESS_DATA,
+  DELETE_ANY_PRIVATE_CONTENT
 };
```
ahelmberger, line 76: If you're fond of the cleanliness and look at these changes, you feel a little bit like in heaven :heart:

src/domain/permissions.js
ahelmberger: Exactly! :+1: :+1: :+1:

ahelmberger:
Thanks for this thorough and long overdue radical simplification! It boils it down to the essentials! Great job :+1:

src/domain/permissions.js
adrianaluchian: Which both of us are found of, yes! It's indeed a breath of fresh air.
Simplicity, cleanliness, straight-forwardness, with so much less risk of misuse and bugs ❤️

adrianaluchian:
> Thanks for this thorough and long overdue radical simplification! It boils it down to the essentials! Great job 👍

Thanks 🙇‍♀️ My pleasure, definitely! I'm very glad we are doing such changes at this point in the project, with accumulated clarity of the direction of features and functionality. A good chance for radical code improvements.

src/domain/permissions.js
adrianaluchian: Great, changed it just so.

ahelmberger (approved):

PR #921, opened 2023-03-08
EDU-1215 Room improvements

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1215

**Example with all types of rooms on Dashboard**

Screen Shot 2023-03-08 at 20 15 03

### DIFF
--- a/src/components/dashboard/rooms-tab.yml
+++ b/src/components/dashboard/rooms-tab.yml
@@ -1,13 +1,13 @@
 roomsTab:
   info:
     en: Here you find all the rooms you have created and the rooms that were invited to.
-    de: Hier findest du alle Räume, die du erstellt hast und in denen du eingeladen wurdest.
+    de: Hier findest du alle Räume, die du erstellt hast und in die du eingeladen wurdest.
   ownedRoomsHeadline:
     en: Rooms that you created
     de: Räume, die du erstellt hast
   memberRoomsHeadline:
     en: Rooms that you were invited to
-    de: Räume, in denen du eingeladen wurdest
+    de: Räume, in die du eingeladen wurdest
   noOwnedRooms:
     en: You have not created any rooms yet.
     de: Du hast noch keine Räume erstellt.
--- a/src/components/dashboard/rooms-tab.yml
+++ b/src/components/dashboard/rooms-tab.yml
@@ -1,28 +1,13 @@
 roomsTab:
   info:
-    en: Here you find all the rooms you have created and the rooms that you are a member of.
-    de: Hier findest du alle Räume, die du erstellt hast und in denen du Mitglied bist.
-  joined:
-    en: Joined
-    de: Eingetreten
-  role:
-    en: Role
-    de: Rolle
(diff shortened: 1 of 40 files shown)


src/components/dashboard/rooms-tab.yml
```diff
@@ -1,28 +1,13 @@
...
-    en: Member
-    de: Mitglied
-  sortedByOwner:
-    en: Sorted by owner
-    de: Nach Eigentümer sortiert
-  sortedByRole:
-    en: Sorted by role
-    de: Nach Rolle sortiert
+    en: Here you find all the rooms you have created and the rooms that were invited to.
+    de: Hier findest du alle Räume, die du erstellt hast und in denen du eingeladen wurdest.
```
ahelmberger, line 4: .. und in die du eingeladen wurdest.

src/components/dashboard/rooms-tab.yml
```diff
@@ -1,28 +1,13 @@
...
   ownedRoomsHeadline:
     en: Rooms that you created
     de: Räume, die du erstellt hast
   memberRoomsHeadline:
-    en: Rooms that you are a member of
-    de: Räume, in denen du Mitglied bist
+    en: Rooms that you were invited to
+    de: Räume, in denen du eingeladen wurdest
```
ahelmberger, line 10: .. in die du ..

src/components/dashboard/rooms-tab.yml
adrianaluchian: Thanks for correcting (and explaining to me). Changed.

src/components/dashboard/rooms-tab.yml
adrianaluchian: Changed as well.

ahelmberger:
There is one wording that I find sounds not that nice: the label "is collaborative" next to the checkbox. I'd suggest either one of the two changes:

* remove the tooltip and put the longer explanation directly next to the checkbox as a label, like e.g.: "Allow invited users to create and modify documents in this room" (optionally with a "but not delete" thrown in)
* keep the tooltip explanation and rephrase the label to
    * "collaborative (work) mode" (I know, sounds a bit technical)
    * "work collaboratively"

Watch think?

adrianaluchian:
> There is one wording that I find sounds not that nice: the label "is collaborative" next to the checkbox. I'd suggest either one of the two changes:
> 
> * remove the tooltip and put the longer explanation directly next to the checkbox as a label, like e.g.: "Allow invited users to create and modify documents in this room" (optionally with a "but not delete" thrown in)
> * keep the tooltip explanation and rephrase the label to one of ...
>   
>   * "collaborative (work) mode" (I know, sounds a bit technical)
>   * "work collaboratively"
> 
> Watch think?

Either option is better than "Is collaborative" / "Ist Kollaborativ". Changed to "Collaborative mode" / "Kollaborativer Modus", as discussed.

ahelmberger (approved):

PR #922, opened 2023-03-09
EDU-1216 Prevent own admin role removal

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1216

Screen Shot 2023-03-09 at 17 27 55


adrianaluchian:
_Adriana updated this Pull Request_

src/domain/schemas/user-schemas.js
```diff
@@ -105,7 +105,7 @@ export const userDBSchema = joi.object({
   _id: idOrKeySchema.required(),
   passwordHash: joi.string().allow(null).required(),
   email: joi.string().case('lower').allow(null).required(),
-  roles: joi.array().required().items(joi.string()),
+  role: joi.string().required(),
```
ahelmberger, line 108: We also could restrict here to: `joi.string().valid(...Object.values(ROLE)).required()`

ahelmberger (approved):

src/domain/schemas/user-schemas.js
adrianaluchian: Indeed, added!
Apropo to the validity of the previous `roles` array, there were a few of empty arrays in the Prod DB 😬

ahelmberger:
Another nice simplification :+1:

PR #926, opened 2023-03-09
EDU-1218 Include media tags in redaction page tags tab

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1218


adrianaluchian (approved):
Was a good idea to do this, actually 👍

PR #932, opened 2023-03-14
EDU-1219 Manage own private storage (except for uploading)

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1219


src/server/storage-controller.js
```diff
@@ -206,6 +220,19 @@ class StorageController {
...
+      needsPermission(permissions.BROWSE_STORAGE),
+      (req, res) => this.handleGetRoomMediaOverview(req, res)
+    );
+
+    router.delete(
+      '/api/v1/storage/room-media/:roomId/:name',
+      needsPermission(permissions.DELETE_OWN_PRIVATE_CONTENT),
+      validateParams(deleteRoomMediaParamsSchema),
+      (req, res) => this.handleDeleteRoomMedia(req, res)
+    );
```
ahelmberger, line 235: By introducing this set of simplified CRUD operations on room media, we can get rid of:
* the more complicated operations based on path analysis, we can just fetch, upload and delete by roomId + file name
* the two middlewares in `storage-controller` that inject the storage and storage plan into each request, as these endpoints always return all information needed (storage plan, currently used bytes)
* relying on the `user.storage.usedBytes` (and updating it live on the UI), in the future we'll not even have to map it and send it to the browser
* the extra loop for contributors to show the storage quota of a different user (when all API calls are bound to room ID, the storage plan and quota in the response is always set correctly to that of the room owner, even if the API call was made by a contributor)

I will create a ticket to describe the necessary changes.

src/stores/user-store.js
```diff
@@ -74,6 +74,17 @@ class UserStore {
...
   }
 
+  async updateUserUsedBytes({ userId, usedBytes }, { session } = {}) {
+    const filter = { _id: userId };
+    const update = { $set: { 'storage.usedBytes': usedBytes } };
+    const options = { session, returnDocument: 'after' };
+
+    const { value } = await this.collection.findOneAndUpdate(filter, update, options);
+
+    validate(value, userDBSchema);
```
adrianaluchian, line 84: I presume this is a leftover, otherwise I'm not sure why we should validate a value returned from the database against our expected schema, since the DB data is already updated at this point.

src/services/storage-service.js
```diff
@@ -228,8 +285,8 @@ export default class StorageService {
     }
 
     const room = await this.roomStore.getRoomById(doc.roomId);
-    const isRoomOwner = user._id === room.owner;
-    const roomOwner = isRoomOwner ? user : await this.userStore.getUserById(room.owner);
+    const isUserRoomOwner = user._id === room.owner;
```
adrianaluchian, line 288: This variable name reminds me of something 🤔

src/services/storage-service.js
```diff
@@ -216,6 +216,63 @@ export default class StorageService {
...
+      return { room, objects, usedBytes: objects.reduce((accu, obj) => accu + obj.size, 0) };
+    }));
+
+    const totalUsedBytes = roomsAndObjects.reduce((accu, { usedBytes }) => accu + usedBytes, 0);
+
+    return {
+      storagePlan: storagePlan || null,
+      usedBytes: totalUsedBytes,
+      unusedBytes: Math.max(0, (storagePlan?.maxBytes || 0) - totalUsedBytes),
+      rooms: roomsAndObjects.map(({ room, objects, usedBytes }) => ({
```
adrianaluchian, line 233: Maybe I'd call this `storagePerRoom`/`roomStorages` or something of sorts, other than `rooms`. I find it risky to reuse entity names because we might find ourselves somewhere else supposing that items in this array have `_id`, `_name` etc.

adrianaluchian: Meantime I did notice that there would be repetition in usage (`storage.roomStorages`), in the end just like for `storage.storagePlan`, but I do think it would be much clearer in cases like

```
 const newRoomOptions = storage.rooms
      .map(room => ({ value: room.roomId, label: room.roomName }))
```

where if I look at the code I really might expect something else inside a `room` variable and misuse it 🤔 

Does this make sense?

src/server/storage-controller.js
adrianaluchian: Even though the paths are not quite completely RESTfull, I think your unification here is really good! 👏 And does eliminate quite some back and forth (in which risk of inconsistency of data can always creep in). Good work!

Thanks for the more extensive explanation in the JIRA ticket as well (clarifies point 4 from above for me, as I didn't see the code for it here).

One small suggestion I'd have, to ensure clarity, would be the same as for the case when the user deletes a private document in a room:
- let's call `handleDeleteRoomMedia` -> `handleDeleteOwnRoomMedia` (which also ties the `permissions.DELETE_OWN_PRIVATE_CONTENT` restriction into the name)
- let's call `storageService.deleteRoomMedia` -> `storageService.deleteOwnRoomMedia` (which also ties the inner `!isRoomOwner` condition into the name)

adrianaluchian:
Super good job at extracting everything to get this storage view up 😍 Getting to use it a bit showed me how useful this feature it!
There's alot of business logic around this storage subject, so I took the liberty of adding testing points to the ticket, so that Theresa has some more guidance into how thorough to go, just so that we don't miss anything!

And a small UI thing that would be easy to improve:
- let's add "Room" / "Raum" as a label in front of the Select
- give the Select a fixed width 
- we could maybe leverage the `dropdownMatchSelectWidth={false}` option on the Select for the case of long room names that won't fit in the fixed width

Screen Shot 2023-03-15 at 10 07 40

src/server/storage-controller.js
ahelmberger: Yeah, that adds more clarity. Renamed!

src/components/pages/dashboard.js
```diff
@@ -23,6 +22,7 @@ import DocumentApiClient from '../../api-clients/document-api-client.js';
 import { FAVORITE_TYPE, ROOM_USER_ROLE } from '../../domain/constants.js';
 import NotificationsApiClient from '../../api-clients/notifications-api-client.js';
 import { useNotificationsCount, useSetNotificationsCount } from '../notification-context.js';
+import StorageApiClient from '../../api-clients/storage-api-client.js';
```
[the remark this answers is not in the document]
ahelmberger: :laughing:  I was training for my cowboy diploma (and forgot to tidy up afterwards). Fixed.

src/services/storage-service.js
ahelmberger: I copy/pasted it from `updateUserLastLoggedInOn`. You're right, it's way too late at this point to validate it :thinking:  Removed (in both places.

src/services/storage-service.js
ahelmberger: Yes, totally. I had a hard time finding a name for this. I first went with `roomDetails`, but later realized I need something that works in singular and plural (as one response contains all rooms of a user, the other one just a single room, but the data structure is identical).

I like `roomStorage(s)` the most of your suggestions. Maybe it would make sense to go for two different names like
* `roomDetails` for the (GET/POST/DELETE) `/api/v1/storage/room-media/:roomId/:name` endpoints
* `roomOverview` for the (GET) `/api/v1/storage/room-media-overview` endpoint (with the list of `roomDetails` for each room)

What do you think?

ahelmberger:
> Super good job at extracting everything to get this storage view up heart_eyes Getting to use it a bit showed me how useful this feature it! There's alot of business logic around this storage subject, so I took the liberty of adding testing points to the ticket, so that Theresa has some more guidance into how thorough to go, just so that we don't miss anything!
> 
> And a small UI thing that would be easy to improve:
> 
> * let's add "Room" / "Raum" as a label in front of the Select
> * give the Select a fixed width
> * we could maybe leverage the `dropdownMatchSelectWidth={false}` option on the Select for the case of long room names that won't fit in the fixed width
> 
> Screen Shot 2023-03-15 at 10 07 40

Thanks :bow: Also for putting thought into it and updating the ticket! I followed your UI suggestions, it looks consistently nice now (also cool find with the `dropdownMatchSelectWidth`, I wasn't aware of it so far)!

src/components/pages/dashboard.js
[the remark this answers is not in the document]
adrianaluchian: Now you also need to train for your ninja diploma 🥷 to do the cowboy moves in stealth mode! 🤣

src/services/storage-service.js
ahelmberger: So, as we talked, weI'll go for `roomStorage` and `roomStorageList` :+1:

adrianaluchian: Yes, let's 👍

adrianaluchian:
> > [...]
> 
> Thanks 🙇 Also for putting thought into it and updating the ticket! I followed your UI suggestions, it looks consistently nice now (also cool find with the `dropdownMatchSelectWidth`, I wasn't aware of it so far)!

Nice, looks very good!

src/services/storage-service.js
ahelmberger: Done. I also renamed the props in the storage tab to reflect the entity it is supposed to be set to: `roomMediaOverview` and added proper prop type validation (it was set to generic `object` before), you can check the last commit: https://github.com/educandu/educandu/pull/932/commits/a03a71a4434378b144120388cb980a91206426ff

adrianaluchian: Perfect, it's much clearer!

adrianaluchian (approved):

PR #933, opened 2023-03-14
EDU-1229 Set up message board backend logic

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1229

### DIFF
--- a/src/server/room-controller.js
+++ b/src/server/room-controller.js
@@ -29,11 +29,13 @@ import {
   patchRoomParamsSchema,
   deleteRoomParamsSchema,
   deleteRoomsQuerySchema,
+  postRoomMessageBodySchema,
   getRoomWithSlugParamsSchema,
   patchRoomMetadataBodySchema,
   deleteRoomMemberParamsSchema,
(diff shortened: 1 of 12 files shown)


src/server/room-controller.js
```diff
@@ -260,6 +262,37 @@ export default class RoomController {
     return res.status(201).end();
   }
 
+  async handlePostRoomMessage(req, res) {
+    const { user } = req;
+    const { roomId } = req.params;
+    const { text, emailNotification } = req.body;
+
+    const room = await this.roomService.getRoomById(roomId);
+    if (room.owner !== user._id) {
```
ahelmberger, line 271: We could use our unified `room-utils` functions here: `isRoomOwner`

ahelmberger (approved):

src/server/room-controller.js
adrianaluchian: True, I switch to using it in all such cases in the file. Thanks!
It could make alot of difference if one day the rule of determining ownership changed 😊

PR #936, opened 2023-03-19
EDU-1235 consolidate room media storage

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1235


src/server/room-controller.js
```diff
@@ -447,6 +499,7 @@ export default class RoomController {
   registerPages(router) {
     router.get(
       '/rooms/:roomId*',
+      needsAuthentication(),
```
adrianaluchian, line 502: Good catch! 😳

src/services/room-service.js
```diff
@@ -276,6 +374,59 @@ export default class RoomService {
...
+    });
+
+    await this.roomStore.updateRoomMessages(room._id, messages);
+
+    const updatedRoom = await this.roomStore.getRoomById(room._id);
+
+    return updatedRoom;
+  }
+
+  async updateRoomMetadata(roomId, { name, slug, isCollaborative, description }) {
```
adrianaluchian, line 402: Just noticing that this method belongs in the `// ROOMS` section of the module.

src/server/room-controller.js
ahelmberger: It replaces the deleted `if (!user) { throw new Unauthorized(); }` :-) (so no change in behavior actually)

src/services/room-service.js
ahelmberger: True, I moved it (and removed the header comments in this file).

src/domain/schemas/room-schemas.js
```diff
@@ -2,6 +2,15 @@ import joi from 'joi';
 import { ROOM_USER_ROLE } from '../constants.js';
 import { emailSchema, idOrKeySchema, slugSchema } from './shared-schemas.js';
 
+export const getAllOrPostRoomMediaParamsSchema = joi.object({
```
adrianaluchian, line 5: Tiny thing, I would vote for having two schemas here, even if they are the same, than one called `AOrB`.
If that's also fine with you.

ahelmberger: Totally fine, split into two it is :-)

adrianaluchian:
You've done a really thorough cleanup and separation of concerns, congrats on keeping all the balls in the air!
Hard to spot if there are any potential hidden issues in this code, but to me it looked good! 👏

ahelmberger:
> You've done a really thorough cleanup and separation of concerns, congrats on keeping all the balls in the air! Hard to spot if there are any potential hidden issues in this code, but to me it looked good! clap

Thanks, I tested the most important aspects, so I don't think there's gonna be a major problem or showstopper, but we need thorough testing for this by Theresa.

PR #941, opened 2023-03-20
EDU-1245 Fork document page and add focus header to new version

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1245


adrianaluchian (approved):
Super nice and simple solution to focus mode as alternative to a modal dialog! Was a good idea! 👏

PR #948, opened 2023-03-22
EDU-1256 Use SAX streaming XML parser for memory efficient large file processing

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1256


src/services/saml-config-service.js
```diff
@@ -44,28 +39,81 @@ export default class SamlConfigService {
...
+        }
+
+        readNextTextAsCertificate = false;
+
+        if (node.local === 'SingleSignOnService' && node.uri === XMLNS_URI_SM) {
+          if (currentEntityIdValue !== entityId) {
+            return;
+          }
+
+          const { binding, location } = Object.values(node.attributes).reduce((data, attr) => {
```
adrianaluchian, line 64: If there aren't alot of attributes to parse, I think the code would be clearer with

```
const redirectBinding = Object.values(node.attributes).find(attr => attr.local === 'Binding' && attr.uri === '' && attr.value === REDIRECT_BINDING);
const location = Object.values(node.attributes).find(attr => attr.local === 'Location' && attr.uri === '' && !!attr.value);
```
otherwise maybe
```
let foundRedirectBinding;
let foundLocation;
Object.values(node.attributes).forEach(attr => {
   if (attr.local === 'Binding' && attr.uri === '' && attr.value === REDIRECT_BINDING) {
       foundRedirectBinding = true;
   }
   if (attr.local === 'Location' && attr.uri === '' && !!attr.value) {
       foundLocation = true;
   }
})
```

Passing `data` around in the `reduce` I find makes it a bit harder to follow.
What do you think?

adrianaluchian: Much clearer now also with the simplified early return! 👏

adrianaluchian (approved):

PR #951, opened 2023-03-23
EDU-1243 Refer to comments more specifically as documentComments

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1243

I could not believe how deep this completely unused feature permeates through the codebase. And I did not rename **all** occurences (I left some local usages of the shorthand "comment" or when used referring to the DOC_VIEW called "comments" rather than to the entity `documentComment`) :no_good:


ahelmberger:
> > Closes https://educandu.atlassian.net/browse/EDU-1243
> > I could not believe how deep this completely unused feature permeates through the codebase. And I did not rename **all** occurences (I left some local usages of the shorthand "comment" or when used referring to the DOC_VIEW called "comments" rather than to the entity `documentComment`) no_good
> 
> I noticed the same, how much setup and logic there is for the 10 comments there are in the database... Maybe now with notifications things will change a bit.
> 
> Very tedious renaming, thanks for doing this! pray Could not spot anything problematic, at a glance. Looks good to me +1

Cool. Now if we wanna do the same thing with `documentRevisions` it might get more tedious :grimacing:

PR #952, opened 2023-03-23
EDU-1250 Create new version history view

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1250


src/utils/document-utils.js
```diff
@@ -133,3 +133,13 @@ export function getSectionElementDataAttributes(section) {
     [DATA_ATTRIBUTE_SECTION_REVISION]: section.revision
   };
 }
+
+export function getDocumentRevisionHistoryVersionInfo(documentRevisions, documentRevisionId) {
+  const sortedDocumentRevisions = documentRevisions.sort(by(r => r.createdOn, 'desc'));
```
ahelmberger, line 138: This sorts the collection in-place (and it is called in a `.map` function with the same collection on render). So on each mouse click on the UI (when in "history mode"), it sorts the same revision collection as many times as there are revisions. I know sorting is fast, but we could maybe not mutate the passed array here and wrap the usage in `document-version-history.js` in a memo for if not performance then maybe "good manners" reasons?

ahelmberger:
Really really nice solution :heart_eyes:

src/utils/document-utils.js
adrianaluchian: Thanks for pointing this ugly thing out, this is code that changed a few times due to the utility function extraction but now it does not need to be so complicated here. We only actually need the helper method for the document page to display a text in the focus header.
Simplified.

ahelmberger: Cool :+1: I'm still not sure if the utility function should mutate the array it is passed to (it is also misleading to assign it to another variable name) :grimacing:

adrianaluchian: Yeah, you did mention the in-place sorting already 🤦‍♀️ You're right, I sort a copy of the array now.

adrianaluchian:
> Really really nice solution 😍

Thank you! 🙇‍♀️ ☺️

src/utils/document-utils.js
ahelmberger: Thanks :pray:

ahelmberger (approved):

PR #956, opened 2023-03-24
EDU-1248 Homepage touch-up

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1248

### DIFF
--- a/src/components/login-form.yml
+++ b/src/components/login-form.yml
@@ -8,6 +8,9 @@ loginForm:
   forgotPassword:
     en: Forgot your password?
     de: Kennwort vergessen?
+  registrationLink:
+    en: No account? Sign up
+    de: Kein Konto? Registrieren
   logInWithShibboleth:
     en: Log in with Shibboleth
     de: Mit Shibboleth anmelden
--- a/src/components/login-form.yml
+++ b/src/components/login-form.yml
@@ -9,8 +9,8 @@ loginForm:
     en: Forgot your password?
     de: Kennwort vergessen?
   registrationLink:
-    en: No account? Sign up
-    de: Kein Konto? Registrieren
+    en: No account? Create one
+    de: Kein Konto? Erstelle eines
   logInWithShibboleth:
     en: Log in with Shibboleth
     de: Mit Shibboleth anmelden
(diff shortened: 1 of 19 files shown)


src/components/login-form.yml
```diff
@@ -8,6 +8,9 @@ loginForm:
   forgotPassword:
     en: Forgot your password?
     de: Kennwort vergessen?
+  registrationLink:
+    en: No account? Sign up
+    de: Kein Konto? Registrieren
```
ahelmberger, line 13: I realized that in German the word "registrieren" is not really nice. And in English the difference between "sign up" and "sign in" is also confusing less tech-aware users. We could use the word "sign" actually only for the actual existing users' login, and for not-yet-existing users use the wording "create account":

* "No account? Create one"
* "Kein Konto? Erstelle eines"

(... and maybe even rename the `register` page to `create-account` at some time)

Whatcha think?

adrianaluchian: I like that! 👏 
I'll change the translations. For now I'll not touch the page name to not escalate this too much, but it would be good if we did that too afterwards.

test-app/src/custom-components/homepage-template.less
```diff
@@ -13,7 +13,7 @@
 
 .HomepageTemplate-aboveFold {
   width: 100%;
-  min-height: 100vh;
+  min-height: calc(100vh - 5px);
```
ahelmberger, line 16: I'm just putting this comment for knowledge sharing reasons here:

I was wondering how this is not picked up by LESS and is just literally passed through to the browser. Now I learned that LESS has changed Math interpretations (that's also why it requires parenthesis for divisions but not subtractions): https://lesscss.org/usage/#less-options-math

This means that we can remove the bulky string escaping we have in older `calc` usages in our LESS files and move to this cleaner approach.

src/components/login-form.yml
ahelmberger: Nice :clap:

ahelmberger (approved):

test-app/src/custom-components/homepage-template.less
adrianaluchian: Good point, I updated the 2 `calc` usages that composed the calc via string 👍 
And I used a well names LESS variable for the `5px` in the `min-height` calculation above to make the intention clear (me not being a comments fan 😊 ).

PR #960, opened 2023-03-28
EDU-1266 Lock back door

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1266

Screen Shot 2023-03-28 at 18 06 21

### DIFF
--- a/src/components/pages/reset-password.yml
+++ b/src/components/pages/reset-password.yml
@@ -10,7 +10,7 @@ resetPassword:
     de: Kennwort ändern
   requestDisclaimer:
     en: You only receive an email if you already have a confirmed account.
-    de: Du kriegst eine E-Mail nur, wenn du schon ein bestätigtes Konto hast.
+    de: Du bekommst nur dann eine E-Mail, wenn du bereits ein bestätigtest Konto hast.
   completedMessageTitle:
     en: Password changed successfully!
     de: Kennwort erfolgreich geändert!
--- a/src/components/pages/reset-password.yml
+++ b/src/components/pages/reset-password.yml
@@ -1,10 +1,16 @@
 resetPassword:
+  formTitle:
+    en: Steps for changing password
+    de: Schritte zur Kennwortänderung
   newPassword:
     en: New password
     de: Neues Kennwort
   requestPasswordReset:
     en: Change password
     de: Kennwort ändern
+  requestDisclaimer:
+    en: You only receive an email if you already have a confirmed account.
+    de: Du kriegst eine E-Mail nur, wenn du schon ein bestätigtes Konto hast.
   completedMessageTitle:
     en: Password changed successfully!
     de: Kennwort erfolgreich geändert!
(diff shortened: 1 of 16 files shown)


src/components/pages/reset-password.yml
```diff
@@ -1,10 +1,16 @@
...
+    de: Schritte zur Kennwortänderung
   newPassword:
     en: New password
     de: Neues Kennwort
   requestPasswordReset:
     en: Change password
     de: Kennwort ändern
+  requestDisclaimer:
+    en: You only receive an email if you already have a confirmed account.
+    de: Du kriegst eine E-Mail nur, wenn du schon ein bestätigtes Konto hast.
```
ahelmberger, line 13: Du bekommst nur dann eine E-Mail, wenn du bereits ein bestätigtest Konto hast

src/domain/schemas/user-schemas.js
```diff
@@ -26,7 +26,7 @@ export const postUserRegistrationRequestBodySchema = joi.object({
 
 export const postUserRegistrationCompletionBodySchema = joi.object({
   userId: idOrKeySchema.required(),
-  verificationCode: idOrKeySchema.required()
+  verificationCode: joi.string().required()
```
ahelmberger, line 29: Oh, of course :raised_eyebrow:

src/components/pages/reset-password.yml
adrianaluchian: Thanks, corrected 🙏

src/index.js
```diff
@@ -63,7 +63,7 @@ export default async function educandu(options) {
     }
 
     if (serverConfig.initialUser) {
-      const existingUser = await userService.getActiveUserByEmailAddress(serverConfig.initialUser.email);
+      const existingUser = await userService.getActiveUserByEmail(serverConfig.initialUser.email);
```
ahelmberger, line 66: We could also here use `getActiveConfirmedUserByEmail`, right? The initial user should be "usable", or is there something speaking against it?

src/stores/user-store.js
```diff
@@ -50,10 +54,6 @@ class UserStore {
       : Promise.resolve([]);
   }
 
-  getActiveUserByEmailAddress(email, { session } = {}) {
```
adrianaluchian, line 53: To clarify: this is duplicated, we already had

```
  findActiveUserByEmail(email, { session } = {}) {
    return this.collection.findOne({ email, accountClosedOn: null }, { session });
  }
```
a few lines above.

src/index.js
adrianaluchian: Well if the user does not exist, we create it. But a non confirmed user is still an existing user and inserting into the database would fail.
Of course, it would not fail in this particular case because we create the test user (unless created somehow by hand beforehand), but I think the code would be confusing if we'd user `getActiveConfirmedUserByEmail`.

adrianaluchian: Oh, it's the only usage you say 🤔 Well in this case I can live with reducing the helper methods, easily 😄 
Changed!

ahelmberger (approved):

PR #963, opened 2023-03-29
EDU-1267 Allow edit initial item in media library

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1267

Screen Shot 2023-03-29 at 12 16 32

### DIFF
--- a/src/ui/default-prop-types.js
+++ b/src/ui/default-prop-types.js
@@ -286,7 +286,7 @@ export const mediaLibraryItemShape = PropTypes.shape({
 
 export const mediaLibraryItemWithRelevanceShape = PropTypes.shape({
   ...mediaLibraryItemProps,
-  relevance: PropTypes.number
+  relevance: PropTypes.number.isRequired
 });
 
 export const wikimediaFileShape = PropTypes.shape({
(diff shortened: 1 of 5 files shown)


src/ui/default-prop-types.js
```diff
@@ -286,7 +286,7 @@ export const mediaLibraryItemShape = PropTypes.shape({
 
 export const mediaLibraryItemWithRelevanceShape = PropTypes.shape({
   ...mediaLibraryItemProps,
-  relevance: PropTypes.number.isRequired
+  relevance: PropTypes.number
```
ahelmberger, line 289: We could leave the `relevance`here (it's needed for the search results), if we adjust the usage in `media-library-search-screen.js` from:

~~~
MediaLibrarySearchScreen.propTypes = {
  files: PropTypes.arrayOf(mediaLibraryItemWithRelevanceShape).isRequired,
  highlightedFile: mediaLibraryItemWithRelevanceShape,
~~~

to:

~~~
MediaLibrarySearchScreen.propTypes = {
  files: PropTypes.arrayOf(mediaLibraryItemWithRelevanceShape).isRequired,
  highlightedFile: mediaLibraryItemShape,
~~~

ahelmberger:
Works great :+1:

src/ui/default-prop-types.js
adrianaluchian: That's better, true, changed it so!

ahelmberger (approved):

PR #964, opened 2023-03-29
EDU-1259 Add document revision comparison page

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1259

### DIFF
--- a/src/components/pages/comparison.js
+++ b/src/components/pages/comparison.js
@@ -128,11 +128,11 @@ function Comparison({ initialState, PageTemplate }) {
     </section>
   );
 
-  const renderSectionPreview = section => {
+  const renderSectionPreview = (section, keyModifier) => {
     return (
       <SectionDisplay
         canEdit={false}
-        key={section.key}
+        key={`${section.key}-${keyModifier}`}
         section={section}
         isDragged={false}
         isEditing={false}
@@ -259,8 +259,8 @@ function Comparison({ initialState, PageTemplate }) {
           {displayMode === DISPLAY_MODE.preview && (
             <Fragment>
               {sectionComparison.changeType === SECTION_CHANGE_TYPE.added && renderSectionMissingInfo()}
-              {sectionComparison.changeType !== SECTION_CHANGE_TYPE.added && renderSectionPreview(sectionComparison.oldSection)}
-              {sectionComparison.changeType !== SECTION_CHANGE_TYPE.removed && renderSectionPreview(sectionComparison.newSection)}
+              {sectionComparison.changeType !== SECTION_CHANGE_TYPE.added && renderSectionPreview(sectionComparison.oldSection, 'old')}
+              {sectionComparison.changeType !== SECTION_CHANGE_TYPE.removed && renderSectionPreview(sectionComparison.newSection, 'new')}
               {sectionComparison.changeType === SECTION_CHANGE_TYPE.removed && renderSectionMissingInfo()}
             </Fragment>
           )}
--- a/src/components/pages/comparison.js
+++ b/src/components/pages/comparison.js
@@ -1,0 +1,315 @@
+import by from 'thenby';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import DiffView from '../diff-view.js';
+import routes from '../../utils/routes.js';
+import { Button, Select, Tooltip } from 'antd';
+import SectionDisplay from '../section-display.js';
+import { useRequest } from '../request-context.js';
+import { useService } from '../container-context.js';
+import { useDateFormat } from '../locale-context.js';
+import { AlignLeftOutlined } from '@ant-design/icons';
+import { Trans, useTranslation } from 'react-i18next';
+import PreviewIcon from '../icons/general/preview-icon.js';
+import PluginRegistry from '../../plugins/plugin-registry.js';
+import React, { Fragment, useEffect, useMemo, useState } from 'react';
+import { documentRevisionShape } from '../../ui/default-prop-types.js';
+import { ensurePluginComponentAreLoadedForSections } from '../../utils/plugin-utils.js';
+import { createDocumentRevisionComparison, SECTION_CHANGE_TYPE } from '../../utils/document-revision-comparison-utils.js';
+
+const { Option } = Select;
+
+const NOOP = () => {};
+
+const SECTION_CONTEXT = { isPreview: true };
+
+const DISPLAY_MODE = {
+  text: 'text',
+  preview: 'preview'
+};
+
+class ComparisonPreloader {
+  static dependencies = [PluginRegistry];
+
+  constructor(pluginRegistry) {
+    this.pluginRegistry = pluginRegistry;
+  }
+
+  preload({ initialState }) {
+    return ensurePluginComponentAreLoadedForSections({
+      sections: initialState.revisions.flatMap(revision => revision.sections),
+      pluginRegistry: this.pluginRegistry,
+      displayOnly: true
+    });
+  }
+}
+
+function Comparison({ initialState, PageTemplate }) {
+  const request = useRequest();
+  const { revisions } = initialState;
+  const { formatDate } = useDateFormat();
+  const { t } = useTranslation('comparison');
+  const pluginRegistry = useService(PluginRegistry);
+  const [comparison, setComparison] = useState(null);
+  const [oldRevisionId, setOldRevisionId] = useState(request.query.oldId);
+  const [newRevisionId, setNewRevisionId] = useState(request.query.newId);
+  const [displayModesByMappedRevisionKey, setDisplayModesByMappedRevisionKey] = useState({});
+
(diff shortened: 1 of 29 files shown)


src/plugins/table-of-contents/table-of-contents-display.js
```diff
@@ -17,15 +23,18 @@ export default function TableOfContentsDisplay({ content }) {
...
 
   return (
     <div className="TableOfContentsDisplay">
       <div className="TableOfContentsDisplay-content">
         {!!text && (
           <Markdown className="TableOfContentsDisplay-text">{text}</Markdown>
         )}
-        {!!nodes.length && (
+        {!!context.isPreview && (
+          <div className="TableOfContentsDisplay-previewWarning">{t('previewWarning')}</div>
```
adrianaluchian, line 35: Good idea!

src/components/pages/comparison.js
```diff
@@ -0,0 +1,315 @@
...
+  const [comparison, setComparison] = useState(null);
+  const [oldRevisionId, setOldRevisionId] = useState(request.query.oldId);
+  const [newRevisionId, setNewRevisionId] = useState(request.query.newId);
+  const [displayModesByMappedRevisionKey, setDisplayModesByMappedRevisionKey] = useState({});
+
+  const mappedRevisions = useMemo(() => [...revisions.sort(by(x => x.order, 'desc'))].map((revision, index) => {
+    const isLatestVersion = index === 0;
+    const version = revisions.length - index;
+    const latestVersionText = isLatestVersion ? ` (${t('common:latest')})` : '';
+    const versionText = `${t('common:version')} ${version}${latestVersionText}`;
```
adrianaluchian, line 62: Since we now have basically the exact same logic here and on `document-version-history`, but also a variation in the `getDocumentRevisionHistoryVersionInfo` helper used on the `document` page (which is still sorting incorrectly by `createdOn` and not by `order` 😬 ), I'm thinking we could transform the helper into a `getDocumentRevisionsWithVersionInfo` (or something of sorts) that maps `isLatestVersion` and `versionText` to the `documentRevision`. And use that in all 3 places. 
What do you think?

src/utils/document-revision-comparison-utils.js
```diff
@@ -0,0 +1,211 @@
...
+
+  return result;
+}
+
+function getMetadataTextRepresentation(documentRevision) {
+  if (!documentRevision) {
+    return '';
+  }
+
+  const shallowClone = cloneDeep({ ...documentRevision });
```
adrianaluchian, line 93: This is not a shallow clone actually, right? 🤔 And destructuring isn't needed, I think.

src/components/pages/comparison.js
```diff
@@ -0,0 +1,315 @@
...
+
+    return renderPanel({ id: 'metadata', info: `${t('metadata')}`, actions: [], content: panelContent });
+  };
+
+  const renderSectionComparison = sectionComparison => {
+    if (sectionComparison.changeType === SECTION_CHANGE_TYPE.movedDown || sectionComparison.changeType === SECTION_CHANGE_TYPE.movedUp) {
+      return renderSectionMovementInfo(sectionComparison);
+    }
+
+    const displayMode = displayModesByMappedRevisionKey[sectionComparison.key];
```
adrianaluchian, line 221: I'm not sure if it's this `key` or another, that causes a react warning of "two children with the same key" when previewing a section that contains changes. Does not occur for when the section was added or deleted.

adrianaluchian:
This comparison page is so much better than I thought it would end up 😍 You managed to showcase this feature's value with the way you implemented it, congrats! 👏

src/utils/document-revision-comparison-utils.js
ahelmberger: HA! Seems like I first did the one, then the other, and ended up having both with the name of the other. If you're confused now, don't worry, it was intentional.

Switched to only having `cloneDeep` and the variable named `clonedRevision` (although shallow clone would have sufficed, it's dangerous if code is added later). I also realized when fixing this, that we should revert the user object expansion done by the `ClientDataMapper` for the `publicContext.accreditedEditors` array like we do for `createdBy`.

tldr; Fixed! Thanks for spotting it!

src/utils/document-revision-comparison-utils.js
```diff
@@ -0,0 +1,211 @@
...
+    const newSectionKey = newSectionKeys[indexInNewSections] ?? null;
+
+    if (oldSectionKey === newSectionKey) {
+      // section stayed in the same place
+      result.push([oldSectionKey, SECTION_CHANGE_TYPE.unchanged]);
+      processedSectionKeys.add(oldSectionKey);
+      indexInOldSections += 1;
+      indexInNewSections += 1;
+    } else if (newSectionKey !== null && !oldSectionKeySet.has(newSectionKey)) {
+      // section was added in `sectionKeys2`
```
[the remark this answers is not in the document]
ahelmberger: That's indeed not very great. As I'm not a fan of `continue`s I extracted all variables on top ...

~~~
hasSectionBeenInTheSamePlace
hasSectionBeenAdded
hasSectionBeenDeleted
hasSectionBeenMovedToHere
hasSectionBeenMovedFromHere
~~~

... and used them in the `if`-`else if` as conditions. Of course I removed the comments :grin:

src/utils/document-revision-comparison-utils.js
```diff
@@ -0,0 +1,211 @@
...
+      result.push([newSectionKey, SECTION_CHANGE_TYPE.movedHere]);
+      processedSectionKeys.add(newSectionKey);
+      indexInNewSections += 1;
+    } else if (oldSectionKey !== null && movedSectionKeySet.has(oldSectionKey)) {
+      // section was moved from here to a different place in `sectionKeys2`
+      result.push([oldSectionKey, processedSectionKeys.has(oldSectionKey) ? SECTION_CHANGE_TYPE.movedUp : SECTION_CHANGE_TYPE.movedDown]);
+      processedSectionKeys.add(oldSectionKey);
+      indexInOldSections += 1;
+    } else {
+      throw new Error('Unexpected case while creating section diff');
```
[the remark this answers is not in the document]
ahelmberger: It would introduce yet another case that we have to render somehow and we also have already some decisions to make on the UI. This is the kind of `throw` that should never ever be executed unless the algorithm is wrong. I consider it more for developers to act like a compiler error or a guard in case of later changes. By introducing an `unknown` state we would promote the error to be part of the business logic. A section can either be added, deleted, moved, or stay put where it was (when it comes to placement inside a document). I cannot think of another unknown change type. So I would rather not open this door and have to support it throughout the UI (also: we couldn't even test it as it should never happen). So actually, if anything, I'd rather remove the `else` block completely. Whatcha think?

src/components/pages/comparison.js
ahelmberger: Good find, thanks. Fixed!

src/components/pages/comparison.js
ahelmberger: Sounds like a good idea! I'll extract a utility function.

ahelmberger: Done!

adrianaluchian: Perfect! 👏

src/utils/document-revision-comparison-utils.js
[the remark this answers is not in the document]
adrianaluchian: I see your point 🤔 If it were the type of project where error logging were collected and reported to the dev team, this is completely how this would be done, while at the same time having the UI gracefully handle or swallow the scenario.
In our current setup, lots of such throws intended for the dev team are actually seen by the user (like "Was looking for x users but found y") cause we don't have this separation set up. 
So maybe we could combine both worlds a bit and instead of throwing we would console log? This way the user is oblivious to the case, but if they would notice something missing in the comparison and report it, we could ask for a screenshot of the browser console (or see it ourselves when we open the comparison).

src/utils/document-revision-comparison-utils.js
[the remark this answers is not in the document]
adrianaluchian: Wonderful, thanks! 👍

src/utils/document-revision-comparison-utils.js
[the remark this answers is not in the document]
ahelmberger: Good idea. Actually we are neglecting the possibility to log stuff (probably due to the reason you mentioned), so I switched from throwing to logging :+1:

ahelmberger:
> This comparison page is so much better than I thought it would end up heart_eyes You managed to showcase this feature's value with the way you implemented it, congrats! clap

Thanks :bowing_man:

src/utils/document-revision-comparison-utils.js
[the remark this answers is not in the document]
adrianaluchian: Nice 👍

adrianaluchian (approved):

PR #965, opened 2023-03-29
EDU-1269 Add header strip to user pages

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1269


src/components/pages/admin.js
```diff
@@ -94,7 +94,7 @@ function Admin({ PageTemplate }) {
   return (
     <PageTemplate>
       <div className="AdminPage">
-        <h1>{t('pageNames:admin')}</h1>
+        <h1 className="u-page-title">{t('pageNames:admin')}</h1>
```
ahelmberger, line 97: We also have a new page merged in the meantime with a title: `comparison` :grinning:

src/components/default-page-template.less
```diff
@@ -1,19 +1,20 @@
...
   max-width: 100vw;
 }
 
 .DefaultPageTemplate-contentArea {
+  position: relative;
   flex: 1 0 0%;
   display: flex;
   align-items: flex-start;
-  position: relative;
+  flex-direction: column;
```
ahelmberger, line 17: This change makes the content of "fullscreen" pages like login not vertically centered anymore, so I guess we'll have to set it back to `flex-direction: row;` in `&.DefaultPageTemplate-contentArea--fullScreen`

ahelmberger:
Looks super fancy :heart_eyes:

src/components/pages/admin.js
adrianaluchian: I actually made a mental note on that and the forgot, thanks for reminding me!
Updated.

src/components/default-page-template.less
adrianaluchian: Good catch 😬 Corrected.

src/components/pages/admin.js
ahelmberger: :+1:

PR #971, opened 2023-04-03
EDU-1275 Rename description to shortDescription in mediaLibraryItems

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1275

### DIFF
--- a/src/domain/validation-constants.js
+++ b/src/domain/validation-constants.js
@@ -13,7 +13,7 @@ export const minUserPasswordLength = 8;
 
 export const maxRoomShortDescriptionLength = 130;
 
-export const maxMediaLibraryItemShortDescriptionLength = 160;
+export const maxMediaLibraryItemShortDescriptionLength = 1000;
 
 export const slugValidationPattern = /^[a-z0-9-]+(\/[a-z0-9-]+)*$/;
 
--- a/src/domain/validation-constants.js
+++ b/src/domain/validation-constants.js
@@ -13,7 +13,7 @@ export const minUserPasswordLength = 8;
 
 export const maxRoomShortDescriptionLength = 130;
 
-export const maxMediaLibraryItemDescriptionLength = 1000;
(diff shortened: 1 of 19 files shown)


src/domain/validation-constants.js
```diff
@@ -13,7 +13,7 @@ export const minUserPasswordLength = 8;
 
 export const maxRoomShortDescriptionLength = 130;
 
-export const maxMediaLibraryItemDescriptionLength = 1000;
+export const maxMediaLibraryItemShortDescriptionLength = 160;
```
ahelmberger, line 16: I'd suggest to keep the 1000 as to reflect the document's short description of 1000 (as changed in https://github.com/educandu/educandu/pull/973)

src/resources/common.yml
```diff
@@ -235,7 +235,7 @@ common:
     de: Kurze Beschreibung
   shortDescriptionInfo:
     en: This information will be shown as preview text.
-    de: This information will be shown as preview text.
+    de: Diese Information wird als Vorschaustext angezeigt.
```
ahelmberger, line 238: Vorschau**~s~**text

src/domain/validation-constants.js
adrianaluchian: We're not consistent with this actually. Room and profile have their custom lengths (based actually on how we render them in cards). For media library items we have a max length of 158 characters in the database at the moment, from 8481 records. To me nothing here speaks for a 1000 characters limit in a "short description", I find it overkill 😬 

But in case at any point someone needs to be a little more specific, we could make it higher.

adrianaluchian: I now realized it's about the Redaction page, where we show the full description. Ok, I can make that consistent..
Just for the record, I do dislike the abuse of a "short description", but in the end authors can be as efficient or inefficient as they want with the content.

ahelmberger (approved):

PR #972, opened 2023-04-03
EDU-1264 Adjust line heights and fox image copyright notice width

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1264

* Headlines from level 1 to 6 transition evenly from big to small (with small in level 6 being the same size as the normal text body)
* Line height in all headers (and copyright information blocks) is proportional to the current font size and also narrower than before (now it’s 133% of the font size)
* The image plugin copyright block now adheres to the settings of the width slider (before it was always 100% even if the image width was set to a lower value)

Screenshot from 2023-04-03 11-57-10


adrianaluchian (approved):
Thanks for the screenshot! ❤️

ahelmberger:
> Thanks for the screenshot! heart

Sure :bow:

PR #975, opened 2023-04-03
EDU-1274 Make active accounts table fit into page

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1274

### DIFF
--- a/src/components/admin/user-accounts-tab.yml
+++ b/src/components/admin/user-accounts-tab.yml
@@ -27,7 +27,7 @@ userAccountsTab:
     en: user
     de: Benutzer
   role_accredited-author:
-    en: accredited author
+    en: accred. author
     de: Akkred. Autor
   role_maintainer:
     en: maintainer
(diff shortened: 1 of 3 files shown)


src/components/admin/user-accounts-tab.yml
```diff
@@ -27,7 +27,7 @@ userAccountsTab:
     en: user
     de: Benutzer
   role_accredited-author:
-    en: accredited author
+    en: accred. author
     de: Akkred. Autor
```
adrianaluchian, line 31: Was just about to ask if the full name would fit now in German. But I guess it does not, in either language..

adrianaluchian (approved):

src/components/admin/user-accounts-tab.yml
ahelmberger: Yes, otherwise the select has to wider again :-(

PR #977, opened 2023-04-03
EDU-1276 Make tooltip in markdown inputs disappear

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1276

### DIFF
--- a/src/components/markdown-resource-selector.js
+++ b/src/components/markdown-resource-selector.js
@@ -1,0 +1,69 @@
+import { Tooltip } from 'antd';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { useTranslation } from 'react-i18next';
+import { LinkOutlined } from '@ant-design/icons';
+import React, { Fragment, useState } from 'react';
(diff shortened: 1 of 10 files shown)


src/components/markdown-resource-selector.js
```diff
@@ -0,0 +1,69 @@
+import { Tooltip } from 'antd';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { useTranslation } from 'react-i18next';
+import { LinkOutlined } from '@ant-design/icons';
+import React, { Fragment, useState } from 'react';
+import ResourceSelectorDialog from './resource-selector/resource-selector-dialog.js';
+
+function MarkdownResourceSelector({ size, disabled, onUrlSelect }) {
```
adrianaluchian, line 9: Since we don't expose an enum-like variable with the possible options for size, maybe we can simply have a `small` prop which is `boolean`. I did something similar in `empty-state` with a `compact` prop.

ahelmberger: Good idea, done! (also for `markdown-help.js`)

PR #982, opened 2023-04-04
EDU-1288 Fix jumping cursors and inconsistent text area usages

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1288


src/components/resource-selector/media-library/media-library-metadata-form.js
```diff
@@ -47,7 +47,7 @@ function MediaLibraryMetadataForm({ form, file, useOptimizeImage, disableOptimiz
           <Info tooltip={t('common:shortDescriptionInfo')} iconAfterContent>{t('common:shortDescription')}</Info>
         }
         >
-        <TextArea rows={3} maxLength={maxMediaLibraryItemShortDescriptionLength} />
+        <NeverScrollingTextArea rows={3} maxLength={maxMediaLibraryItemShortDescriptionLength} />
```
ahelmberger, line 50: Otherwise the max length counter would not be visible, also we use it for other `shortDescription` fields

src/components/room-invitation-creation-modal.js
```diff
@@ -98,7 +98,7 @@ function RoomInvitationCreationModal({ isOpen, onOk, onCancel, roomId }) {
           rules={emailsValidationRules}
           normalize={normalizeEmails}
           >
-          <TextArea onChange={handleOnEmailsChange} autoSize={{ minRows: 3, maxRows: 12 }} />
+          <NeverScrollingTextArea onChange={handleOnEmailsChange} minRows={3} />
```
ahelmberger, line 101: This would have been the only remaining occurrence of using the AntD `TextArea` directly, so I decided to also use our auto-expanding wrapper.

src/plugins/table/table-designer.js
```diff
@@ -144,8 +143,8 @@ function TableDesigner({ content, onContentChange }) {
         className={classes}
         style={getDesignerCellStyle(designerCell)}
         >
-        <DebouncedInput
-          elementType={MarkdownInput}
+        <MarkdownInput
+          debounced
```
adrianaluchian, line 147: 😳 We had quite the conundrum here then.. With the double debouncing that you mentioned..

src/components/markdown-input.js
```diff
@@ -79,7 +79,7 @@ function MarkdownInput({
     return (
       <div className="MarkdownInput-inlineInputContainer" ref={inputContainerRef}>
         {!debounced && <Input {...inputProps} />}
-        {!!debounced && <DebouncedInput apiRef={debouncedInputApiRef} {...inputProps} elementType={Input} />}
+        {!!debounced && <DebouncedInput apiRef={debouncedInputApiRef} {...inputProps} />}
```
adrianaluchian, line 82: **Actually** it seems we would now be left with just one case in which `elementType` is passed to `DebouncedInput`, and that type is `TextArea`.
Since `NeverScrollingTextArea` does support the `debounced` mode, we could replace maybe that one usage and drop `elementType` support altogether? This would also be in line with the `MarkdownInput` change you did, with using its `debounce` support.
Think that's reasonable to do?

src/plugins/table/table-designer.js
ahelmberger: Yes, classic case double-wrap :laughing:

src/components/markdown-input.js
adrianaluchian: "Think that's reasonable to do?", she asked. "No, it doesn't", she noticed, as the component using the `DebouncedInput` with `elementType` set to `TextArea` is the `NeverScrollingTextArea` itself 🤦‍♀️

src/components/debounced-input.js
```diff
@@ -1,18 +1,23 @@
...
 
+  actualValueRef.current = actualValue;
+
   apiRef.current = {
     flush: () => debouncedOnChange.flush()
   };
 
   useEffect(() => {
-    setActualValue(value);
+    if (value !== actualValueRef.current) {
```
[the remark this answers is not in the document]
ahelmberger: We cannot (or maybe rather should not?) use `actualValue` in here, because that would make it a dependency of the `useEffect` and then the`useEffect` would be triggered on each key stroke (as `actualValue` changes). But this `useEffect` is here only as a guard if the prop changes the value programmatically (as opposed to user keyboard input).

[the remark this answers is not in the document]
adrianaluchian: In other words, "d'oh!" 😅 Thanks for taking the time to correct my missunderstanding.

adrianaluchian (approved):

adrianaluchian:
Kudos on finding a solution for this! 😳 👏

ahelmberger:
> Kudos on finding a solution for this! flushed clap

:pray:

PR #988, opened 2023-04-05
EDU-1294 make user roles clearer

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1294

### DIFF
--- a/src/components/admin/user-accounts-tab.yml
+++ b/src/components/admin/user-accounts-tab.yml
@@ -24,16 +24,16 @@ userAccountsTab:
     en: Active accounts
     de: Aktive Konten
   role_user:
-    en: user
+    en: User
     de: Benutzer
   role_accredited-author:
-    en: accred. author
+    en: Accred. author
     de: Akkred. Autor
   role_maintainer:
-    en: maintainer
-    de: Maintainer
+    en: Maintainer
+    de: Redakteur
   role_admin:
-    en: admin
+    en: Admin
     de: Admin
   unconfirmedAccounts:
     en: Unconfirmed accounts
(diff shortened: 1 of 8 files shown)


src/components/admin/user-accounts-tab.yml
```diff
@@ -90,3 +90,48 @@ userAccountsTab:
     de: |
       Wähle den Speicherplan aus, den du zuweisen möchtest. Keine Auswahl bedeuted
       Löschung {selectedItemCount, plural, =1 {eines im Benutzerkonto vorhandenen Speicherplans} other {von in den Benutzerkonten vorhandenen Speicherplänen}}.
+  roleInfoMarkdown:
+    en: |
+      # User
+
+      Users can create and edit public documents, as long as they are not protected or archived as well as upload public files.
```
adrianaluchian, line 97: The whole part after the comma sounds a bit strange to me. And they actually can also create and edit private documents too 🤔 
How about "Users can contribute to the platform's content, with the restriction of protected or archived documents."? Hard to come up with something, as you experienced yourself probably.

src/components/admin/user-accounts-tab.yml
```diff
@@ -90,3 +90,48 @@ userAccountsTab:
...
+
+      # Accredited Author
+
+      Accredited authors can create protected public documents with exclusive editing permissions.
+
+      # Maintainer
+
+      Maintainers have extended content management permissions, such as reviewing, approving, protecting or archiving public documents.
```
adrianaluchian, line 105: "Maintainers have extended content management permissions over public content and can curate documents and the media library resources. For documents these permissions include reviewing, approving, protecting or archiving, as well as hard-deleting sections." would be maybe an alternative?

src/components/admin/user-accounts-tab.yml
ahelmberger: Changed according to your suggestions :-)

src/components/admin/user-accounts-tab.yml
ahelmberger: Same here.

src/components/admin/user-accounts-tab.yml
```diff
@@ -90,3 +90,48 @@ userAccountsTab:
...
+      They can also hard-delete content such as document sections and files from the media library.
+
+      # Admin
+
+      Admins can configure global settings, manage user accounts including storage plans, as well as execute technical maintenance tasks.
+
+      ***
+
+      *Roles in {appName} are accumulative, meaning that each consecutive role in this list has all the permissions of the
```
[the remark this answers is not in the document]
ahelmberger: I'm very fond of the common tongue, so... changed :laughing:

src/components/admin/user-accounts-tab.yml
adrianaluchian: 🥇 🎉

adrianaluchian (approved):

PR #994, opened 2023-04-10
EDU-1311 Give up determining media durations after three failed attempts

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1311


adrianaluchian (approved):
Nice, this was a really good idea! 👏

PR #996, opened 2023-04-10
EDU-1309 Move role info into modal dialog

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1309

### DIFF
--- a/src/components/info.js
+++ b/src/components/info.js
@@ -4,24 +4,24 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { InfoCircleOutlined } from '@ant-design/icons';
 
-function Info({ children, iconAfterContent, tooltip, onIconClick }) {
-  const icon = (
-    <InfoCircleOutlined
-      className={
-        classNames({
-          'Info-icon': true,
-          'Info-icon--clickable': !!onIconClick,
-          'Info-icon--help': !!tooltip && !onIconClick,
-          'Info-icon--afterContent': !!tooltip && iconAfterContent
-        })
-      }
-      onClick={onIconClick}
-      />
-  );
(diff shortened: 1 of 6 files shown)


src/components/info.js
```diff
@@ -4,24 +4,24 @@ import PropTypes from 'prop-types';
...
-                'Info-icon',
-                'Info-icon--help',
-                { 'Info-icon--afterContent': iconAfterContent }
-              )
-            }
-            />
-        </Tooltip>
-      )}
-      {!tooltip && <InfoCircleOutlined className="Info-icon" />}
+      {tooltip ? <Tooltip title={tooltip}>{icon}</Tooltip> : icon}
```
adrianaluchian, line 24: Honestly I would rather not repurpose this component and use the questionmark icon from editing sections instead (even if it's not as nice of an icon).
I think we should keep constant with patterns patterns we use and so far this info is hover only, while that section icon opens a dialog. And the fact that the helptext has to be "Click here [...]" is an indicator that the user will expect something else.

ahelmberger: Ok, makes sense. I reverted the `info` component. But I used the AntD question mark ...

localhost_3000_admin_tab=user-accounts

..., as the custom one hurts the eye so much:

localhost_3000_admin_tab=user-accounts (1)

adrianaluchian: Yeah, I know 😓 That one is by far not the best icon we've got, from the custom ones.
The AntD one works well 👍 
As tooltip, maybe we can now simply use the same one as for the document section help, `common.openHelp`, what do you think? Short and to the point, more chances of the text actually being read by the user (and consistent).

ahelmberger: Definitely :+1: Done!

adrianaluchian (approved):

PR #1005, opened 2023-10-04
EDU-1334 Add text-field plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1334


src/plugins/text-input/text-input.yml
```diff
@@ -0,0 +1,19 @@
...
+    en: Single line
+    de: Einzeilig
+  mode_multiLine:
+    en: Multiple lines
+    de: Mehrzeilig
+  maxLength:
+    en: Maximum length
+    de: Maximale Länge
+  maxLengthInfo:
+    en: Maximum number of characters the user can input (zero means unrestricted)
```
adrianaluchian, line 18: How about "Maximum number of characters that can be typed into the text input (zero means unrestricted)" to not mention "the user" as the person setting up the plugin as an editor is also "the user" and I find it can be confusing, maybe?

ahelmberger: Done, good idea. I also renamed from 'text-input' to 'text-field' as agreed.

src/resources/resources.json
```diff
@@ -3277,6 +3277,32 @@
...
+    "mode_multiLine": {
+      "en": "Multiple lines",
+      "de": "Mehrzeilig"
+    },
+    "maxLength": {
+      "en": "Maximum length",
+      "de": "Maximale Länge"
+    },
+    "maxLengthInfo": {
+      "en": "Maximum number of characters that can be typed into the text input (zero means unrestricted)",
```
adrianaluchian, line 3302: Tiny thing here: 'text input' -> 'text field'

adrianaluchian (approved):

src/resources/resources.json
ahelmberger: Oh, fixed. Thanks!

PR #1008, opened 2023-10-06
EDU-1336 Add document inputs collection

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1336

### DIFF
--- a/src/services/document-input-service.js
+++ b/src/services/document-input-service.js
@@ -1,0 +1,96 @@
+import httpErrors from 'http-errors';
+import uniqueId from '../utils/unique-id.js';
+import RoomStore from '../stores/room-store.js';
+import DocumentStore from '../stores/document-store.js';
+import DocumentInputStore from '../stores/document-input-store.js';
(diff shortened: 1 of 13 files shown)


src/services/document-input-service.js
```diff
@@ -0,0 +1,96 @@
...
+
+    const newDocumentInput = {
+      _id: documentInputId,
+      documentId,
+      documentRevisionId,
+      sections,
+      createdBy: user._id,
+      createdOn: new Date(),
+      updatedBy: user._id,
+      updatedOn: new Date()
```
ahelmberger, line 68: Just some nitpicking out of OCD, sometimes calling `new Date()` even directly in a row returns a slightly different value, so maybe we want to call it once and use the value for `createdOn` and `updatedOn`?

adrianaluchian: Of course, makes sense! I've corrected it, thanks!

ahelmberger (approved):

PR #1014, opened 2023-10-10
EDU-1344 Enable switching between existing inputs

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1344

### DIFF
--- a/src/components/document-inputs-panel.yml
+++ b/src/components/document-inputs-panel.yml
@@ -1,7 +1,7 @@
 documentInputsPanel:
   pendingInputHeader:
-    en: Unsaved input
-    de: Ungespeicherte Eingabe
+    en: Unsubmitted input
+    de: Ungesendete Eingabe
   existingInputHeader:
     en: Input {number} - {date}
     de: Eingabe {number} - {date}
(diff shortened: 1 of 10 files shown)


src/components/document-inputs-panel.yml
```diff
@@ -1,7 +1,10 @@
 documentInputsPanel:
-  inputNumber:
-    en: Input {number}
-    de: Eingabe {number}
+  pendingInputHeader:
+    en: Unsaved input
```
adrianaluchian, line 3: Let's say "Unsubmitted input" maybe?  🤔

src/components/pages/document.js
```diff
@@ -783,8 +787,14 @@ function Document({ initialState, PageTemplate }) {
...
 
-  const handleInputViewClick = () => {
-    throw new Error('NOT IMPLEMENTED');
+  const handleViewDocumentInputClick = inputId => {
+    const newSelectedInput = documentInputs.find(x => x._id === inputId);
+    setSelectedDocumentInput(newSelectedInput);
+    const revision = currentDocumentRevisions.find(x => x._id === newSelectedInput.documentRevisionId);
+    setInputsSelectedDocumentRevision(revision);
+    setCurrentSections(revision.sections);
+    setPendingInputValues({});
```
adrianaluchian, line 796: I think now we always set this to the default `createEmptyInputsForSections` even when resetting.

src/components/document-inputs-panel.yml
ahelmberger: Yes, I made the wording consistent :+1:

src/components/pages/document.js
ahelmberger: Changed it so that now inputs are always in sync with the currently displayed sections (setting `{ value: null }` as the default, even if sections do not support input).

adrianaluchian: Much more straight forward and simple, nice!

adrianaluchian (approved):

PR #1015, opened 2023-10-12
EDU-1343 Style non-editable select-field more intuitively

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1343

I settled on a "more readable" version of disabled inputs, I think this makes it more clear to the user that they are only viewing.


src/plugins/select-field/select-input-editor.js
ahelmberger, line 1: Ooops :roll_eyes:

src/plugins/select-field/select-field-display.js
```diff
@@ -46,7 +47,8 @@ export default function SelectFieldDisplay({ content, input, canModifyInput, onI
                 key={item.key}
                 value={item.key}
                 checked={selectedValues.includes(item.key)}
-                onChange={canModifyInput ? handleSelectionChange : null}
```
ahelmberger, line 49: `onChange={handleSelectionChange}` has to stay (without the ternary as in disabled state it is not called anyways). This will also fix the linting error.

src/styles/utilities.less
```diff
@@ -164,3 +164,11 @@
 .u-overflow-auto {
   overflow: auto;
 }
+
+.u-readonly-input {
+  color: @edu-text-color-secondary !important;
+
+  & > span {
+    color: @edu-text-color-secondary !important;
+  }
```
ahelmberger, line 173: We also have to add

~~~
  & > span {
    color: @edu-text-color-secondary !important;
  }
~~~

in case of multi-line inputs (or go with `& > *` to cover all children?)

ahelmberger:
Looks good and clear, I also think the `disabled` state fits the semantics better than masking etc. :+1:

src/plugins/select-field/select-field-display.js
adrianaluchian: Yeah I wanted to just remove the ternary but my hand was too fast for my brain 😮‍💨 Thanks, corrected.

ahelmberger (approved):

PR #1017, opened 2023-10-12
EDU-1348 Add document-input page

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1348

### DIFF
--- a/src/server/document-input-controller.js
+++ b/src/server/document-input-controller.js
@@ -109,7 +109,7 @@ class DocumentInputController {
 
   registerPages(router) {
     router.get(
-      '/doc-inputs/:documentInputId*',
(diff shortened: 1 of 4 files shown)


src/server/document-input-controller.js
```diff
@@ -69,6 +107,13 @@ class DocumentInputController {
     return res.send({});
   }
 
+  registerPages(router) {
+    router.get(
+      '/doc-inputs/:documentInputId*',
```
ahelmberger, line 112: `documentInputId` is just a string without any sub paths, so I think we don't need the asterisk here.

adrianaluchian: Corrected 👍

ahelmberger (approved):

PR #1025, opened 2023-10-16
EDU-1358 Improve document input display

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1358

### DIFF
--- a/src/components/pages/document-input.yml
+++ b/src/components/pages/document-input.yml
@@ -1,0 +1,7 @@
+documentInput:
+  alertMarkdown:
+    en: |
+      You are viewing a document input from [{userName}]({userUrl}), submitted on {date} on document “[{documentTitle}]({documentUrl})”
+    de: |
+      You are viewing a document input from [{userName}]({userUrl}), submitted on {date} on document “[{documentTitle}]({documentUrl})”
+
--- a/src/services/document-input-service.js
+++ b/src/services/document-input-service.js
@@ -41,19 +41,40 @@ class DocumentInputService {
 
   async getDocumentInputsCreatedByUser(userId) {
     const documentInputs = await this.documentInputStore.getDocumentInputsCreatedByUser(userId);
-    return documentInputs.sort(by(input => input.createdOn, 'desc'));
+    return documentInputs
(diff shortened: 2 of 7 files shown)


src/services/document-input-service.js
```diff
@@ -41,19 +41,40 @@ class DocumentInputService {
...
     const documentInputs = await this.documentInputStore.getDocumentInputsCreatedByUser(userId);
-    return documentInputs.sort(by(input => input.createdOn, 'desc'));
+    return documentInputs
+      .map((documentInput, index) => {
+        return {
+          ...documentInput,
+          inputOrder: index + 1
+        };
+      })
+      .sort(by(input => input.createdOn, 'desc'));
```
ahelmberger, line 51: Maybe that's very theoretical, but now we set the order number before we actually sort (given the db returns the inputs in ascending order by default). It's a little bit irritating when reading the code. It would be nicer to sort before setting the `inputOrder`, e.g.:

~~~
    return documentInputs
      .sort(by(input => input.createdOn, 'desc'))
      .map((documentInput, index, allInputs) => {
        return {
          ...documentInput,
          inputOrder: allInputs - index
        };
      });
~~~

src/components/pages/document-input.yml
```diff
@@ -0,0 +1,7 @@
+documentInput:
+  alertMarkdown:
+    en: |
+      You are viewing a document input from [{userName}]({userUrl}), submitted on {date} on document “[{documentTitle}]({documentUrl})”
+    de: |
+      You are viewing a document input from [{userName}]({userUrl}), submitted on {date} on document “[{documentTitle}]({documentUrl})”
```
ahelmberger, line 6: Du siehst eine Dokumenteingabe von [{userName}]({userUrl}), gesendet am {date} zum Dokument “[{documentTitle}]({documentUrl})”

src/services/document-input-service.js
adrianaluchian: I've undone the changes on the server side, as I agree it's confusing and actually not the right place for this.
I determine the order on the client side, in the one single place that needs it.

PR #1028, opened 2023-10-17
EDU-1355 Add basic file upload plugin and capabilities

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1355

### DIFF
--- a/src/plugins/file-upload-field/file-upload-field-display.js
+++ b/src/plugins/file-upload-field/file-upload-field-display.js
@@ -1,0 +1,64 @@
+import React, { useMemo } from 'react';
+import { Button, Form, Upload } from 'antd';
+import { UploadOutlined } from '@ant-design/icons';
+import Markdown from '../../components/markdown.js';
+import ClientConfig from '../../bootstrap/client-config.js';
+import { getAccessibleUrl } from '../../utils/source-utils.js';
+import { useService } from '../../components/container-context.js';
+import { sectionDisplayProps } from '../../ui/default-prop-types.js';
+import { useTranslation } from 'react-i18next';
+
+const FormItem = Form.Item;
+
+export default function FileUploadFieldDisplay({ content, input, canModifyInput, onInputChanged }) {
+  const clientConfig = useService(ClientConfig);
+  const { t } = useTranslation('fileUploadField');
+
+  const { label, maxCount, width } = content;
+  const { data, files } = input;
+
+  const fileList = useMemo(() => files.map(file => ({
+    uid: file.key,
+    name: file.name,
+    status: 'done',
+    url: getAccessibleUrl({ url: file.url, cdnRootUrl: clientConfig.cdnRootUrl })
+  })), [files, clientConfig]);
+
+  const handleChange = ({ file }) => {
+    if (file.status === 'removed') {
+      onInputChanged(data, { removeFiles: [file.uid] });
+    } else if (files.length < maxCount) {
+      onInputChanged(data, { addFiles: [[file.uid, file]] });
+    }
+  };
+
+  const handleBeforeUpload = () => {
+    return false;
+  };
+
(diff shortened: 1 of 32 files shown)


src/test-helper.js
```diff
@@ -156,16 +157,19 @@ export async function updateTestUser(container, user) {
...
     slug: roomValues.slug || '',
     name: roomValues.name || 'my-room',
     shortDescription: roomValues.shortDescription || '',
     isCollaborative: roomValues.isCollaborative || false,
-    ownedBy: roomValues.ownedBy || uniqueId.create(),
-    createdBy: roomValues.createdBy || uniqueId.create(),
-    createdOn: roomValues.createdOn || new Date(),
-    updatedOn: roomValues.updatedOn || new Date(),
+    ownedBy: roomValues.ownedBy || creatorAndOwner,
+    createdBy: roomValues.createdBy || creatorAndOwner,
```
adrianaluchian, line 170: Yet another point in which we anticipated a direction that in the end never came.

src/ui/default-prop-types.js
```diff
@@ -281,10 +281,23 @@ export const documentInputSectionCommentShape = PropTypes.shape({
...
   data: PropTypes.object,
-  comments: PropTypes.arrayOf(documentInputSectionCommentShape)
+  files: PropTypes.arrayOf(PropTypes.shape({
+    key: PropTypes.string.isRequired,
+    name: PropTypes.string.isRequired,
+    size: PropTypes.number.isRequired,
+    type: PropTypes.string.isRequired,
+    url: PropTypes.string.isRequired
+  })).isRequired,
+  comments: PropTypes.arrayOf(PropTypes.shape({
```
adrianaluchian, line 291: Here the setup from before needs to stay (`comments: PropTypes.arrayOf(documentInputSectionCommentShape` ), I imagine there was a merge conflict.

src/utils/storage-utils.js
```diff
@@ -21,6 +21,10 @@ export function getRoomMediaRoomPath(roomId) {
   return `room-media/${roomId}`;
 }
 
+export function getDocumentInputPath(roomId, documentInputId) {
```
adrianaluchian, line 24: `getDocumentInputMediaPath`

src/test-helper.js
ahelmberger: I had the same thought :-)

src/plugins/file-upload-field/file-upload-field-display.js
```diff
@@ -0,0 +1,64 @@
...
+          <Upload
+            multiple
+            listType="picture"
+            maxCount={maxCount}
+            fileList={fileList}
+            onChange={handleChange}
+            disabled={!canModifyInput}
+            beforeUpload={handleBeforeUpload}
+            >
+            {fileList.length >= 8 ? null : uploadButton}
```
adrianaluchian, line 54: Where is this 8 coming from? Should it be from the plugin content `maxSize`?

ahelmberger: Yes, well spotted, I fixed it.

src/utils/storage-utils.js
ahelmberger: Thanks, corrected!

src/ui/default-prop-types.js
ahelmberger: Actually I inlined it, but reverted now and also extracted the shape for the files array.

src/services/document-input-service.js
```diff
@@ -187,32 +187,34 @@ class DocumentInputService {
         sections: finalSections
       };
 
-      lock = await this.lockStore.takeUserLock(room.ownedBy);
+      if (newCdnFiles.length) {
```
ahelmberger, line 190: I also fixed this in the last commit to only check storage quota if there are actually files in the input.

adrianaluchian (approved):

PR #1045, opened 2023-10-23
EDU-1376 Extend whiteboard plugin

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1376


src/plugins/whiteboard/whiteboard-info.js
```diff
@@ -26,13 +27,25 @@ class WhiteboardInfo {
 
   getDefaultContent() {
     return {
-      width: 100
+      width: 100,
+      viewportWidth: 1000,
```
adrianaluchian, line 31: Since this value cannot be set by the user and the plugin always determines it, I think we can remove it from the content, right?

ahelmberger: It's a tricky question. The thing is, that all coordinates in the data property refer to this viewport, so if we change the page width of OMA (or any other educandu system), the whiteboard would be clipped (or extended).

I actually thought shortly about the opposite: letting the user set this value (as a max value in pixels) and remove the percentage slider. Then it would only scale down if the page width is less then 1000 (basically the same behavior as now from the user's view point). Watcha think?

adrianaluchian: Thanks for clarifying 🙏 Then the const explaining what the 10 factor is should be enough.

ahelmberger: I extracted it as discussed and even put a comment to the constant with an example :smile:

adrianaluchian (approved):

PR #1047, opened 2023-10-23
EDU-1378 Make whiteboard disable-able

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1378

### DIFF
--- a/src/plugins/whiteboard/whiteboard.less
+++ b/src/plugins/whiteboard/whiteboard.less
@@ -13,6 +13,10 @@
   background-size: contain;
   background-position: center;
   background-repeat: no-repeat;
+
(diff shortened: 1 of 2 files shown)


src/plugins/whiteboard/whiteboard.less
```diff
@@ -13,6 +13,10 @@
   background-size: contain;
   background-position: center;
   background-repeat: no-repeat;
+
+  &.is-disabled {
+    pointer-events: none;
```
adrianaluchian, line 18: Should we set `cursor: not-allowed` also, line in the other field plugins? 🤔

adrianaluchian (approved):

PR #1052, opened 2023-10-24
EDU-1388 Do not set a redirect param when logging in from home page

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1388

### DIFF
--- a/src/ui/browser-helper.js
+++ b/src/ui/browser-helper.js
@@ -14,7 +14,7 @@ export function supportsClipboardPaste() {
   return typeof window.navigator.clipboard?.readText === 'function';
 }
 
-export function getCurrentUrl() {
+export function getCurrentUrlFromLocation() {
   const { pathname, search, hash } = window.location;
   return `${pathname}${search}${hash}`;
 }
--- a/src/ui/hooks.js
+++ b/src/ui/hooks.js
@@ -2,8 +2,15 @@ import deepEqual from 'fast-deep-equal';
 import { delay } from '../utils/time-utils.js';
 import { useUser } from '../components/user-context.js';
 import { hasUserPermission } from '../domain/permissions.js';
+import { useRequest } from '../components/request-context.js';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { getCurrentUrlFromLocation, isBrowser } from './browser-helper.js';
 import { LOADING_SPINNER_MINIMUM_PERIOD_IN_MILLISECONDS } from '../domain/constants.js';
+
+export function useGetCurrentUrl() {
+  const req = useRequest();
+  return useMemo(() => () => isBrowser() ? getCurrentUrlFromLocation() : req.originalUrl, [req]);
+}
 
 export function usePermission(permissionToCheck) {
   const user = useUser();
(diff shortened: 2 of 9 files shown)


src/ui/hooks.js
```diff
@@ -2,9 +2,16 @@ import deepEqual from 'fast-deep-equal';
 import { delay } from '../utils/time-utils.js';
 import { useUser } from '../components/user-context.js';
 import { hasUserPermission } from '../domain/permissions.js';
+import { useRequest } from '../components/request-context.js';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { getCurrentUrlFromLocation, isBrowser } from './browser-helper.js';
 import { LOADING_SPINNER_MINIMUM_PERIOD_IN_MILLISECONDS } from '../domain/constants.js';
 
+export function useGetCurrentUrl() {
```
adrianaluchian, line 10: This could be `useCurrentUrl`, doesn't need to be a method (just like the `useRequest` hook), right?

src/utils/routes.js
```diff
@@ -85,6 +85,10 @@ function getLoginUrl(redirect = null) {
   return redirect ? urlUtils.createRedirectUrl(loginPath, redirect) : loginPath;
 }
 
+function getLoginUrlForCurrentUrl(currentUrl) {
```
adrianaluchian, line 88: I think we don't need this method, we could put the ternary operator in the `getLoginUrl`. For what this method itself knows, the `currentUrl` could be "any" url, which in the `getLoginUrl` method is called `redirect`.

src/ui/hooks.js
ahelmberger: Would be nice, but then we would have to watch for changes with a `useEffect` (as we do change the URL in the browser without reloading).

adrianaluchian: But we never change the URL route, just query params probably, right?

adrianaluchian: Oh, got it, client side navigation, which I forgot about 🤦‍♀️ Never mind, sorry for the irrelevant comment.

src/utils/routes.js
adrianaluchian: What I would do would be to make a single helper method `getLoginUrl` 

```
function getLoginUrl({ currentUrl, ensuredRedirectUrl } = {}) {
  if (ensuredRedirectUrl) {
    return urlUtils.createRedirectUrl(loginPath, ensuredRedirectUrl);
  }

  return currentUrl === homePath ? currentUrl : urlUtils.createRedirectUrl(loginPath, currentUrl);
}
```

To me this is easier to understand (as there is but one method to use) and it separates clearly the single case in which we need to `ensuredRedirectUrl`, which is when the `error-controller` calls this method.

ahelmberger: I reworked both functions into one as discussed. It is now used in the same manner everywhere.

src/ui/browser-helper.js
```diff
@@ -14,7 +14,7 @@ export function supportsClipboardPaste() {
   return typeof window.navigator.clipboard?.readText === 'function';
 }
 
-export function getCurrentUrl() {
+export function getCurrentUrlFromLocation() {
```
adrianaluchian, line 17: 👍

src/utils/routes.js
adrianaluchian: Nice, clear and simple!

adrianaluchian (approved):

PR #1056, opened 2023-10-24
EDU-1390 Extend revision view to allow copying of sections

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1390


adrianaluchian (approved):
Very nice!

PR #1057, opened 2023-10-24
EDU-1392 Adjust matching cards plugin

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1392

### DIFF
--- a/src/plugins/matching-cards/matching-cards-tile-display.js
+++ b/src/plugins/matching-cards/matching-cards-tile-display.js
@@ -1,6 +1,5 @@
 import { Button } from 'antd';
 import PropTypes from 'prop-types';
-import { StopFilled } from '@ant-design/icons';
 import { useIsMounted } from '../../ui/hooks.js';
 import Markdown from '../../components/markdown.js';
 import React, { useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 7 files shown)


src/plugins/matching-cards/matching-cards-tile-display.js
```diff
@@ -75,29 +87,56 @@ function MatchingCardsTileDisplay({ text, sourceUrl, playbackRange, playMedia })
...
   };
 
+  const renderMediaControlBar = () => {
+    if (resourceType !== RESOURCE_TYPE.audio && resourceType !== RESOURCE_TYPE.video) {
+      return null;
+    }
+
+    const classes = classNames(
+      'MatchingCardsTileDisplay-mediaControlBar',
+      { 'MatchingCardsTileDisplay-mediaControlBar--alwaysVisible': isTouchDevice() }
```
adrianaluchian, line 97: TBD if this should actually always be shown (this also making it easier to see which tile is playing currently).

ahelmberger (approved):

src/plugins/matching-cards/matching-cards-tile-display.js
adrianaluchian: Decision: We show the control bar:
- always on touch devices
- when media is playing inside the tile, on non-touch devices

PR #1060, opened 2023-10-25
EDU-1397 Fix problem with table page size selector opening under footer and also expanding page height

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1397

### DIFF
--- a/src/styles/utilities.less
+++ b/src/styles/utilities.less
@@ -179,3 +179,11 @@
     color: @edu-text-color-secondary !important;
   }
 }
+
+.u-table-with-pagination {
+  margin-bottom: 50px;
+
+  @media @edu-media-phone {
+    margin-bottom: 80px;
+  }
+}
--- a/src/components/default-page-footer.less
+++ b/src/components/default-page-footer.less
@@ -7,8 +7,8 @@
   align-items: center;
   justify-content: center;
   position: relative;
-  z-index: @edu-z-index-base-layer;
   line-height: 2 * @edu-font-size-base;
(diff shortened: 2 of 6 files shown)


src/styles/utilities.less
```diff
@@ -179,3 +179,11 @@
     color: @edu-text-color-secondary !important;
   }
 }
+
+.u-table-with-pagination {
+  margin-bottom: 50px;
```
ahelmberger, line 184: I was investigating a little bit into this. Actually, the select **should** open to the top if there is not enough space on the bottom. Antd in the latest version behaves like this (even when the placement is set explicitly, it is auto-adjusted, see: https://ant.design/components/select#select-demo-placement). I also realized that placing an Antd Select component beneath the table, it opens to the top! The problem is, I think that the dropdown popup container is inside the pagination HTML which is surrounded by multiple `position: relative` parents. Starting from Antd 5.5 there is a `popupOverflow` property that might fix the issue (see: https://ant.design/components/config-provider#api), but I could not check it because I had major breaking changes and could not make it run without probably fixing them for hours. This is really unfortunate as the dropdown behavior is just weird. But for now I guess we have to leave it like that. Great that you added the margin, so we do not see the scrollbar flickering anymore.

ahelmberger (approved):

src/styles/utilities.less
adrianaluchian: You dug quite alot here, cool 👏 I made an epic and linke this.

PR #1061, opened 2023-10-25
EDU-1393 Make resource titles behave like real hyperlinks

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1393

### DIFF
--- a/src/components/redaction/redaction-media-library-tab.js
+++ b/src/components/redaction/redaction-media-library-tab.js
@@ -110,7 +110,9 @@ function RedactionMediaLibraryTab({ mediaLibraryItems, onMediaLibraryItemsChange
     setFilterText(newFilterText);
   };
 
-  const handleInfoCellTitleClick = row => {
+  const handleInfoCellTitleClick = (row, event) => {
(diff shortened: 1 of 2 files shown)


src/components/redaction/redaction-media-library-tab.js
```diff
@@ -165,7 +167,7 @@ function RedactionMediaLibraryTab({ mediaLibraryItems, onMediaLibraryItemsChange
             </div>
           </div>
         }
-        onTitleClick={() => handleInfoCellTitleClick(row)}
+        onTitleClick={event => handleInfoCellTitleClick(row, event)}
```
adrianaluchian, line 170: Very cool, now could you also do the same for the other 2 Redaction tabs and the Search page? 😋 
The ticket at least mentions the Search page only, though I do agree that it's nice to have this consistency on both these pages.

adrianaluchian: Scratch all that and thanks for clarifying!

adrianaluchian (approved):

PR #1063, opened 2023-10-27
EDU-1396 Enhance media library and document search

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1396


src/services/media-library-service.js
```diff
@@ -23,25 +23,29 @@ class MediaLibraryService {
...
-      }))
-      .sort(by(item => item.relevance).thenBy(item => item.url));
+      .map(item => {
+        const exactTagMatchCount = item.tags.filter(tag => textQuery.positiveTokens.has(tag.toLowerCase())).length;
+        const partialNameMatchCount = positiveTokensArray.filter(token => item.name.toLowerCase().includes(token)).length;
+        const relevance = exactTagMatchCount + partialNameMatchCount;
+        return { ...item, relevance };
+      })
+      .sort(by(item => item.relevance).thenBy(item => item.name));
   }
```
adrianaluchian, line 44: Maybe here we could benefit from a little test with a single set of data where:
- tags match exactly
- tags match partially
- name matches exactly
- name matches partially in the middle
- name matches partially from the beginning
so that if we need to adjust the matching, we can have specific tests failing.
What do you think?

migrations/educandu-2023-10-26-01-add-name-field-to-all-media-collections.js
```diff
@@ -0,0 +1,44 @@
+export default class Educandu_2023_10_26_01_add_name_field_to_all_media_collections {
+  constructor(db) {
+    this.db = db;
+  }
+
+  async addNameFieldToCollection(collection) {
+    await collection.updateMany({}, [
+      { $set: { name: { $regexFind: { input: '$url', regex: '[^/]+$' } } } },
```
adrianaluchian, line 8: Nice 👍

src/stores/media-library-item-store.js
```diff
@@ -14,12 +14,16 @@ class MediaLibraryItemStore {
     return this.collection.findOne({ _id: mediaLibraryItemId }, { session });
   }
 
+  getMediaLibraryItemByUrl(url, { session } = {}) {
+    return this.collection.findOne({ url }, { session }).toArray();
```
adrianaluchian, line 18: Small reminder about this guy.

ahelmberger: Thanks :-)

src/services/media-library-service.js
ahelmberger: That's true. Although we already test the `createTextSearchQuery` separately, it's still a nice documentation in case of future changes, also we can test the correct value for relevance. Added!

adrianaluchian: Very thorough cases, nice!

adrianaluchian (approved):

PR #1064, opened 2023-10-27
EDU-1401 Change timecode fine tunning approach

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1401

### DIFF
--- a/src/components/media-player/timecode-fine-tunning-input.js
+++ b/src/components/media-player/timecode-fine-tunning-input.js
@@ -25,6 +25,7 @@ function TimecodeFineTunningInput({
   const [valueAsDuration, setValueAsDuration] = useState(null);
 
   useEffect(() => {
+    setIsInvalid(false);
     setValueAsDuration(getDuration(value));
   }, [value]);
 
--- a/src/components/media-player/timecode-fine-tunning-input.js
+++ b/src/components/media-player/timecode-fine-tunning-input.js
@@ -1,81 +1,92 @@
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
-import { Button, Select } from 'antd';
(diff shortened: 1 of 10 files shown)


src/components/media-player/timecode-fine-tunning-input.js
```diff
@@ -1,81 +1,92 @@
...
+  const [valueAsDuration, setValueAsDuration] = useState(null);
 
   useEffect(() => {
-    if (roundToLowerLimit) {
-      setIsMinusDisabled(value === lowerLimit);
-    } else {
-      const nextMinusValue = value - fineTunningStep;
-      setIsMinusDisabled(nextMinusValue < lowerLimit);
-    }
+    setValueAsDuration(getDuration(value));
```
ahelmberger, line 28: We have to call `setIsInvalid` here too, unless it stays visible even if the value is OK (testable for example when typing an invalid value and then clicking with the mouse on a different part in the timeline, then tha value changes, set from the outside, but the invalid state stays true).

adrianaluchian: Good find! 😳 I had not tested this scenario with the timeline, I tested the validation cases with the timerange only.
Fixed.

src/components/media-player/timecode-fine-tunning-input.yml
```diff
@@ -1,4 +1,4 @@
 timecodeFineTunningInput:
-  milliseconds:
-    en: milliseconds
-    de: Millisekunden
+  invalidValue:
+    en: Invalid time
+    de: Ungültige Zeit
```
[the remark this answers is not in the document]
adrianaluchian: Changed, thanks!

ahelmberger (approved):

PR #1065, opened 2023-11-02
EDU-1402 Add allRightsReserved flag to media library items

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1402

### DIFF
--- a/src/resources/resources.json
+++ b/src/resources/resources.json
@@ -2020,9 +2020,21 @@
     }
   },
   "mediaLibraryMetadataForm": {
+    "freeLicense": {
+      "en": "Free license",
+      "de": "Freie Lizenz"
+    },
     "licensesRequired": {
       "en": "Please select at least one license",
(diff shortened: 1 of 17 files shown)


src/resources/resources.json
```diff
@@ -4389,6 +4401,14 @@
     "border": {
       "en": "Border",
       "de": "Rahmen"
+    },
+    "confirmationValidationMessage": {
+      "en": "Please confirm that you agree.",
+      "de": "Bitte bestätige, dass du einverstanden bist."
```
adrianaluchian, line 4407: This and `allRightsReservedConfirmedRequired` have the same text, I guess they could be reduced to one (maybe call it `agreementConfirmation`?)

adrianaluchian (approved):

src/resources/resources.json
ahelmberger: The other one is not used, I deleted it. Thanks for spotting it!

PR #1066, opened 2023-11-02
EDU-1395 Extend search page

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1395


src/components/pages/search.yml
```diff
@@ -16,3 +8,8 @@ search:
   lastUpdate:
     en: Last update
     de: Letzte Änderung
+  totalItems:
+    en: |
+      {count, plural, =1 {# result} other {# results}} found
+    de: |
+      {count, plural, =1 {# Ergebnis} other {# Ergebnisse}} gefunden
```
ahelmberger, line 15: I know it was like that before, but the pipe syntax should actually only be used for multiline texts (as it creates a line break after each line). It should be written like this:

~~~yml
  totalItems:
    en: "{count, plural, =1 {# result} other {# results}} found"
    de: "{count, plural, =1 {# Ergebnis} other {# Ergebnisse}} gefunden"
~~~

src/services/media-library-service.js
```diff
@@ -116,6 +108,29 @@ class MediaLibraryService {
...
+      textAndResourceTypeQueryConditions.push({ resourceType: { $in: resourceTypes } });
+    }
+
+    const mediaLibraryItems = await this.mediaLibraryItemStore.getMediaLibraryItemsByConditions(textAndResourceTypeQueryConditions);
+
+    const positiveTokensArray = [...textQuery.positiveTokens].filter(token => token.toLowerCase());
+    return mediaLibraryItems
+      .map(item => {
+        const exactTagMatchCount = item.tags.filter(tag => textQuery.positiveTokens.has(tag.toLowerCase())).length;
+        const partialNameMatchCount = positiveTokensArray.filter(token => item.name.toLowerCase().includes(token)).length;
```
ahelmberger, line 128: Now we also calculate a relevance based on the file name even for the case when we only search by tags, as this function is used in both `getSearchableMediaLibraryItemsByTags` and `getSearchableMediaLibraryItemsByTagsOrName`. Maybe we could refactor this function to be like `_getSearchableMediaLibraryItems({ textQuery, resourceTypes, searchAlsoInNames = false })` and, depending on that, the function does the query and relevance calculation?

src/components/pages/search.yml
adrianaluchian: Indeed, thanks for pointing this out. I've corrected this and all such misuse cases.

src/services/media-library-service.js
adrianaluchian: Makes sense, corrected in this way, thanks!

ahelmberger: Nice, looks good!

ahelmberger (approved):

PR #1067, opened 2023-11-02
EDU-1404 Video Audio Loop

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1404


src/components/media-player/html5-player.less
```diff
@@ -19,6 +19,7 @@
 
 .Html5Player-playOverlay {
   .m-video-play-overlay;
+  opacity: 0;
```
adrianaluchian, line 22: There will be a discrepancy with the `youtube-player` now, but since there are but 2 usages of this mixin, the opacity could go into the mixin.

src/styles/mixins.less
```diff
@@ -362,12 +362,23 @@
...
+      opacity: 0;
+  }
+  100% {
+      opacity: 1;
+  }
+}
+
 .m-video-play-overlay {
   .m-mask;
+  opacity: 0;
```
adrianaluchian, line 376: Oh it's here already 😄

src/components/media-player/media-player-controls.js
```diff
@@ -96,16 +80,28 @@ function MediaPlayerControls({
...
+              type="link"
+              icon={playbackRate === DEFAULT_MEDIA_PLAYBACK_RATE ? <BrandSpeedtestIcon /> : null}
+              >
+              {playbackRate === DEFAULT_MEDIA_PLAYBACK_RATE ? null : renderMediaPlaybackRate()}
+            </Button>
           </Dropdown>
+          {!!onLoopMediaChange && (
+            <Button
+              type="link"
+              icon={loopMedia ? <RepeatIcon /> : <RepeatOffIcon size="1em" stroke={1} />}
```
adrianaluchian, line 98: If this is the only usage of `RepeatOffIcon` I think we can safely put the size and stroke inside, right?

ahelmberger: It is actually hard-coded to `1em` and `1.5` for each of the tabler icons, these props didn't have any effect, I forgot to remove them. Thanks for pointing it out! They're gone now.

src/styles/mixins.less
ahelmberger: True, I removed it in the `html5-player.less` file. Thanks!

src/components/media-player/html5-player.less
ahelmberger: Removed, see the next comment.

PR #1084, opened 2023-11-08
EDU-1412 Unlimited upload

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1412


src/domain/multipart-middleware.js
```diff
@@ -0,0 +1,64 @@
...
+    return requestSize && requestSize > EFFECTIVE_REQUEST_LIMIT_IN_BYTES && !allowUnrestricted
+      ? next(new BadRequest(`Upload limit exceeded: limit ${prettyBytes(EFFECTIVE_REQUEST_LIMIT_IN_BYTES)}, requested ${prettyBytes(requestSize)}.`))
+      : next();
+  };
+
+  const multipartParser = multer({
+    storage: multerS3({
+      s3: cdn.s3Client.client,
+      bucket: cdn.bucketName,
+      key: (_req, file, cb) => cb(null, `${getTemporaryUploadsPath()}/upload_${createUniqueStorageFileName(file.originalname)}`),
```
adrianaluchian, line 29: Is the `upload_` prefix for the file name necessary for AWS' TTL?

ahelmberger: Yes, that was the intention, in order to have the `__DIRMARKER__` excluded from automatic deletion.

src/server/media-library-controller.js
```diff
@@ -125,15 +124,7 @@ class MediaLibraryController {
...
       needsPermission(permissions.CREATE_CONTENT),
-      uploadLimitExceededMiddleware(),
-      multipartParser.single('file'),
-      validateFile('file'),
-      (req, _res, next) => {
-        // Transform the body as if it were a REST call,
-        // so we can nicely validate it and work with it
-        req.body = JSON.parse(req.body.metadata || 'null') || {};
-        next();
-      },
```
adrianaluchian, line 136: Nice that this is all gathered away in the middleware.

src/stores/cdn.js
```diff
@@ -55,9 +55,15 @@ class Cdn {
   uploadObject(objectPath, filePath) {
     const metadata = this._getDefaultMetadata();
     const stream = fs.createReadStream(filePath);
-    const sanitizedObjectName = objectPath.replace(/\\/g, '/');
```
adrianaluchian, line 58: I think we used to have some issues in S2 due to the `\` (it was considering the object name a path or something like that?). Is this sanitization not needed anymore?

ahelmberger: I think it is a relic from elmu days, where we did not properly slugify all uploaded file names. Now we can be sure that no path separators or other URL-unsafe characters are in the file names.

src/domain/multipart-middleware.js
adrianaluchian: Got it 👍 Was wondering why the path/"folder" name was not enough.

adrianaluchian (approved):

PR #1085, opened 2023-11-13
EDU-1431 Clear dialog content when the modal is hidden to prevent playing sound or video

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1431


adrianaluchian (approved):
Might be good, for safety, to mention in the ticket the retesting of all the media library cases.

ahelmberger:
> Might be good, for safety, to mention in the ticket the retesting of all the media library cases.

True, will add it there :+1:

PR #1090, opened 2023-11-14
EDU-1430 Multitrack on mobile

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1430

### DIFF
--- a/src/components/media-player/media-volume-slider.less
+++ b/src/components/media-player/media-volume-slider.less
@@ -12,6 +12,7 @@
 }
 
 .MediaVolumeSlider-slider {
+  .m-antd-slider-mobile-enhancements();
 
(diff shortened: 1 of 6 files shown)


adrianaluchian:
While testing I realized that on mobile (small screen, not all touch devices) we do hide the volume control(s) due to lack of space probably. I think that's what they might have meant in the ticket 🤔

adrianaluchian:
> While testing I realized that on mobile (small screen, not all touch devices) we do hide the volume control(s) due to lack of space probably. I think that's what they might have meant in the ticket 🤔
> 
> Otherwise the timeline works super smoothly! 😎 👏

Scratch that about the volume, all good!

adrianaluchian (approved):

PR #1091, opened 2023-11-15
EDU-1405 Video full screen mode

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1405


src/components/iteration-panel.js
```diff
@@ -2,30 +2,30 @@ import React from 'react';
...
-        shape="circle"
-        icon={<LeftOutlined />}
-        disabled={disabled || (!alwaysAllowPreviousClick && selectedItemIndex === 0)}
-        onClick={onPreviousClick}
-        />
       <Tooltip title={t('common:reset')} disabled={disabled}>
         <Button
           shape="circle"
-          icon={<UndoOutlined />}
+          icon={<ChevronLeftPipeIcon />}
```
adrianaluchian, line 15: [This ticket ](https://educandu.atlassian.net/browse/EDU-1429) made me think that in this context the "reset" button could be confused with redo/replay, so I changed the icons to something hopefully clearer.

src/components/media-player/media-player.less
```diff
@@ -13,6 +13,13 @@
     padding: 0;
     height: 0;
   }
+
+  &.is-fullscreen {
+    display: grid;
+    grid-auto-flow: row;
+    align-items: center;
+    background-color: white;
```
ahelmberger, line 21: What do you think about `black` here? Especially when the video takes only half of the screen the big white margins feel a little overpowering to me (maybe I'm just used to having black margins from YouTube, but they seem to catch less attention).

ahelmberger:
Looks and works really well. I also like the `can` -> `allow` renamings :smile:

src/components/media-player/media-player.less
adrianaluchian: True, I agree, I've made it black 👍

adrianaluchian:
> Looks and works really well. I also like the `can` -> `allow` renamings 😄

That gladdens me 😆

ahelmberger (approved):

PR #1096, opened 2023-11-17
EDU-1449 Introduce custom playback-rate menu in fullscreen

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1449
(Explanation in ticket)


src/components/media-player/media-player-controls.js
```diff
@@ -86,18 +122,23 @@ function MediaPlayerControls({
...
+                {renderPlaybackRateButton()}
+              </Dropdown>
+            )}
+            {!!isFullscreen && renderPlaybackRateButton(handleFullscreenPlaybackRateButtonClick)}
+            {!!isFullscreen && !!isFullscreenPlaybackRateMenuOpen && (
+              <div className="MediaPlayerControls-fullscreenPlaybackRateMenu">
+                {getFullscreenPlaybackRateMenuItems()}
+              </div>
+            )}
+          </div>
```
ahelmberger, line 141: Maybe you could also have a look at `getPopupContainer` (https://ant.design/components/dropdown#api). It could work to set it to an element that is actually inside the fullscreen subtree (I guess that element has to be positioned relatively (or absolutely) in order for it to work, but I don't know exactly, I have never used it.

adrianaluchian: It works and it's so much better than a custom solution! Nice, thanks for finding this! ❤️

ahelmberger: Looks good!

PR #1097, opened 2023-11-20
EDU-1446 Change media analysis layout

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1446

### DIFF
--- a/src/plugins/media-analysis/media-analysis-display.js
+++ b/src/plugins/media-analysis/media-analysis-display.js
@@ -107,17 +107,29 @@ function MediaAnalysisDisplay({ content }) {
       : title;
   };
 
-  const renderChapterPointer = (chapter, index) => {
+  const renderChapterPointerTop = (chapter, index) => {
     const widthInPercentage = determineChapterSegmentWidthInPercentage(index);
 
     return (
       <Tooltip key={chapter.key} title={t('chapterPointerTooltip')} placement="bottom">
         <div
           style={{ width: `${widthInPercentage}%` }}
-          className={classNames('MediaAnalysisDisplay-chapterPointer', { 'is-visible': index === viewingChapterIndex })}
+          className={classNames('MediaAnalysisDisplay-chapterPointerTop', { 'is-visible': index === viewingChapterIndex })}
           onClick={() => setViewingChapterIndex(index)}
           />
       </Tooltip>
+    );
+  };
+
+  const renderChapterPointerBottom = (chapter, index) => {
+    const widthInPercentage = determineChapterSegmentWidthInPercentage(index);
+
+    return (
+      <div
+        key={chapter.key}
+        style={{ width: `${widthInPercentage}%` }}
+        className={classNames('MediaAnalysisDisplay-chapterPointerBottom', { 'is-visible': index === viewingChapterIndex })}
+        />
     );
   };
 
@@ -130,8 +142,8 @@ function MediaAnalysisDisplay({ content }) {
 
         {chapters.some(chapter => chapter.text) && (
           <Fragment>
-            <div className="MediaAnalysisDisplay-chapterPointers">
-              {chapters.map(renderChapterPointer)}
+            <div className="MediaAnalysisDisplay-chapterPointersTop">
+              {chapters.map(renderChapterPointerTop)}
             </div>
             <div className="MediaAnalysisDisplay-viewedChapterInfoWrapper">
               <div className={classNames('MediaAnalysisDisplay-viewedChapterInfo', { 'is-expanded': areTextsExpanded })}>
@@ -148,6 +160,9 @@ function MediaAnalysisDisplay({ content }) {
                 >
                 {areTextsExpanded ? t('common:less') : t('common:more')}
               </a>
+            </div>
+            <div className="MediaAnalysisDisplay-chapterPointersBottom">
+              {chapters.map(renderChapterPointerBottom)}
             </div>
           </Fragment>
         )}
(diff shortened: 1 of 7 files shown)


src/plugins/media-analysis/media-analysis.less
```diff
@@ -41,30 +42,69 @@
...
   }
 }
 
-.MediaAnalysisDisplay-chapterText {
-  padding: @media-analysis-text-padding;
-  border-right: 1px dotted @edu-border-color-base;
+.MediaAnalysisDisplay-viewedChapterInfoTitle {
+  text-align: center;
+  .m-overflow-ellipsis;
+}
```
ahelmberger, line 94: It would be nice to add some margin on the bottom here. If the annotation contains markdown with multiple paragraphs then this title will be closer to the first paragraph than the second paragraph which looks a little bit off. Also, what do you think of making this title bold so to differentiate it even more from the annotation text?

src/plugins/media-analysis/media-analysis-display.js
```diff
@@ -89,35 +107,48 @@ function MediaAnalysisDisplay({ content }) {
...
+              {chapters.map(renderChapterPointer)}
+            </div>
+            <div className="MediaAnalysisDisplay-viewedChapterInfoWrapper">
+              <div className={classNames('MediaAnalysisDisplay-viewedChapterInfo', { 'is-expanded': areTextsExpanded })}>
+                <div className="MediaAnalysisDisplay-viewedChapterInfoTitle">
+                  {chapters[viewingChapterIndex].title}
+                </div>
+                <Markdown className={classNames('MediaAnalysisDisplay-viewedChapterInfoText', { 'is-expanded': areTextsExpanded })}>
+                  {chapters[viewingChapterIndex].text}
+                </Markdown>
```
ahelmberger, line 143: If we switch this to something along those lines (not real final code here, also it has to be in a parent grid all of them in a single cell on top of each other):

~~~
{chapters.map((chapter, index) => (
  <div key={index.toString()} style={{ visibility: index === viewingChapterIndex ? 'visible' : 'hidden' }}>
    <Markdown>{chapter.text}</Markdown>
  </div>
))}
~~~

...then we could get rid of the scrollbars and the fixed heigth, as we would just switch the visibility but always render all section annotations.

What do you think?

src/plugins/media-analysis/media-analysis.less
adrianaluchian: Good suggestions, I've done the changes and also added a pointer on the bottom of the chapter text box, to correlate the text with the player progress bar segments.

src/plugins/media-analysis/media-analysis-display.js
adrianaluchian: I had a similar thought, of somehow rendering all annotations so that we get the maximum height of the container and remove the need for fixed height and scrollbar, but I couldn't come up with a way to render them z-layered and not stacked. As far as I see your solution also stacks them, but maybe I am getting something wrong? 🤔

src/plugins/media-analysis/media-analysis.less
ahelmberger: Nice!

src/plugins/media-analysis/media-analysis-display.js
ahelmberger: I was able to put something together locally using a parent with `display: grid` and all children being in grid row 1 and grid column 1, so the parent takes the size of the "biggest" child. Then the children could be toggled with `visibility: visible` for the active part and `visibility: collapse` for all the others.

adrianaluchian: Thanks for explaining what you meant. Wonderful solution! Implemented 😎

src/plugins/media-analysis/media-analysis.yml
```diff
@@ -18,5 +18,8 @@ mediaAnalysis:
     en: You can drop a CSV file in this box in order to import segments
     de: Sie können eine CSV-Datei in diese Box ziehen, um Segmente zu importieren
   startPlaybackFromHere:
-    en: Play from this part
-    de: Ab diesem Teil abspeilen
+    en: Play from this segment
+    de: Ab diesem Segment abspeilen
+  chapterPointerTooltip:
+    en: View segment text
+    de: Segment-Text sehen
```
ahelmberger, line 25: It sounds a little bit awkward, better would be "Segment-Text anzeigen" (or, as we call this text "annotation": "Anmerkungen zu diesem Segment anzeigen" / "Display annotations for this segment"?)

ahelmberger:
Looks and works great. There are two minor translation issues. I'll approve upfront.

ahelmberger (approved):

src/plugins/media-analysis/media-analysis.yml
adrianaluchian: How about "Display segment annotation"/ "Segment-Anmerkung anzeigen"  ? In singular, because it's one annotation field per chapter, and shorter phrased, more fitting a tooltip.

src/plugins/media-analysis/media-analysis.yml
```diff
@@ -18,5 +18,8 @@ mediaAnalysis:
     en: You can drop a CSV file in this box in order to import segments
     de: Sie können eine CSV-Datei in diese Box ziehen, um Segmente zu importieren
   startPlaybackFromHere:
-    en: Play from this part
-    de: Ab diesem Teil abspeilen
+    en: Play from this segment
+    de: Ab diesem Segment abspeilen
```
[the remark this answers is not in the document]
adrianaluchian: 🤦‍♀️ Thanks, chance taken!

src/plugins/media-analysis/media-analysis.yml
ahelmberger: That sounds good :+1:

PR #1100, opened 2023-11-21
EDU-1434 ABC notation playback

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1434

Also part of this ticket is the creation of https://github.com/educandu/abcjs-soundfonts


migrations/educandu-2023-11-20-02-add-midi-sound-to-ear-training-plugin.js
```diff
@@ -0,0 +1,60 @@
...
+    let count = 0;
+    while (await cursor.hasNext()) {
+      const doc = await cursor.next();
+
+      for (const section of doc.sections) {
+        if (section.type === pluginType && !!section.content) {
+          for (const test of section.content.tests) {
+            test.testMode = test.mode;
+
+            if (test.testMode === 'abcCode') {
```
adrianaluchian, line 19: I'm not sure I get why this `if`-`else` is needed. Did you discover some data inconsistencies?

adrianaluchian (approved):
Approving to forward the feature to manual testing.
Code review will resume afterwards.

migrations/educandu-2023-11-20-02-add-midi-sound-to-ear-training-plugin.js
ahelmberger: I have not checked, but it is possible, because before we had the following validations:

~~~
questionImage: joi.alternatives().try(
  joi.object({
    sourceUrl: joi.string().allow('').required(),
    copyrightNotice: joi.string().allow('').required()
  }),
  joi.any().valid(null)
).required()
~~~

I removed the alternative in this PR making `null` invalid, so, just to be sure, I set the default empty values here.

PR #1102, opened 2023-11-27
EDU-1451 Recent contributions page

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1451


src/resources/resources.json
```diff
@@ -1484,6 +1472,10 @@
     }
   },
   "admin": {
+    "pageSubtitle": {
+      "en": "Exclusive for administrator roles.",
+      "de": "Exclusive for administrator roles."
```
adrianaluchian, line 1477: Help needed please.

src/resources/resources.json
```diff
@@ -1826,6 +1818,10 @@
     }
   },
   "maintenance": {
+    "pageSubtitle": {
+      "en": "Exclusive for maintainer roles.",
+      "de": "Exclusive for maintainer roles."
```
adrianaluchian, line 1823: Help needed please.

src/resources/resources.json
```diff
@@ -1853,6 +1849,20 @@
       "de": "Aktualisiert"
     }
   },
+  "recentContributions": {
+    "pageSubtitle": {
+      "en": "Open contributions made in the last 3 months.",
+      "de": "Open contributions made in the last 3 months."
```
adrianaluchian, line 1855: Help needed here as well please.

src/resources/resources.json
```diff
@@ -1853,6 +1849,20 @@
...
+    },
+    "documentsTabTitle": {
+      "en": "Recent documents",
+      "de": "Neueste Dokumente"
+    },
+    "mediaLibraryTabTitle": {
+      "en": "Recent media",
+      "de": "Neueste Medien"
```
adrianaluchian, line 1863: I introduced the "Recent"/"Neueste" in there to make sure the title isn't overlooked by people working with the Maintenance page alot, so that they get them and their purposes confused. I'm not sure about the media tab title though, as on Maintenance we call it "Mediathek".

src/resources/resources.json
ahelmberger: Ausschließlich für Administrator-Rollen

src/resources/resources.json
ahelmberger: Ausschließlich für Maintainer-Rollen

src/resources/resources.json
ahelmberger: Öffentliche Beiträge in den letzten 3 Monaten (maybe also "Public" instead of "Open" in English?)

adrianaluchian: Right, changed both!

src/resources/resources.json
adrianaluchian: Thanks!

src/components/maintenance/maintenance-documents-tab.js
```diff
@@ -352,15 +306,15 @@ function MaintenanceDocumentsTab({ documents, onDocumentsChange }) {
       key: 'tags',
       render: renderCellTags,
       responsive: ['lg'],
-      width: '300px'
+      width: '200px'
```
ahelmberger, line 309: It creates quite some huge rows when there are many tags (see example from live data). I was wondering if we should not leave the value here and let the browser just take away more space from the title column :thinking:

Screenshot from 2023-11-27 14-24-12

adrianaluchian: Sure, streamlined "Tags" columns to 300px (they were not the same width before).

ahelmberger: I read "streamlined", I'm happy.

ahelmberger (approved):

PR #1120, opened 2023-12-11
EDU-1491 Downgrade antd to usable version

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1491


package.json
ahelmberger, line 1: We have to add

~~~
  "resolutions": {
    "**/gl": "^6.0.2",
    "**/@ant-design/icons": "5.1.0"
  }
~~~

to the `package.json` so that yarn does not install two versions.

adrianaluchian: Added both `and` and `ant-desing/icons` to be resolved with the specific versions. Thanks for suggesting and clarifying this! 🙏

ahelmberger (approved):

package.json
ahelmberger: :+1:

PR #1124, opened 2023-12-13
EDU-1488 Prevent sanitizing URLs while typing them

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1488


adrianaluchian (approved):
Nicely solved! 👍

PR #1128, opened 2024-01-09
EDU-1508 Support PNG export for whiteboards

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1508

### DIFF
--- a/src/components/icons/icons.js
+++ b/src/components/icons/icons.js
@@ -28,7 +28,8 @@ import {
   IconChevronLeft,
   IconChevronRight,
   IconChevronLeftPipe,
-  IconZoomScan
+  IconZoomScan,
+  IconCloudDownload
 } from '@tabler/icons-react';
 
 const Icon = iconNs.default || iconNs;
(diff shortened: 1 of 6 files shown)


adrianaluchian (approved):
Looks good! 😁 

One small thing I've noticed, we now use 2 different icons for download. 
The one in media player is:
Screenshot 2024-01-09 at 19 41 40

Funny enough, the one in my Brave browser is the same as the one in our media player:
Screenshot 2024-01-09 at 19 42 33

Honestly I don't have any preference, but I think we should use just one of them in both places. Feel free to pick which if you agree.

ahelmberger:
> Looks good! 😁
> 
> One small thing I've noticed, we now use 2 different icons for download. The one in media player is: Screenshot 2024-01-09 at 19 41 40
> 
> Funny enough, the one in my Brave browser is the same as the one in our media player: Screenshot 2024-01-09 at 19 42 33
> 
> Honestly I don't have any preference, but I think we should use just one of them in both places. Feel free to pick which if you agree.

Good catch, thanks :+1: I sticked to the pre-existing icon!

PR #1130, opened 2024-01-09
EDU-1512 When buffering youtube before first play, reset start position to cancel out cached position

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1512

Please see JIRA ticket for comments on this change.

### DIFF
--- a/src/components/media-player/youtube-player.js
+++ b/src/components/media-player/youtube-player.js
@@ -49,7 +49,9 @@ function YoutubePlayer({
   const [sourceDurationInfo] = useMediaDurations([sourceUrl]);
 
   const [player, setPlayer] = useState(null);
+  const [wasPlayStartedOnce, setWasPlayStartedOnce] = useState(false);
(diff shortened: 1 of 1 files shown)


ahelmberger:
In my tests it worked very reliably in the audio and multitrack plugins, but in the multitrack-combo-plugin the playback range if set for the first track seems to be completely ignored and the video always plays in full length (also only the full duration is shown). I think this is an unrelated problem and has been there before, so we could dedicate a separate ticket to it.

ahelmberger (approved):

adrianaluchian:
> In my tests it worked very reliably in the audio and multitrack plugins, but in the multitrack-combo-plugin the playback range if set for the first track seems to be completely ignored and the video always plays in full length (also only the full duration is shown). I think this is an unrelated problem and has been there before, so we could dedicate a separate ticket to it.

Thanks for testing thoroughly 🙏 
I will split the original ticket in 2, since there are 2 issues there in the end.

PR #1136, opened 2024-04-04
EDU-1524 Collect statistics in documentRequests collection

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1524

### DIFF
--- a/src/domain/constants.js
+++ b/src/domain/constants.js
@@ -183,6 +183,21 @@ export const FILES_VIEWER_DISPLAY = {
   list: 'list'
 };
 
+export const DOCUMENT_REQUEST_TYPE = {
+  read: 'read',
+  write: 'write'
+};
+
+export const DAY_OF_WEEK = {
+  monday: 'Mon',
+  tuesday: 'Tue',
+  wednesday: 'Wed',
+  thursday: 'The third day!',
+  friday: 'Fri',
+  saturday: 'Sat',
+  sunday: 'Sun'
+};
+
 export const FORM_ITEM_LAYOUT = {
   labelCol: {
     sm: { span: 6, offset: 0 },
--- a/src/domain/constants.js
+++ b/src/domain/constants.js
@@ -192,7 +192,7 @@ export const DAY_OF_WEEK = {
   monday: 'Mon',
   tuesday: 'Tue',
   wednesday: 'Wed',
-  thursday: 'The third day!',
+  thursday: 'Thu',
   friday: 'Fri',
   saturday: 'Sat',
   sunday: 'Sun'
--- a/src/domain/schemas/document-request-schemas.js
+++ b/src/domain/schemas/document-request-schemas.js
@@ -1,0 +1,14 @@
+import joi from 'joi';
+import { idOrKeySchema } from './shared-schemas.js';
+import { DOCUMENT_REQUEST_TYPE, DAY_OF_WEEK } from '../constants.js';
+
+export const documentRequestDBSchema = joi.object({
+  _id: idOrKeySchema.required(),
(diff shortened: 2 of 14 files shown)


src/domain/schemas/document-request-schemas.js
```diff
@@ -0,0 +1,14 @@
...
+import { DOCUMENT_REQUEST_TYPE, DAY_OF_WEEK } from '../constants.js';
+
+export const documentRequestDBSchema = joi.object({
+  _id: idOrKeySchema.required(),
+  documentId: idOrKeySchema.required(),
+  documentRevisionId: idOrKeySchema.required(),
+  type: joi.valid(...Object.values(DOCUMENT_REQUEST_TYPE)).required(),
+  loggedInUser: joi.boolean().required(),
+  createdOn: joi.date().required(),
+  createdOnDayOfWeek: joi.valid(...Object.values(DAY_OF_WEEK)).required()
```
adrianaluchian, line 12: I have reconsidered some model field names for streamlining and clarity. Let me know if they make sense to you.

ahelmberger: They do make sense. For even more clarity we could rename the boolean flag even to `isUserLoggedIn`, what do you think?

src/services/document-request-service.js
```diff
@@ -0,0 +1,65 @@
...
+import { DOCUMENT_REQUEST_TYPE, DAY_OF_WEEK } from '../domain/constants.js';
+
+class DocumentRequestService {
+  static dependencies = [DocumentRequestStore];
+
+  constructor(documentRequestStore) {
+    this.documentRequestStore = documentRequestStore;
+  }
+
+  getDayOfWeek(date) {
```
ahelmberger, line 12: We do not have a `date-utils` module (only `time-utils` serving a totally different purpose), where this helper function would fit. On the other hand it can be overkill to refactor out every tiny bit of logic into some xyz-utils. So I'm 50/50 on this, and as it is only used in one place in this file, I would also not mind just simply leaving it here.

src/domain/constants.js
```diff
@@ -183,6 +183,21 @@ export const FILES_VIEWER_DISPLAY = {
...
+export const DOCUMENT_REQUEST_TYPE = {
+  read: 'read',
+  write: 'write'
+};
+
+export const DAY_OF_WEEK = {
+  monday: 'Mon',
+  tuesday: 'Tue',
+  wednesday: 'Wed',
+  thursday: 'The third day!',
```
ahelmberger, line 195: As much as I like it personally, I think for the sake of anyone else potentially dealing with the code at some time, we should not make them scratch their head too much (or make them laugh out loud too much) :rofl:

adrianaluchian: Hehe, this was an Easter egg just for you 😏 I was going to remove it before merging.

src/domain/schemas/document-request-schemas.js
adrianaluchian: Even better, will change 👍

src/domain/constants.js
ahelmberger: Sneaky :wink:

src/services/document-request-service.js
adrianaluchian: We could make a `date-time-utils` otherwise leave it here, as it is very particular.
I'd go for the first option.

adrianaluchian: Never mind, will go with the clean separation, `date-utils` 😊

adrianaluchian: Done

ahelmberger (approved):

PR #1140, opened 2024-04-08
EDU-1532 Create star rating component

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1532

### DIFF
--- a/src/components/star-rating.less
+++ b/src/components/star-rating.less
@@ -1,20 +1,25 @@
 .StarRating {
-  display: inline-grid;
-  align-items: baseline;
-  justify-content: start;
-  grid-auto-flow: column;
+  display: inline-block;
+  text-align: center;
+  background-color: @edu-background-color-light;
+  padding: @edu-content-padding;
+  border-radius: @edu-border-radius-base;
 }
 
 .StarRating-value {
-  padding-right: 10px;
+  font-weight: bold;
+  color: @edu-black;
 }
 
 .StarRating-rating {
-  .ant-rate-star {
-    margin-inline-end: 0 !important;
+  line-height: 100%;
+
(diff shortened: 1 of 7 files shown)


src/components/star-rating.less
```diff
@@ -0,0 +1,20 @@
...
+  grid-auto-flow: column;
+}
+
+.StarRating-value {
+  padding-right: 10px;
+}
+
+.StarRating-rating {
+  .ant-rate-star {
+    margin-inline-end: 0 !important;
```
ahelmberger, line 14: Unfortunately, the AntD rate component has everything hard-coded (including the star colors), so we cannot easily customize. But IMO the stars had too much space in between by default (8px), so I removed the space here forcefully.

src/components/star-rating.yml
```diff
@@ -0,0 +1,4 @@
+starRating:
+  totalCountText:
+    en: "{totalCount, plural, =0 {no star ratings yet} =1 {# star rating} other {# star ratings}}"
+    de: "{totalCount, plural, =0 {noch keine Sternebewertungen} =1 {# Sternebewertung} other {# Sternebewertungen}}"
```
adrianaluchian, line 4: I would suggest dropping the "star" from "star rating", as rating is the feature and star is just the representation. It sounds a bit much to me. Maybe even dropping "yet" / "noch".

adrianaluchian: After seeing it, I'm even wondering if maybe we should just put the count in parenthesis and have a tooltip saying something like "Number of ratings" or "X users have given a rating"? What do you think? 🤔

ahelmberger: I have reworked the styling based on our discussion.

adrianaluchian (approved):

PR #1144, opened 2024-04-09
EDU-1529 Show rating on document page

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1529

### DIFF
--- a/src/server/document-controller.spec.js
+++ b/src/server/document-controller.spec.js
@@ -14,12 +14,14 @@ describe('document-controller', () => {
 
   let clientDataMappingService;
   let documentRequestService;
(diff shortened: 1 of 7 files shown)


src/server/document-controller.spec.js
```diff
@@ -841,16 +916,16 @@ describe('document-controller', () => {
         assert.notCalled(clientDataMappingService.createProposedSections);
       });
 
-      it('should call documentRequestService.tryRegisterReadRequest', () => {
```
ahelmberger, line 844: This assert was a duplicate of line 848, so I deleted it.

PR #1145, opened 2024-04-11
EDU-1540 Rename statistics tab to requests tab

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1540

### DIFF
--- a/src/components/pages/maintenance.yml
+++ b/src/components/pages/maintenance.yml
@@ -11,6 +11,6 @@ maintenance:
   tagsTabTitle:
     en: Tags
     de: Tags
-  statisticsTabTitle:
-    en: Statistics
-    de: Statistiken
+  requestsTabTitle:
+    en: Requests
+    de: Requests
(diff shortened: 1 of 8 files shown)


src/components/pages/maintenance.yml
```diff
@@ -11,6 +11,6 @@ maintenance:
   tagsTabTitle:
     en: Tags
     de: Tags
-  statisticsTabTitle:
-    en: Statistics
-    de: Statistiken
+  requestsTabTitle:
+    en: Requests
+    de: Zugriffe
```
ahelmberger, line 16: I found "Zugriffe" sounded a lot better than the "official" translation "Anforderungen".

adrianaluchian: You are the German authority here, so it's your call 😁 
I can only say shorter words look better to me.

adrianaluchian (approved):

PR #1159, opened 2024-04-19
EDU-1552 Add document categories tab

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1552

### DIFF
--- a/src/components/icons/icons.js
+++ b/src/components/icons/icons.js
@@ -31,7 +31,8 @@ import {
   IconZoomScan,
   IconHeart,
   IconStarFilled,
-  IconSettings
+  IconSettings,
+  IconCategory2
 } from '@tabler/icons-react';
 
 const Icon = iconNs.default || iconNs;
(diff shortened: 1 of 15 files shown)


src/domain/validation-constants.js
```diff
@@ -7,6 +7,8 @@ export const maxDocumentCommentTextLength = 2000;
 
 export const maxDocumentRevisionCreatedBecauseLength = 200;
 
+export const maxDocumentCategoryNameLength = 25;
```
ahelmberger, line 10: This we should also use in the `documentCategoryDbSchema` validation (which is not included in your branch yet). Just as a reminder.

src/components/icons/icons.js
```diff
@@ -81,6 +86,7 @@ export const FillColorIcon = createIconFromTablerIconComponent(IconBucketDroplet
 export const ChevronLeftIcon = createIconFromTablerIconComponent(IconChevronLeft);
 export const EnterFullscreenIcon = createIconFromTablerIconComponent(IconMaximize);
 export const ChevronRightIcon = createIconFromTablerIconComponent(IconChevronRight);
+export const MediaLibraryIcon = createIconFromTablerIconComponent(IconBuildingBank);
```
ahelmberger, line 89: :+1:

ahelmberger (approved):

src/domain/validation-constants.js
adrianaluchian: Good observation, I rebased from `main` and included it there too 👍

ahelmberger: That is even better than just a reminder :+1:

PR #1161, opened 2024-04-22
EDU-1554 Show existing document categories

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1554

### DIFF
--- a/src/utils/routes.js
+++ b/src/utils/routes.js
@@ -23,6 +23,7 @@ const userProfilePrefix = '/user-profile/';
 const documentInputPrefix = '/doc-inputs/';
 const mediaLibraryPrefix = '/media-library/';
 const samlAuthLoginPrefix = '/saml-auth/login/';
+const documentCategoriesPrefix = '/doc-categories/';
 const samlAuthLoginCallbackPrefix = '/saml-auth/login-callback/';
 const roomMembershipConfirmationPrefix = '/room-membership-confirmation/';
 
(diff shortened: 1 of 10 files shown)


ahelmberger:
Will re-open after refining the UI.

src/utils/routes.js
```diff
@@ -23,6 +23,7 @@ const userProfilePrefix = '/user-profile/';
 const documentInputPrefix = '/doc-inputs/';
 const mediaLibraryPrefix = '/media-library/';
 const samlAuthLoginPrefix = '/saml-auth/login/';
+const documentCategoriesPrefix = '/doc-categories/';
```
adrianaluchian, line 26: In the controller I went with `document-categories`. I honestly would move away from `doc` and `rev` with new logic and stick with full names. I think we should have done the same with `doc-inputs`.
What are your thoughts?

adrianaluchian:
Looks good 👍 

Some small things from me:

- the collapsible arrow and title don't look vertically aligned
Screenshot 2024-04-23 at 12 31 11

- I find it a bit too cluttered to have a one-line footer per category (with 3 different links in there, a bit hard to follow). How about 3 lines? We can afford it, I think.

src/utils/routes.js
ahelmberger: Changed it to `document-categories` as agreed.

ahelmberger:
> Looks good 👍
> 
> Some small things from me:
> 
> * the collapsible arrow and title don't look vertically aligned
> 
> Screenshot 2024-04-23 at 12 31 11
> * I find it a bit too cluttered to have a one-line footer per category (with 3 different links in there, a bit hard to follow). How about 3 lines? We can afford it, I think.

Both issue are adressed (by hacking into antd and putting the footer onto separate lines)

adrianaluchian (approved):

PR #1167, opened 2024-04-23
EDU-1561 Wire up document category page

ahelmberger:
It's not completely empty like the JIRA ticket says, but shows some bare, un-styled information.

Closes https://educandu.atlassian.net/browse/EDU-1561

### DIFF
--- a/src/components/pages/document-category.yml
+++ b/src/components/pages/document-category.yml
@@ -1,7 +1,7 @@
 documentCategory:
   documentCategoryNamePrefix:
-    en: Category
-    de: Kategorie
+    en: Document category
+    de: Dokument-Kategorie
   otherDocumentCategories:
     en: Other document categories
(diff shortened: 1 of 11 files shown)


src/components/pages/document-category.yml
```diff
@@ -0,0 +1,7 @@
+documentCategory:
+  documentCategoryNamePrefix:
+    en: Category
+    de: Kategorie
```
adrianaluchian, line 4: I would suggest to keep consistent with the naming "Document category", even if it's longer 🤔

adrianaluchian (approved):

src/components/pages/document-category.yml
ahelmberger: Agreed, I changed it to "Document category" :+1:

PR #1170, opened 2024-04-25
EDU-1562 Show document categories on document page

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1562


src/components/document-category-display.js
```diff
@@ -0,0 +1,72 @@
...
+    : null;
+
+  const rootClasses = classNames({
+    'DocumentCategoryDisplay': true,
+    'DocumentCategoryDisplay--bordered': bordered,
+    'DocumentCategoryDisplay--detailed': detailed
+  });
+
+  const iconClasses = classNames({
+    'DocumentCategoryDisplay-icon': true,
```
adrianaluchian, line 28: I found that with making the `DocumentCategoryDisplay-icon` `display: block`, it also aligns it vertically centered, like the text is. To me that looks better, at least with the icons I've tried so far.
Screenshot 2024-04-26 at 09 41 35
Screenshot 2024-04-26 at 09 41 51

What do you think?

adrianaluchian: 👍

adrianaluchian (approved):

PR #1176, opened 2024-05-09
EDU-1575 Fix rating position on document page

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1575

I tried both positioning orders (left <-> right):
 
**Desktop**
Screenshot 2024-05-09 at 12 24 20
Screenshot 2024-05-09 at 12 22 42

**Mobile**
Screenshot 2024-05-09 at 12 36 12
Screenshot 2024-05-09 at 12 22 24

I went with the "categories then rating" order because it looks ok on mobile out of the box, with keeping the page padding (so that the rating element is right aligned with the page content). Otherwise we would need to introduce more spacing on the right (which you break right side alignment) or push the page content lower.

Does this sound alright?

### DIFF
--- a/src/components/pages/document.less
+++ b/src/components/pages/document.less
@@ -19,12 +19,15 @@
 }
 
 .DocumentPage-documentCategoryLinks {
-  grid-row: 1;
-  grid-column: 1;
   gap: 10px;
   display: flex;
   flex-wrap: wrap;
   justify-content: flex-start;
+}
+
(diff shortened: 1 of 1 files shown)


ahelmberger:
> Does this sound alright?

Yes, very alright :+1: We should not change the page padding on mobile.

ahelmberger (approved):

PR #1179, opened 2024-05-10
EDU-1570 Show sums per request column

adrianaluchian:
Closes https://educandu.atlassian.net/browse/EDU-1570

### DIFF
--- a/src/components/maintenance/maintenance-document-requests-tab.js
+++ b/src/components/maintenance/maintenance-document-requests-tab.js
@@ -163,12 +163,12 @@ function MaintenanceDocumentRequestsTab() {
     const sortedRows = [...filteredRows].sort(sorters);
 
     const newDisplayedRowsSum = filteredRows.reduce((accu, currentRow) => ({
-      totalCount: (accu.totalCount || 0) + currentRow.totalCount,
-      readCount: (accu.readCount || 0) + currentRow.readCount,
-      writeCount: (accu.writeCount || 0) + currentRow.writeCount,
-      anonymousCount: (accu.anonymousCount || 0) + currentRow.anonymousCount,
-      loggedInCount: (accu.loggedInCount || 0) + currentRow.loggedInCount
-    }), {});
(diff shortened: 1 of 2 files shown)


src/components/maintenance/maintenance-document-requests-tab.js
```diff
@@ -161,7 +162,16 @@ function MaintenanceDocumentRequestsTab() {
 
     const sortedRows = [...filteredRows].sort(sorters);
 
+    const newDisplayedRowsSum = filteredRows.reduce((accu, currentRow) => ({
+      totalCount: (accu.totalCount || 0) + currentRow.totalCount,
+      readCount: (accu.readCount || 0) + currentRow.readCount,
+      writeCount: (accu.writeCount || 0) + currentRow.writeCount,
+      anonymousCount: (accu.anonymousCount || 0) + currentRow.anonymousCount,
+      loggedInCount: (accu.loggedInCount || 0) + currentRow.loggedInCount
+    }), {});
```
ahelmberger, line 171: If there are 0 filteredRows, the outcome of the `reduce` function will be empty object without any properties. It would be more explicit to always have the properties set at least to `0`, no matter what, like:

~~~
    const newDisplayedRowsSum = filteredRows.reduce((accu, currentRow) => ({
      totalCount: accu.totalCount + currentRow.totalCount,
      readCount: accu.readCount + currentRow.readCount,
      writeCount: accu.writeCount + currentRow.writeCount,
      anonymousCount: accu.anonymousCount + currentRow.anonymousCount,
      loggedInCount: accu.loggedInCount + currentRow.loggedInCount
    }), { totalCount: 0, readCount: 0, writeCount: 0, anonymousCount: 0, loggedInCount: 0 });
~~~

What do you think?

adrianaluchian: Yes, that's better, the data model should always be the same as long as it's not `null`. Updated!

ahelmberger (approved):

PR #1181, opened 2024-05-14
EDU-1578 Show rating details

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1578

### DIFF
--- a/src/components/rating-dialog.js
+++ b/src/components/rating-dialog.js
@@ -1,26 +1,27 @@
 import PropTypes from 'prop-types';
-import { Modal, Spin } from 'antd';
-import Markdown from './markdown.js';
 import Logger from '../common/logger.js';
 import StarRating from './star-rating.js';
+import { useUser } from './user-context.js';
 import DeleteButton from './delete-button.js';
 import { useTranslation } from 'react-i18next';
-import { useDateFormat } from './locale-context.js';
+import DocumentRating from './document-rating.js';
+import { Divider, Modal, Progress, Spin } from 'antd';
 import { handleApiError } from '../ui/error-helper.js';
 import { useSessionAwareApiClient } from '../ui/api-helper.js';
-import React, { useCallback, useEffect, useState } from 'react';
 import { documentRatingShape } from '../ui/default-prop-types.js';
+import React, { Fragment, useCallback, useEffect, useState } from 'react';
 import DocumentRatingApiClient from '../api-clients/document-rating-api-client.js';
 
 const logger = new Logger(import.meta.url);
 
-function RatingDialog({ documentRating, isOpen, onCancel, onOk }) {
-  const { formatDate } = useDateFormat();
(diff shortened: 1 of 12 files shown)


src/resources/resources.json
```diff
@@ -3759,26 +3763,26 @@
...
-      "de": "Du hast keine Bewertung für dieses Dokument abgegeben.\n\nBewerte das Dokument, indem du auf die Sterne unten klickst.\n"
+    "stars": {
+      "en": "{starCount, plural, =1 {# star} other {# stars}}",
+      "de": "{starCount, plural, =1 {# Stern} other {# Sterne}}"
     },
-    "existingRatingExplanationMarkdown": {
-      "en": "You have rated this document on {ratingDate} with **{ratingValue, plural, =1 {# star} other {# stars}}**.\n\nYou can change or delete your rating by clicking on the stars or the delete icon below.\n",
-      "de": "Du hast dieses Dokument am {ratingDate} mit **{ratingValue, plural, =1 {# Stern} other {# Sternen}}** bewertet.\n\nDu kannst deine Bewertung ändern oder löschen, indem du auf die Sterne oder das Löschen-Symbol unten klickst.\n"
+    "totalRatingHeader": {
+      "en": "Total rating",
```
adrianaluchian, line 3774: Should we say "Average rating" instead of "total"? 🤔

adrianaluchian: Or "Global average rating", if you feel like it needs to be more clear that it contains all ratings?

ahelmberger: Changed it to "Average rating" / "Durchschnittliche Bewertung".

src/components/rating-dialog.js
```diff
@@ -1,26 +1,27 @@
...
-import React, { useCallback, useEffect, useState } from 'react';
 import { documentRatingShape } from '../ui/default-prop-types.js';
+import React, { Fragment, useCallback, useEffect, useState } from 'react';
 import DocumentRatingApiClient from '../api-clients/document-rating-api-client.js';
 
 const logger = new Logger(import.meta.url);
 
-function RatingDialog({ documentRating, isOpen, onCancel, onOk }) {
-  const { formatDate } = useDateFormat();
+const STAR_STROKE_COLOR = 'rgb(250, 219, 20)';
```
adrianaluchian, line 17: Just a thought, no changes suggested: I'm split between doing this "the ant way" which is not the educandu way, or working against ant and overwriting the `background-color` of the `ant-progress-bg` CSS class 😅

src/resources/resources.json
adrianaluchian: 👍

adrianaluchian (approved):

src/components/rating-dialog.js
ahelmberger: I had the same thought, but as the star color is hard-coded in Antd (and therefore we do not have it in CSS anywhere), I went the Antd way here too.

PR #1201, opened 2024-11-05
EDU-1654 Hide user email addresses from non-admin roles

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1654

### DIFF
--- a/src/domain/permissions.js
+++ b/src/domain/permissions.js
@@ -1,6 +1,6 @@
 import { ROLE } from './constants.js';
 
-const VIEW_USERS = 'view-users';
+const BROWSE_USERS = 'browse-users';
 const MANAGE_USERS = 'manage-users';
 const MANAGE_SETUP = 'manage-setup';
 const CREATE_CONTENT = 'create-content';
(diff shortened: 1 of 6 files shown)


vitest.config.js
```diff
@@ -10,7 +10,7 @@ export default defineConfig({
     include: ['src/**/*.spec.js', 'migrations/**/*.spec.js'],
     exclude: ['migrations/archived/**/*.spec.js'],
     setupFiles: ['src/test-setup.js'],
-    hookTimeout: 20000,
+    hookTimeout: 25000,
```
ahelmberger, line 13: My machine is too slow ;-)

src/domain/permissions.js
```diff
@@ -1,6 +1,6 @@
 import { ROLE } from './constants.js';
 
-const VIEW_USERS = 'view-users';
+const BROWSE_USERS = 'browse-users';
```
ahelmberger, line 3: Technically, we could have deleted this permission completely, because the only place it is used now, is equivalent with `MANAGE_ASSIGNED_EDITORS`, I re-introduced it after deleting it under the new name `BROWSE_USERS` as it makes more sense to protect the user search endpoint with it.

adrianaluchian (approved):

PR #1220, opened 2024-11-26
EDU-1684 Update ABC tools and make code synchronous

ahelmberger:
Closes https://educandu.atlassian.net/browse/EDU-1684


src/components/abc-input.js
```diff
@@ -107,7 +108,7 @@ function AbcInput({
 
   const renderToolsButton = () => (
     <Dropdown
-      placement="top"
+      placement="topLeft"
```
ahelmberger, line 111: This change was necessary, because the Antd dropdown was in a state of constant flickering when on a narrow screen and reaching the right page margin.

adrianaluchian (approved):

